PR #6, opened 2021-02-11
Deployment configurations

petertrr:
### What's done:
* Added spring-boot-gradle-plugin to enable packaging of executable jars
* Added docker-compose.yaml
* Added Dockerfiles for backend and orchestrator
* Added prometheus.yml
* Expose prometheus metrics with spring-boot-actuator
* Added gradle plugin to discover version based on git tag
* Enabled build-info.properties generation for spring boot

# TODO:
* [x] Finish dockerfiles
* [x] Test individual containers locally
* [x] ~Start registry before building images in compose file~ separate command for registry
* [x] Test compose & swarm locally
* [x] Add third party services configs
* [x] Consider building images with [spring boot gradle plugin](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#build-image) (also solves problem with image versions)
* [x] [Endpoint with build info](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#integrating-with-actuator-build-info)
* [ ] Pushing images to local docker registry doesn't work, but it's fine for deploy on single machine now and needs to be tested properly on actual server
* [x] Set image version in docker-compose.yaml during build


save-orchestrator/src/main/resources/application.properties
```diff
@@ -1 +1,2 @@
-server.port = 5100
\ No newline at end of file
+server.port = 5100
```
akuleshov7, line 1: create a table with ports

akuleshov7 (approved):
lgtm

PR #24, opened 2021-02-25
Support more granular AgentStates instead of ERROR

### LINKED ISSUE #17: Agent: support new AgentStates
Instead of current `ERROR` state the following states need to be added and supported:
* `CLI_FAILED` when SAVE CLI returns non-zero status
* `BACKEND_FAILURE` when backend returns non-OK code while uploading test results
* `BACKEND_UNREACHABLE` when backend doesn't respond

petertrr:
### What's done:
* Added states CLI_FAILED, BACKEND_FAILED, BACKEND_UNREACHABLE
* Changed logic in Agent

Closes #17

### DIFF
--- a/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -140,7 +140,7 @@ class SaveAgent(private val config: AgentConfiguration,
                 postExecutionData(executionData)
             }
             if (result.isSuccess && result.getOrNull()?.statusCode == HttpStatusCode.OK) {
-                return@repeat
+                return
             } else {
                 val reason = if (result.isSuccess && result.getOrNull()?.statusCode != HttpStatusCode.OK) {
                     state.value = AgentState.BACKEND_FAILURE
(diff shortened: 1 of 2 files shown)


save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -128,18 +139,25 @@ class SaveAgent(private val config: AgentConfiguration,
             val result = runCatching {
                 postExecutionData(executionData)
             }
-            if (result.isSuccess) {
+            if (result.isSuccess && result.getOrNull()?.statusCode == HttpStatusCode.OK) {
                 return@repeat
```
aktsay6, line 143: Why do we return at repeat here?

petertrr: I guess I initially planned to add some logic after `repeat`, but now it has no difference from usual `return`. I'll change.

petertrr: Actually, `repeat` is a non-inlined function and therefore plain `return` is forbidden there. So I added it back :)

aktsay6 (approved):
lgtm

PR #31, opened 2021-03-01
Backend status handler

### LINKED ISSUE #21: Backend: support controller to get test results and return result status
Backend should have a rest controller that will receive test results and answer to an agent with a status of processing after writing to a database

kentr0w:
### What's done:
Added controller for test status
closes #21

### DIFF
--- a/save-backend/build.gradle.kts
+++ b/save-backend/build.gradle.kts
@@ -27,6 +27,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:${Versions.kotlin}")
     implementation("org.slf4j:slf4j-api:${Versions.slf4j}")
     implementation("ch.qos.logback:logback-core:${Versions.logback}")
+    implementation("com.google.code.gson:gson:${Versions.gson}")
 }
 
 configureJacoco()
--- a/save-backend/src/main/resources/application.properties
+++ b/save-backend/src/main/resources/application.properties
@@ -1,3 +1,4 @@
 server.port = 5000
 management.endpoints.web.exposure.include=health,info,prometheus
 spring.liquibase.enabled=false
+spring.profiles.active=dev
--- a/save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
+++ b/save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
@@ -1,25 +1,21 @@
 package org.cqfn.save.backend
 
-import org.cqfn.save.backend.repository.ProjectRepository
-
 import org.junit.jupiter.api.Test
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.boot.test.autoconfigure.web.reactive.WebFluxTest
-import org.springframework.boot.test.mock.mockito.MockBean
+import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient
+import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.core.io.ByteArrayResource
 import org.springframework.http.client.MultipartBodyBuilder
+import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.web.reactive.server.WebTestClient
 import org.springframework.web.reactive.function.BodyInserters
 
 import kotlin.io.path.ExperimentalPathApi
 
-@WebFluxTest
-class DownloadFilesTest {
-    @MockBean
-    var repository: ProjectRepository? = null
-
-    @Autowired
-    lateinit var webClient: WebTestClient
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("dev")
+@AutoConfigureWebTestClient
+class DownloadFilesTest(@Autowired private val webClient: WebTestClient) {
 
     @Test
     fun checkDownload() {
--- a/save-backend/src/test/kotlin/org/cqfn/save/backend/SaveResultTest.kt
+++ b/save-backend/src/test/kotlin/org/cqfn/save/backend/SaveResultTest.kt
@@ -1,0 +1,29 @@
+package org.cqfn.save.backend
+
+import com.google.gson.Gson
+import org.cqfn.save.entities.Result
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.MediaType
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.web.reactive.server.WebTestClient
+import org.springframework.web.reactive.function.BodyInserters
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("dev")
+@AutoConfigureWebTestClient
+class SaveResultTest(@Autowired private val webClient: WebTestClient) {
+
+    @Test
+    fun checkSave() {
+        val body = Gson().toJson(listOf(Result(1, "run", "3.1.21")))
+        webClient.post()
+            .uri("/result")
+            .contentType(MediaType.APPLICATION_JSON)
+            .body(BodyInserters.fromValue(body))
+            .exchange().expectBody(String::class.java)
+            .isEqualTo<Nothing>("Save")
+    }
+}
--- a/save-common/src/commonMain/kotlin/org/cqfn/save/entities/Result.kt
+++ b/save-common/src/commonMain/kotlin/org/cqfn/save/entities/Result.kt
@@ -1,0 +1,8 @@
+package org.cqfn.save.entities
+
+@Entity
+class Result(
+    @Id val id: Int,
+    val status: String,
+    val date: String // fixme should be date
+)
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -14,4 +14,5 @@ object Versions {
     const val micrometer = "1.6.3"
     const val h2 = "1.4.200"
     const val jpa = "1.0.2.Final"
+    const val gson = "2.8.6"
 }
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ResultController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ResultController.kt
@@ -1,0 +1,30 @@
+package org.cqfn.save.backend.controllers
+
+import org.cqfn.save.backend.service.ResultService
+import org.cqfn.save.entities.Result
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RestController
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+
+/**
+ * Controller to work with test result
+ * @param resultService service for result
+ */
+@RestController
+class ResultController(private val resultService: ResultService) {
+
+    /**
+     * @param results list of test result
+     * @return [Mono] with respone
+     */
+    @PostMapping(value = ["/result"])
+    fun saveResult(@RequestBody results: List<Result>): Mono<ResponseEntity<String>> {
+        Flux.just(results).subscribe {
+            resultService.addResults(it)
+        }
+        return Mono.just(ResponseEntity.ok().body("Save"))
+    }
+}
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ResultRepository.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ResultRepository.kt
@@ -1,0 +1,11 @@
+package org.cqfn.save.backend.repository
+
+import org.cqfn.save.entities.Result
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+
+/**
+ * Repository for test result
+ */
+@Repository
+interface ResultRepository : JpaRepository<Result, Int>
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/ResultService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/ResultService.kt
@@ -1,0 +1,19 @@
+package org.cqfn.save.backend.service
+
+import org.cqfn.save.backend.repository.ResultRepository
+import org.cqfn.save.entities.Result
+import org.springframework.stereotype.Service
+
+/**
+ * Service for test result
+ */
+@Service
+class ResultService(private val resultRepository: ResultRepository) {
+
+    /**
+     * @param testResults list of test result
+     */
+    fun addResults(testResults: List<Result>) {
+        resultRepository.saveAll(testResults)
+    }
+}
--- a/save-backend/src/main/resources/data.sql
+++ b/save-backend/src/main/resources/data.sql
@@ -1,4 +1,5 @@
 DROP TABLE IF EXISTS project;
+DROP TABLE IF EXISTS result;
 
 CREATE TABLE project (
   id INT AUTO_INCREMENT  PRIMARY KEY,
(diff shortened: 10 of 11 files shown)


save-backend/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:${Versions.kotlin}")
     implementation("org.slf4j:slf4j-api:${Versions.slf4j}")
     implementation("ch.qos.logback:logback-core:${Versions.logback}")
+    implementation("com.google.code.gson:gson:${Versions.gson}")
```
petertrr, line 30: We don't use gson, we use kx serialization

save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
```diff
@@ -1,25 +1,21 @@
...
 import kotlin.io.path.ExperimentalPathApi
 
-@WebFluxTest
-class DownloadFilesTest {
-    @MockBean
-    var repository: ProjectRepository? = null
-
-    @Autowired
-    lateinit var webClient: WebTestClient
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
petertrr, line 15: `SpringBootTest` is more about e2e testing. Will combination of `@ExtendWith(SpringExtension.class)`,
`@WebFluxTest` and `@Import` with necessary components be more appropriate here?

save-backend/src/main/resources/application.properties
```diff
@@ -1,3 +1,4 @@
 server.port = 5000
 management.endpoints.web.exposure.include=health,info,prometheus
-spring.liquibase.enabled=false
\ No newline at end of file
+spring.liquibase.enabled=false
+spring.profiles.active=dev
```
petertrr, line 4: This shouldn't be activated here, but rather in tests or manually if you are running services locally

save-backend/src/test/kotlin/org/cqfn/save/backend/SaveResultTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.web.reactive.function.BodyInserters
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("dev")
+@AutoConfigureWebTestClient
+class SaveResultTest(@Autowired private val webClient: WebTestClient) {
+
+    @Test
+    fun checkSave() {
+        val body = Gson().toJson(listOf(Result(1, "run", "3.1.21")))
```
petertrr, line 21: Use kx serialization here; spring boot should be able to work with it too

save-common/src/commonMain/kotlin/org/cqfn/save/entities/Result.kt
```diff
@@ -0,0 +1,8 @@
+package org.cqfn.save.entities
+
+@Entity
+class Result(
+    @Id val id: Int,
+    val status: String,
```
petertrr, line 6: Should be enum of statuses

petertrr (changes requested):

save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
petertrr: Nevermind, let's use `SpringBootTest`, but with testcontainers, which are already in master.

petertrr:
Converted to draft, because it's not finished and will probably need to be updated after we merge liquibase

akuleshov7 (approved):
Lgtm, thx

save-common/src/commonMain/kotlin/org/cqfn/save/entities/Result.kt
```diff
@@ -0,0 +1,18 @@
...
+ * @property id id of test
+ * @property status of test
+ * @property date date of result
+ */
+@Entity
+@Serializable
+data class Result(
+    @Id val id: Int,
+    val status: ResultStatus,
+    val date: String  // fixme should be date
```
akuleshov7, line 17: I would actually use date here (json parsers support it automatically)

petertrr: This is common class now, we could either use `kotlinx.datetime` here or move to `jvmMain` source set and create a dto in common

petertrr (changes requested):
I'll validate it once again after merge with master, in case there are some problems regarding liquibase

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -0,0 +1,28 @@
+package org.cqfn.save.backend.controllers
+
+import org.cqfn.save.backend.service.ProjectService
+import org.cqfn.save.entities.Project
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+import java.net.http.HttpClient
```
petertrr, line 9: Use client from spring

save-backend/src/main/resources/application.properties
petertrr: https://github.com/cqfn/save-cloud/pull/31/commits/c8328125781ebc82757436e6c25c22613b40c85e#r584667392

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -0,0 +1,28 @@
...
+@RestController
+class CloneRepositoryController(private val projectService: ProjectService) {
+
+    @PostMapping(value = ["/backTest"])
+    fun saveResult(@RequestBody text: String): String {
+        val project = Project("fd", "df", "df")
+        projectService.addResults(project)
+        val client = HttpClient.newBuilder().build()
+        val request = HttpRequest.newBuilder()
+                .uri(URI.create("http://preprocessor:5200/testbackAndproc"))
```
petertrr, line 22: URL of preprocessor should be in properties

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -0,0 +1,28 @@
...
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+@RestController
+class CloneRepositoryController(private val projectService: ProjectService) {
+
+    @PostMapping(value = ["/backTest"])
+    fun saveResult(@RequestBody text: String): String {
+        val project = Project("fd", "df", "df")
```
petertrr, line 18: Project should be passed in request params

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Service for test result
+ */
+@Service
+class TestExecutionService(private val testResultRepository: TestExecutionRepository) {
+    /**
+     * @param testExecutions list of test executions
+     */
+    fun saveTestResult(testExecutions: List<TestExecution>) {
+        testExecutions.forEach { testExec ->
+            testResultRepository.run {
```
petertrr, line 17: I think `run` is redundant here, because `testResultRepository` already is non-nullable and immutable, so scope function doesn't give any benefits

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -0,0 +1,32 @@
...
+                val foundTestExec = this.findById(testExec.id)
+                 if (!foundTestExec.isPresent) {
+                     println("error")
```
petertrr, line 20: Use slf4j logger, use `ERROR` logging level and provide way more details inthe message

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -0,0 +1,32 @@
...
+                 } else {
+                     foundTestExec.map {
```
petertrr, line 22: Is `foundTestExec` a collection? I think in norma situation in would be a collection with a single element, so if there are more you shouldn't update all of them

save-backend/src/test/kotlin/org/cqfn/save/backend/DatabaseTest.kt
```diff
@@ -17,7 +17,7 @@ class DatabaseTest : DatabaseTestBase() {
     private lateinit var projectRepository: ProjectRepository
 
     @Autowired
-    private lateinit var testStatusRepository: TestStatusRepository
+    private lateinit var testExecutionRepository: TestExecutionRepository
```
petertrr, line 20: Pls remove `data.sql` from resources

save-backend/src/main/kotlin/org/cqfn/save/backend/service/ProjectService.kt
```diff
@@ -0,0 +1,13 @@
+package org.cqfn.save.backend.service
+
+import org.cqfn.save.backend.repository.ProjectRepository
+import org.cqfn.save.entities.Project
+import org.springframework.stereotype.Service
+
+@Service
+class ProjectService (private val projectRepository: ProjectRepository) {
+
+    fun addResults(project: Project) {
```
petertrr, line 10: What is this method doing? Saving project to projects table when asked to save results?)

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
kentr0w: Yes, forgot to replace it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -16,11 +16,15 @@ import org.springframework.web.bind.annotation.RestController
...
-     * @return
+     * @return response
      */
     @PostMapping(value = ["/saveTestResult"])
     fun saveTestResult(@RequestBody testExecutions: List<TestExecution>): ResponseEntity<String> {
-        testExecutionService.saveTestResult(testExecutions)
+        try {
+            testExecutionService.saveTestResult(testExecutions)
+        } catch (exception: IllegalArgumentException) {
+            return ResponseEntity.ok().body("Error to save")
```
petertrr, line 26: Why `ok` if there is an error? Also, you should never silently return error and do nothing else. You definitely should log an error message in backend so that we can find out what happened here.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -2,29 +2,30 @@ package org.cqfn.save.backend.service
...
-                     println("error")
-                 } else {
-                     foundTestExec.map {
-                         it.startTime = testExec.startTime
-                         it.endTime = testExec.endTime
-                         it.status = testExec.status
-                         this.save(it)
-                     }
+            val foundTestExec = testResultRepository.findById(testExec.id)
+            if (!foundTestExec.isPresent) {
```
petertrr, line 21: Oh, it's `Optional` from Java, I forgot it. Then it has useful methods like `ifPresentOrElse`

db/changelog/v-1/03-changeset-test-status-table.xml
```diff
@@ -5,26 +5,20 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
 
-    <changeSet id="10" author="aktsay" context="dev or prod">
+    <changeSet id="10" author="kentr0w" context="dev or prod">
         <preConditions onFail="MARK_RAN">
             <not>
-                <tableExists tableName="test_status"/>
+                <tableExists tableName="test_execution"/>
```
petertrr, line 11: This file should be renamed from status to executino too

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -0,0 +1,30 @@
...
+class TestExecutionController(private val testExecutionService: TestExecutionService) {
+    /**
+     * @param testExecutions list of test executions
+     * @return response
+     */
+    @PostMapping(value = ["/saveTestResult"])
+    fun saveTestResult(@RequestBody testExecutions: List<TestExecution>): ResponseEntity<String> {
+        try {
+            testExecutionService.saveTestResult(testExecutions)
+        } catch (exception: IllegalArgumentException) {
```
petertrr, line 25: Why specifically `IllegalArgumentException`? If you mean `save` in repository, than it throws IAE only if supplied element is null, but if it is null, then `Optional.get` would've thrown NSEE earlier.

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
kentr0w: No, baceuse method `findById` can throw `IllegalArgumentException`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
kentr0w: But we already log error in `TestExecutionService`. Should we log here too?

db/changelog/v-1/03-changeset-test-status-table.xml
kentr0w: Renamed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
petertrr: Docs also state that it does so only if supplied id is null, if element is not present in the DB, it returns `Optional.empty`. I think IAE is too specific and rare to catch here.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
petertrr: No, you are right, logging is redundant here.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -18,16 +18,18 @@ class TestExecutionService(private val testResultRepository: TestExecutionReposi
...
-            } else {
-                foundTestExec.get().run {
-                    this.startTime = testExec.startTime
-                    this.endTime = testExec.endTime
-                    this.status = testExec.status
-                    testResultRepository.save(this)
-                }
-            }
+            foundTestExec.ifPresentOrElse(
+                {
```
petertrr, line 22: This lambda is  a consumer for optional value, no need to call `get` inside of it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -22,8 +24,8 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
     fun saveTestResult(@RequestBody testExecutions: List<TestExecution>): ResponseEntity<String> {
         try {
             testExecutionService.saveTestResult(testExecutions)
-        } catch (exception: IllegalArgumentException) {
-            return ResponseEntity.ok().body("Error to save")
+        } catch (exception: DataAccessException) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body("Error to save")
```
petertrr, line 28: I'd rather say it is internal server error. Do we have situation when it can be bad request? When test execution id is not present in DB? Then this situation should be separated from the situation when there are other problems with DB. I believe spring data throws different subtypes of `DataAccessException` in these cases

petertrr (changes requested):

petertrr (approved):

PR #33, opened 2021-03-02
Create preprocessor

### LINKED ISSUE #10: Test-preprocessor: Create test-preprocessor capable of cloning git repo
Create a new spring boot microservice, test-preprocessor, which accepts POST with repo link and clones it (using jgit or just ProcessBuilder, need to investigate advantages) into a [docker volume](https://docs.docker.com/storage/volumes/) shared with orchestrator (couple of lines in docker-compose.yaml).

kentr0w:
### What's done:
[x] Created preprocessor
[x] Created controller to clone repo
closes #10


save-preprocessor/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+import org.cqfn.save.buildutils.configureSpringBoot
+
+plugins {
+    kotlin("jvm")
+}
+
+configureSpringBoot()
```
petertrr, line 7: Please use `withDataJpa=true` instead of manual configuration

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/dto/GitRepoDto.kt
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Data class with repository information
+ *
+ * @param url - url of repo
+ * @param username - username to credential
+ * @param password - password to credential
+ * @param branch - branch to clone
+ */
+@Serializable
+data class GitRepoDto(
```
petertrr, line 14: This should be moved to `save-common`, because it will be shared between backend and preprocessor

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/dto/GitRepoDto.kt
```diff
@@ -0,0 +1,19 @@
+package org.cqfn.save.preprocessor.dto
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Data class with repository information
+ *
+ * @param url - url of repo
+ * @param username - username to credential
+ * @param password - password to credential
```
petertrr, line 10: Add a note or a fixme comment, that we will operate not with raw passwords but with some sort of token, e.g. from github integration

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,54 @@
...
+                .setURI(gitRepoDto.url)
+                .setCredentialsProvider(user)
+                .setDirectory(File(volumes))
+                .call().use {
+                    log.info("repository cloned")
+                    // TODO post request to orchestrator
+                    return ResponseEntity.ok("Cloned")
+                }
+        } catch (ex: TransportException) {
+            log.warn("No credentials with private repo")
```
petertrr, line 43: I suspect that `TransportException` can have a lot of different causes. Please log them and/or include in the response too

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,54 @@
...
+        } else {
+            CredentialsProvider.getDefault()
+        }
+        try {
+            Git.cloneRepository()
+                .setURI(gitRepoDto.url)
+                .setCredentialsProvider(user)
+                .setDirectory(File(volumes))
+                .call().use {
+                    log.info("repository cloned")
```
petertrr, line 38: Which repository? Cloned where?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,54 @@
...
+                    // TODO post request to orchestrator
+                    return ResponseEntity.ok("Cloned")
+                }
+        } catch (ex: TransportException) {
+            log.warn("No credentials with private repo")
+            return ResponseEntity.ok("No credentials")
+        } catch (ex: Exception) {
+            log.warn("Error to clone repo")
```
petertrr, line 46: ex.message? stacktrace?

petertrr (changes requested):

docker-compose.yaml.template
```diff
@@ -19,6 +20,15 @@ services:
     deploy:
       labels:
         - "prometheus-job=save-backend"
+  preprocessor:
+    image: 127.0.0.1:6000/save-preprocessor:{{project.version}}
+    volumes:
```
petertrr, line 25: I believe you should have a separate sections for volumes, as describe in [this section of docs](https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose). And that volume can be mounted only to required containers and not exposed to host system.

save-preprocessor/build.gradle.kts
kentr0w: I don't need jpa (at least now). By default `withSpringDataJpa` is false

petertrr (commented):
Also please add the new service to depoyLocal gradle task

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -20,35 +20,36 @@ class DownloadProject {
...
-            log.warn("Error to clone repo")
-            return ResponseEntity.ok("Error to clone")
+            log.warn(ex.stackTraceToString())
+            return ResponseEntity.ok(ex.stackTraceToString())
         }
     }
 
     companion object {
-        private const val volumes = "~/var/lib/docker/volumes/repository"
+        private const val VOLUMES = "~/var/lib/docker/volumes/repository"
```
petertrr, line 53: This can be placed in properties file, so that you can override it in tests and check if the service stores data using some temporary directory.

petertrr (changes requested):
What about tests?

docker-compose.yaml.template
```diff
@@ -9,6 +9,7 @@ services:
       - "5100:5100"
     volumes:
       - "/var/run/docker.sock:/var/run/docker.sock"
+      - save: /home/repositories
```
petertrr, line 12: Please check if space after `:` is valid.

docker-compose.yaml.template
```diff
@@ -19,6 +20,15 @@ services:
     deploy:
       labels:
         - "prometheus-job=save-backend"
+  preprocessor:
+    image: 127.0.0.1:6000/save-preprocessor:{{project.version}}
+    volumes:
+      - save: /home/repositories
+    ports:
+      - "5200:5200"
```
petertrr, line 28: Please add the port to ports allocation table in `save-deploy/README.md`

save-preprocessor/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+
+dependencies {
+    implementation(project(":save-common"))
+    implementation("org.eclipse.jgit:org.eclipse.jgit:${Versions.jgit}")
+    implementation("junit:junit:4.12")
```
petertrr, line 12: Why junit 4?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * @param gitRepository - Dto of repo information to clone
+     * @return response entity with text
+     */
+    @Suppress("TooGenericExceptionCaught")
+    @PostMapping(value = ["/upload"])
+    fun upload(@RequestBody gitRepository: GitRepository): ResponseEntity<String> {
+        val tmpDir = File("$VOLUMES/${gitRepository.url.hashCode()}")
+        tmpDir.deleteRecursively()
+        log.info("dir created")
```
petertrr, line 32: Which dir? Please be more verbose in logging, it will help when it'll be time to investigate bugs

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,58 @@
...
+                .setURI(gitRepository.url)
+                .setCredentialsProvider(user)
+                .setDirectory(tmpDir)
+                .call().use {
+                    log.info("Repository cloned: ${gitRepository.url}")
+                    // TODO post request to orchestrator
+                    return ResponseEntity("Cloned", HttpStatus.ACCEPTED)
+                }
+        } catch (ex: Exception) {
+            log.warn(ex.stackTraceToString())
```
petertrr, line 50: You don't need to get stacktrace explicitly in slf4j, just use `log.warn("some string", ex)`

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/DownloadProjectTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.test.context.junit4.SpringRunner
+import java.io.File
+
+@RunWith(SpringRunner::class)
+@SpringBootTest(classes = [SaveApplication::class],
+        webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class DownloadProjectTest {
+
+    @Autowired
+    lateinit var testRestTemplate: TestRestTemplate
```
petertrr, line 25: For webflix usually WebClient (`WebTestClient` for tests) is preferred over rest template

petertrr (changes requested):

petertrr (changes requested):

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -47,6 +47,6 @@ fun Project.createStackDeployTask() {
     tasks.register<Exec>("deployLocal") {
         dependsOn(subprojects.flatMap { it.tasks.withType<BootBuildImage>() })
         dependsOn("generateComposeFile")
-        commandLine("docker-compose", "--file", "$buildDir/docker-compose.yaml", "up", "-d", "orchestrator", "backend")
+        commandLine("docker-compose", "--file", "$buildDir/docker-compose.yaml", "up", "-d", "orchestrator", "backend", "preprocessor")
```
petertrr, line 50: Did you try this command locally? Are volumes created successfully?

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/utils/RepositoryVolume.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+open class RepositoryVolume {
+    companion object {
+        /**
+         * @param registry
+         */
+        @DynamicPropertySource
+        @JvmStatic
+        fun dbProperties(registry: DynamicPropertyRegistry) {
+            registry.add("save.repository") { "../save-preprocessor/build" }
```
petertrr, line 17: Please use temporary directory, you can then get its path from this property in test.

petertrr (changes requested):

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/utils/RepositoryVolume.kt
```diff
@@ -14,7 +14,7 @@ open class RepositoryVolume {
         @DynamicPropertySource
         @JvmStatic
         fun dbProperties(registry: DynamicPropertyRegistry) {
-            registry.add("save.repository") { "../save-preprocessor/build" }
+            registry.add("save.repository") { "../save-preprocessor/tempDir" }
```
petertrr, line 17: I meant https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/create-temp-directory.html

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,60 @@
...
+     * @param gitRepository - Dto of repo information to clone
+     * @return response entity with text
+     */
+    @Suppress("TooGenericExceptionCaught")
+    @PostMapping(value = ["/upload"])
+    fun upload(@RequestBody gitRepository: GitRepository): Response {
+        val urlHash = gitRepository.url.hashCode()
+        val tmpDir = File("$volumes/$urlHash")
+        tmpDir.deleteRecursively()
+        log.info("For ${gitRepository.url} repository: dir $urlHash was created")
```
petertrr, line 39: Why do you say that it has been created when you have just deleted it? You should call `mkdirs` for it to be created

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,60 @@
...
+        val user = if (gitRepository.username != null && gitRepository.password != null) {
```
petertrr, line 40: Very confusing variable naming

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,60 @@
...
+            UsernamePasswordCredentialsProvider(gitRepository.username, gitRepository.password)
+        } else {
+            CredentialsProvider.getDefault()
+        }
+        try {
+            Git.cloneRepository()
+                .setURI(gitRepository.url)
+                .setCredentialsProvider(user)
+                .setDirectory(tmpDir)
+                .call().use {
```
petertrr, line 50: Why do you need `use` here? Does `call` return Closeable?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,60 @@
...
+                    log.info("Repository cloned: ${gitRepository.url}")
+                    // TODO post request to orchestrator
+                    return Mono.just(ResponseEntity("Cloned", HttpStatus.ACCEPTED))
+                }
+        } catch (ex: Exception) {
+            log.warn("Cloning ${gitRepository.url} repository failed", ex)
+            return Mono.just(ResponseEntity(ex.stackTraceToString(), HttpStatus.BAD_REQUEST))
```
petertrr, line 57: Why is this a bad request? For what we know, it more likely will be internal server error. You should distinguish these cases by catching different kinds of exceptions (e.g. bad credentials - then bad request, io error - then internal server error)

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/DownloadProjectTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.MediaType
+import org.springframework.test.web.reactive.server.WebTestClient
+import org.springframework.web.reactive.function.BodyInserters
+import java.io.File
+import kotlin.io.path.ExperimentalPathApi
+
+@ExperimentalPathApi
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
petertrr, line 20: Is `WebFluxTest` enough here?

petertrr (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -35,26 +38,39 @@ class DownloadProject {
...
+                .setCredentialsProvider(userCredentials)
                 .setDirectory(tmpDir)
                 .call().use {
                     log.info("Repository cloned: ${gitRepository.url}")
                     // TODO post request to orchestrator
                     return Mono.just(ResponseEntity("Cloned", HttpStatus.ACCEPTED))
                 }
+        } catch (invalidRemoteException: InvalidRemoteException) {
+            log.warn("Invalidate remote exception while cloning ${gitRepository.url} repository", invalidRemoteException)
+            return Mono.just(ResponseEntity(invalidRemoteException.stackTraceToString(), HttpStatus.BAD_GATEWAY))
```
petertrr, line 64: I think all jgit-specific exceptions should return BAD_REQUEST, because it means that data supplied from user is incorrect.

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/DownloadProjectTest.kt
```diff
@@ -0,0 +1,56 @@
...
+            .contentType(MediaType.APPLICATION_JSON)
+            .body(BodyInserters.fromValue(wrongRepo))
+            .exchange()
+            .expectStatus()
+            .isEqualTo(HttpStatus.BAD_GATEWAY)
+    }
+
+    @Test
+    fun testCorrectDownload() {
+        val wrongRepo = GitRepository("https://github.com/cqfn/save-cloud.git")
```
petertrr, line 39: Why is this wrongRepo?

petertrr (changes requested):
Almost there. Please also add the new port 5200 to `save-depoy/README.md`.

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
kentr0w: Ok, fixed

kentr0w:
But you have already added it.

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -0,0 +1,77 @@
...
+    @Value("\${save.repository}")
+    private lateinit var volumes: String
+
+    /**
+     * @param gitRepository - Dto of repo information to clone
+     * @return response entity with text
+     */
+    @Suppress("TooGenericExceptionCaught", "TOO_LONG_FUNCTION")
+    @PostMapping(value = ["/upload"])
+    fun upload(@RequestBody gitRepository: GitRepository): Response {
```
petertrr, line 38: We have decided that cloning should occur in background. It may take a lot of time, so this handler should return reposnse immediately (indicating that it has started execution) and then start cloning in background. It can be acieved like this:
```kotlin
return Mono.just( /* something with OK status */ )
    .subscribeOn(Schedulers.boundedElastic())
    .also {
        it.subscribe {
            // clone repo here
        }
    }
```

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -36,42 +37,46 @@ class DownloadProject {
...
-                is InvalidRemoteException,
-                is TransportException,
-                is GitAPIException -> {
-                    log.warn("Error with git API while cloning ${gitRepository.url} repository", exception)
-                    Mono.just(ResponseEntity(exception.stackTraceToString(), HttpStatus.BAD_REQUEST))
-                }
-                else -> {
-                    log.warn("Cloning ${gitRepository.url} repository failed", exception)
-                    Mono.just(ResponseEntity(exception.stackTraceToString(), HttpStatus.INTERNAL_SERVER_ERROR))
+        return Mono.just(ResponseEntity("Cloned", HttpStatus.ACCEPTED))
```
petertrr, line 40:
```suggestion
        return Mono.just(ResponseEntity("Clone pending", HttpStatus.ACCEPTED))
```

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -36,42 +37,46 @@ class DownloadProject {
...
+            .subscribeOn(Schedulers.boundedElastic())
+            .also {
+                it.subscribe {
+                    val urlHash = gitRepository.url.hashCode()
```
petertrr, line 44: It's time to extract this huge lambda to a separate function

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -36,42 +37,46 @@ class DownloadProject {
...
+                        Git.cloneRepository()
+                            .setURI(gitRepository.url)
+                            .setCredentialsProvider(userCredentials)
+                            .setDirectory(tmpDir)
+                            .call().use {
+                                log.info("Repository cloned: ${gitRepository.url}")
+                                // TODO post request to orchestrator
+                            }
+                    } catch (exception: Exception) {
+                        tmpDir.deleteRecursively()
```
petertrr, line 67: Should be in `finally`

petertrr (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
kentr0w: Why it should be in finally? We delete directory only if there was an exception.

petertrr (changes requested):
Please fix code style

petertrr (changes requested):
Please update this branch and add `env` section to preprocessor in `docker-compose.yaml.template`, like for two other services. Then this PR will be done.

petertrr (approved):
lgtm

PR #44, opened 2021-03-15
Added configuration for local database

### LINKED ISSUE #39: Infra: Configure local database for manual testing and local deployment
~We should also think about local deployment of services, including database. There are several options:~
~* Use spring-liquibase integration and add a testcontainers entity to dev profile~

Add mysql DB to docker-compose for local deployment. This option will make liquibase deployment consistent for dev and prod. For this feature we should leverage [docker-compose profiles](https://docs.docker.com/compose/profiles/).

We also should add instruction on how to manually test services when running directly from IDE or via `bootRun` without building an image (it might be useful to automate it in the future too). The difference is in host names which containers can use to access DB and each other.

~_Originally posted by @petertrr in https://github.com/cqfn/save-cloud/issues/20#issuecomment-795071668_~

petertrr:
### What's done:
* Added mysql with default password and default database to docker-compose.yaml.template in dev profile
* Added startMysqlDb task
* Updated save-deploy/README.md
* Added run configuration for IDEA
* Fixed application-dev.properties

Closes #39

### DIFF
--- a/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
+++ b/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
@@ -55,9 +55,11 @@ fun Project.createStackDeployTask(profile: String) {
         commandLine("docker-compose", "--file", "$buildDir/docker-compose.yaml", "--profile", "dev", "up", "-d", "mysql")
         errorOutput = ByteArrayOutputStream()
         if (!errorOutput.toString().contains(" is up-to-date")) {
-            logger.info("Waitnig 5 sec for mysql to start")
-            Thread.sleep(5_000)  // wait 5 sec for mysql to start, can be manually increased when needed
+            val waitIntervalMs = 10_000L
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -49,9 +50,19 @@ fun Project.createStackDeployTask(profile: String) {
 //        }
     }
 
+    tasks.register<Exec>("startMysqlDb") {
+        dependsOn("generateComposeFile")
```
aktsay6, line 54: What about liquibase update task?

petertrr: Thanks, added. Also had to add two different datasource urls in backend: one to use inside the container and one for liquibase task. The latter should also be used when running backend witout docker too, though.

aktsay6 (approved):
lgtm

PR #45, opened 2021-03-15
Preprocessor: add class for configuration properties

### LINKED ISSUE #38: Create and inject classes for configuration properties
Now we even don't have `@Configuration` classes in our microservices. We need to create them properly.

Read this:
https://www.baeldung.com/properties-with-spring
https://habr.com/ru/post/487980/

kentr0w:
### What's done:
Added configuration for preprocessor
closes #38

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -21,4 +21,5 @@ object Versions {
     const val kotlinJsWrappersSuffix = "-pre.148-kotlin-1.4.30"
     const val kotlinReact = "$react$kotlinJsWrappersSuffix"
     const val jgit = "5.10.0.202012080955-r"
+    const val ktReflect = "1.4.31"
 }
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -21,5 +21,4 @@ object Versions {
     const val kotlinJsWrappersSuffix = "-pre.148-kotlin-1.4.30"
     const val kotlinReact = "$react$kotlinJsWrappersSuffix"
     const val jgit = "5.10.0.202012080955-r"
-    const val ktReflect = "1.4.31"
 }
--- a/save-preprocessor/build.gradle.kts
+++ b/save-preprocessor/build.gradle.kts
@@ -9,7 +9,7 @@ configureSpringBoot()
 dependencies {
     implementation(project(":save-common"))
     implementation("org.eclipse.jgit:org.eclipse.jgit:${Versions.jgit}")
-    runtimeOnly("org.jetbrains.kotlin:kotlin-reflect:1.4.31")
+    runtimeOnly("org.jetbrains.kotlin:kotlin-reflect:${Versions.ktReflect}")
 }
 
 tasks.withType<Test> {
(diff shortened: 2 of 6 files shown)


save-preprocessor/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ configureSpringBoot()
 dependencies {
     implementation(project(":save-common"))
     implementation("org.eclipse.jgit:org.eclipse.jgit:${Versions.jgit}")
+    runtimeOnly("org.jetbrains.kotlin:kotlin-reflect:1.4.31")
```
petertrr, line 12: Why?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
```diff
@@ -22,14 +22,13 @@ import java.io.File
 
 /**
  * A Spring controller for git project downloading
+ *
+ * @property configProperties config properties
  */
 @RestController
-class DownloadProject {
+class DownloadProject(val configProperties: ConfigProperties) {
```
petertrr, line 29: Should it be `@autowired`?

petertrr (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProject.kt
kentr0w: No, it shouldn't. We can omit the `@autowired`

save-preprocessor/build.gradle.kts
kentr0w: `@ConstructorBinding` and `@ConfigurationProperties` require it.

petertrr: I don't quite see why they should require it. Maybe you should add `@ConstructorBinding` on constructor and not on the class?

petertrr: Nevermind, I believe it's correct, because it needs to aces setters based on property name known only at runtime.

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -21,4 +21,5 @@ object Versions {
     const val kotlinJsWrappersSuffix = "-pre.148-kotlin-1.4.30"
     const val kotlinReact = "$react$kotlinJsWrappersSuffix"
     const val jgit = "5.10.0.202012080955-r"
+    const val ktReflect = "1.4.31"
```
petertrr, line 24: I suppose it shares the version with kotlin? Why separate variable?

save-preprocessor/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ configureSpringBoot()
 dependencies {
     implementation(project(":save-common"))
     implementation("org.eclipse.jgit:org.eclipse.jgit:${Versions.jgit}")
+    runtimeOnly("org.jetbrains.kotlin:kotlin-reflect:${Versions.ktReflect}")
```
petertrr, line 12: Why is it runtimeOnly? How will it be provided in runtime?

buildSrc/src/main/kotlin/Versions.kt
kentr0w: Yes, fixed

save-preprocessor/build.gradle.kts
kentr0w: I forgot to replace it to `implementation`. Fixed

petertrr (approved):

Issue #47, opened 2021-03-17
Database: add schema description and support models 


akuleshov7:
1. Add schemas of a database to wiki
2. Support database schemas in backend

Issue #43, opened 2021-03-15
Database: create table for test suites and test suite executions


petertrr:
test_suite - a group of tests (note, that they will change between commits)
test_suite_execution - a group of test_executions from a particular test_suite, should have field for path in docker volume

PR #48, opened 2021-03-17
Feature. Controllers for database

### LINKED ISSUE #37: Backend: support controllers for database related work
We (with @petertrr) decided that we should have a single point to work with a database for all microservices.
For example: `orchestrator`, `backend-preprocessor` and `backend` should work with the same microservice that will do the same base-related work. 

So we need 2 endpoints for it:
1. Orchestrator (to receive post requests with Statuses)
2. Backend-preprocessor (to receive post requests with test IDs)

aktsay6:
### What's done:
  * Added schemes for tests, test suites and execution
  * Added controllers to backend to receive post from preprocessor
  * Added tests

This issue closes #37 #43 #47


db/changelog/v-1/04-changeset-test-table.xml
```diff
@@ -0,0 +1,37 @@
...
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="add-test-table" author="aktsay" context="dev or prod">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="test"/>
+            </not>
+        </preConditions>
+        <createTable tableName="test">
+            <column name="id" type="varchar(255)">
```
petertrr, line 15: I believe we should use bigint for all ids now

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Controller that accepts executions
+ */
+@RestController
+class ExecutionController(private val executionService: ExecutionService) {
+    /**
+     * @param execution
+     */
+    @PostMapping("/createExecution")
+    fun createExecution(@RequestBody execution: Execution) {
```
petertrr, line 18: Maybe add `updateExecution` rn too?

save-orchestrator/src/main/resources/application-dev.properties
```diff
@@ -0,0 +1,14 @@
+spring.h2.console.enabled=true
```
petertrr, line 1: Is it by mistake? We don't need DB setttings in orchestrator

petertrr (changes requested):

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/ExecutionControllerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+            .uri("/createExecution")
+            .contentType(MediaType.APPLICATION_JSON)
+            .body(BodyInserters.fromValue(execution))
+            .exchange()
+            .expectStatus()
+            .isOk
+
+        val databaseData = executionRepository.findAll()
+
+        assert(databaseData.any { it.status == execution.status && it.id == execution.id })
```
petertrr, line 68: Use junit or kotlin.test assertions (`assertEquals`, `assertTrue`) instead of JVM-asserts

save-backend/src/main/kotlin/org/cqfn/save/backend/service/ExecutionService.kt
```diff
@@ -15,4 +18,15 @@ class ExecutionService(private val executionRepository: ExecutionRepository) {
     fun saveExecution(execution: Execution) {
         executionRepository.save(execution)
     }
+
+    /**
+     * @param execution
+     */
+    fun updateExecution(execution: ExecutionUpdateDto) {
+        val databaseExecution = executionRepository.findById(execution.id).get()
```
petertrr, line 26: You shouldn't rely on incoming request correctness; therefore you shouldn't call `get()` on optional without aditional checks. If someone tries to update non existing execution, the service should return 4xx error

petertrr (changes requested):

db/changelog/v-1/04-changeset-test-table.xml
aktsay6: As I remember this one should be hash from two files.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
aktsay6: Added

save-orchestrator/src/main/resources/application-dev.properties
aktsay6: Fixed

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/ExecutionControllerTest.kt
aktsay6: Fixed

save-common/src/commonMain/kotlin/org/cqfn/save/execution/ExecutionUpdateDto.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * @property id
+ * @property status
+ */
+class ExecutionUpdateDto(
+    val id: Long,
+    val status: ExecutionStatus
+) {
+    constructor() : this(0, ExecutionStatus.PENDING)
```
petertrr, line 11: Why is this needed?

petertrr (approved):

petertrr (changes requested):
Fix code style

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
```diff
@@ -11,16 +11,18 @@ import javax.persistence.Enumerated
  * @property endTime
  * @property status
  * @property testSuiteIds
- * @property id
  */
 @Entity
-class Execution(
+data class Execution(
```
petertrr, line 16: I believe we have discussed that data classes shouldn't be used with hibernate?

db/changelog/v-1/04-changeset-test-table.xml
petertrr: Yes, right

save-common/src/commonMain/kotlin/org/cqfn/save/execution/ExecutionUpdateDto.kt
```diff
@@ -0,0 +1,12 @@
+package org.cqfn.save.execution
+
+/**
+ * @property id
+ * @property status
+ */
+data class ExecutionUpdateDto(
```
petertrr, line 7: This should be `@Serializable`

petertrr (approved):

PR #52, opened 2021-03-19
Infrastructure enhancements

### LINKED ISSUE #40: Infra: Improvements for insfrastructure
* junit5 extensions instead of base classes for tests using dynamic properties: https://www.baeldung.com/spring-dynamicpropertysource#an-alternative-test-fixtures
* Add docker volume for grafana configs, store them in `save-deploy`. Configure grafana to work behind nginx
* For services that do not need replication (orchestrator (?), prometheus and grafana) add the following to compose file:
  ```yaml
    deploy:
      placement:
        max_replicas_per_node: 1
        constraints:
          - "node.role==manager"
     ```

petertrr:
### What's done:
* Added grafana.ini
* Updated docker-compose.yaml.template
* Extracted DatabaseTestBase to a junit5 extension
* Enabled management endpoints in preprocessor
* Added prometheus storage
* Add mysql to docker compose only in dev profile
* Fixes for prometheus config
* Dependency updates

Closes #40


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -20,8 +20,23 @@ fun Project.createStackDeployTask(profile: String) {
...
+                        // so we create it here only in dev profile
+                        """|$it
+                           |  mysql:
+                           |    image: mysql:8.0.20
+                           |    ports:
+                           |      - "3306:3306"
+                           |    environment:
+                           |      - "MYSQL_ROOT_PASSWORD=123"
+                           |      - "MYSQL_DATABASE=save_cloud"
+                           |    profiles: ["dev"]
```
aktsay6, line 34: Why do we still need `profiles` string?

petertrr: We don't, I'll remove it.

aktsay6 (approved):
lgtm

PR #66, opened 2021-03-25
Database: Configure retention for tests and test_executions

### LINKED ISSUE #46: Database: Configure retention for tests and test_executions
Should be deleted after 30 days.

Cheshiriks:
### What's done:
  * Created mysql event scheduler

closes #46

### DIFF
--- a/db/changelog/v-1/event_scheduler.xml
+++ b/db/changelog/v-1/event_scheduler.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="event-scheduler-1" author="frolov" context="prod or dev">
+        <sql>
+            CREATE EVENT delete_test
+            ON SCHEDULE EVERY 1 DAY
+            STARTS TIMESTAMP(CURRENT_DATE)
+            DO DELETE FROM TEST WHERE DATEDIFF(NOW(), DATE_ADDED) >= 30;
+        </sql>
+    </changeSet>
+
+    <changeSet id="event-scheduler-2" author="frolov" context="prod or dev">
+        <sql>
+            CREATE EVENT delete_test_execution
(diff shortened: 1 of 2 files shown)


db/changelog/v-1/event_scheduler.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="event-scheduler-1" author="frolov" context="prod or dev">
```
petertrr, line 8: I believe it should only be in prod profile, we don't use a persistent database in dev

db/changelog/v-1/event_scheduler.xml
```diff
@@ -0,0 +1,26 @@
...
+        <sql>
+            CREATE EVENT delete_test
+            ON SCHEDULE EVERY 1 DAY
+            STARTS TIMESTAMP(CURRENT_DATE)
```
petertrr, line 12: Will this be a time of script execution?

Cheshiriks: Yes it's

db/changelog/v-1/event_scheduler.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="event-scheduler-1" author="frolov" context="prod">
```
petertrr, line 8: We should think about a way to keep some sort of execution history even after we remove test executions, maybe store list of failed tests in Execution? Anyway, it's a thought for a separate issue

petertrr (approved):
lgtm

db/changelog/v-1/event_scheduler.xml
Cheshiriks: Yes I agree. We need to create a issue for this.

petertrr: #69

PR #71, opened 2021-03-29
Frontend: Retrieve projects from backend

### LINKED ISSUE #61: Frontend: Populate different views with data from backend
Add queries in
* [ ] history view (also will need to support storing aggregated test results in DB, associated with individual Executions). Also, href from ProjectView, which opens HistoryView on submission, is broken: opens `/history` instead of `#/owner/name/history`
* [x] execution view (#124, #123)
* [x] project dashboard (#72)

This issue can be split into smaller ones if needed.

Queries should take place in `componentDidMount` function and update state

### LINKED ISSUE #72: Frontend: Fetch projects from backed and display with pagination
Part of #61

petertrr:
### What's done:
* Added endpoint to backend
* Added RequestUtils.kt in frontend
* Supported client-side pagination in TableComponent
* Fetch data for CollectionView
* Fixed liquibase contexts in dev profile
* Added new tasks to manage docker: restartMysqlDb, stopDockerStack
* Added charset to headers on backend
* Fixed nunito font link
* Fixed unicode symbols
* Updated build_and_test.yml to upload distribution jar instead of zipping the entire folder

Fixme: Error handling on frontend

Closes #72


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/Footer.kt
```diff
@@ -22,7 +22,7 @@ class Footer : RComponent<RProps, RState>() {
             div("container my-auto") {
                 div("copyright text-center my-auto") {
                     span {
-                        +"Copyright &copy; SAVE 2021"
+                        +"Copyright ${js("String.fromCharCode(169)")} SAVE 2021"
```
kentr0w, line 25: Why it cann't be just string?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/tables/TableComponent.kt
```diff
@@ -110,6 +137,31 @@ fun <D : Any> tableComponent(columns: Array<out Column<D, *>>, getData: () -> Ar
                         }
                     }
                 }
+                if (tableInstance.pageCount > 1) {
```
kentr0w, line 140: Why `> 1`?

kentr0w (changes requested):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/Footer.kt
petertrr: It's an ugly workaround to display special characters like `©`. It should be working with HTML code `&copy;`, but for some reason it doesn't, and I've found this workaround.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/tables/TableComponent.kt
petertrr: Because if there is a single page, we don't need pagination controls.

kentr0w (approved):
lgtm

PR #79, opened 2021-03-30
Bugfix/change cloning logic

### LINKED ISSUE #68: Backend: follow-ups for cloning logic
* /upload - very bad, incorrect handling of ResponseEntity. Should be:
  ```kotlin
               .retrieve()
            .toEntity(String::class.java)
            .toMono()
   ```
   Also, for `toMono` kotlin-reqactor-extensions should be added as suggested in spring docs
* Logging: `cloning null project` - incorrect; `projectService.saveProject` should return id of the project so that it can be used in logging

kentr0w:
### What's done:
* Changed logic
* Added test
* Update build_and_test.yml
* Add jacoco in preprocessor

closes  #68


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -35,21 +38,23 @@ class CloneRepositoryController(
      * @return mono string
      */
     @PostMapping(value = ["/submitExecutionRequest"])
-    fun submitExecutionRequest(@RequestBody executionRequest: ExecutionRequest): Mono<ResponseEntity<*>>? {
+    fun submitExecutionRequest(@RequestBody executionRequest: ExecutionRequest): Response {
         val project = executionRequest.project
+        var projectId: String? = null
```
petertrr, line 45: You can declare it as a non-nullable val, and kotlin compiler allows it because this variable is assigned only once inside `try`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -35,21 +38,23 @@ class CloneRepositoryController(
...
         try {
-            projectService.saveProject(project)
-            log.info("Project ${project.id} saved")
+            projectId = projectService.saveProject(project)?.toString() ?: "With non Id"
```
petertrr, line 47: If project is saved, then an id is assigned to it, so `saveProject` should never return null

petertrr (changes requested):

petertrr (approved):

PR #93, opened 2021-04-16
Stop containers on completion of Execution

### LINKED ISSUE #87: Orchestrator: Terminate containers when everything is done
When 'Everything is done' == there are no tests left in DB, all agents have IDLE status in DB and send corresponding heartbeats, stop all agents for this execution

Originally part of #60 , extracted here for better readability

petertrr:
### What's done:
* Added logic in HeartbeatController
* Added test in orchestrator
* Added jackson-module-kotlin
* Tests in backend
* Fixes for liquibase and test data
* Made some entity relations lazy-loaded
* Changed dependency scope for distribution artifacts to runtimeOnly to remove dependency on build tasks

Closes #87


save-common/src/jvmMain/kotlin/org/cqfn/save/entities/Agent.kt
```diff
@@ -10,11 +10,9 @@ import javax.persistence.ManyToOne
...
 @Entity
 class Agent(
-
     var containerId: String,
 
     @ManyToOne
     @JoinColumn(name = "execution_id")
-    var execution: Execution,
-
+    var execution: Execution?,
```
Cheshiriks, line 17: Why is execution nullable? In DB it has nullable = false and FK

petertrr: That's because when orchestrator updates agent statuses based on heartbeats, it doesn't know `Execution` for the agent. It actually needs to save an `AgentStatus`, finding corresponding `agent.agent_id` based on `agent.container_id`.

petertrr: Created a more correct method for this

save-orchestrator/src/test/kotlin/org/cqfn/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
```diff
@@ -73,6 +93,106 @@ class HeartbeatControllerTest {
...
+     * @param tests a batch of tests returned from backend (mocked response)
+     * @param mockAgentStatuses whether a mocked response for `/getAgentsStatusesForSameExecution` should be added to queue
+     * @param verification a lambda for test assertions
+     */
+    private fun testHeartbeat(
+        agentStatuses: List<AgentStatus>,
+        heartbeat: Heartbeat,
+        tests: List<TestDto>,
+        mockAgentStatuses: Boolean = false,
+        verification: () -> Unit,
```
Cheshiriks, line 159: Do you need to change it to a _agentStatuses: List<AgentStatusDto>_?

petertrr: Yes, updated

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/Agent.kt
petertrr: Also, made `Execution` lazily fetched, because on agent status update we only need agent id from this entity

Cheshiriks (approved):

PR #95, opened 2021-04-19
Infra: Remove dependencies on bintray

petertrr:
### What's done:
* Switched frontend from bintray to kotlin space
* Bump kotlin-js-wrappers version
* Fixed App.kt
* Removed bintray repo for kotlinx.datetime
* Removed jcenter repository from buildSrc and root build.gradle.kts

Bintray will be shut down soon, need to remove it from project dependencies. All of our dependencies have already migrated, so it's safe.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,7 +69,6 @@ talaiot {
 
 allprojects {
     repositories {
-        jcenter()
         mavenCentral()
     }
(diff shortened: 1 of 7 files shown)


Cheshiriks (approved):

petertrr:
we should add this

PR #96, opened 2021-04-19
Add JpaSpecificationExecutor

Cheshiriks:
### What's done:
* Add JpaSpecificationExecutor
* Add test

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
@@ -11,7 +11,6 @@ import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
-import java.time.LocalDateTime
 
 /**
  * Controller to manipulate with Agent related data
@@ -60,10 +59,7 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
             agentRepository.findAll { root, cq, cb ->
                 cb.equal(root.get<Execution>("execution"), getAgentByContainerId(agentId).execution)
             }.map {
-                agentStatusRepository.findAll { root, cq, cb ->
-                    cq.orderBy(cb.desc(root.get<LocalDateTime>("time")))
-                    cb.equal(root.get<Agent>("agent").get<String>("containerId"), it.containerId)
-                }.get(0).toDto()
+                agentStatusRepository.findTopByAgentContainerIdOrderByTimeDesc(it.containerId)?.toDto()
             }
 
     /**
@@ -73,13 +69,9 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
      * @return list of agent statuses
      */
     private fun getAgentByContainerId(containerId: String): Agent {
-        val agent = agentRepository.findOne { root, query, cb ->
+        val agent = agentRepository.findOne { root, _, cb ->
             cb.equal(root.get<String>("containerId"), containerId)
         }
-        if (agent.isEmpty) {
-            error("Agent with containerId=$containerId not found in the DB")
-        } else {
-            return agent.get()
-        }
+        return agent.orElseThrow { IllegalStateException("Agent with containerId=$containerId not found in the DB") }
     }
 }
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
@@ -2,15 +2,16 @@ package org.cqfn.save.backend.controllers
 
 import org.cqfn.save.backend.repository.AgentRepository
 import org.cqfn.save.backend.repository.AgentStatusRepository
-import org.cqfn.save.backend.repository.ExecutionRepository
 import org.cqfn.save.entities.Agent
 import org.cqfn.save.entities.AgentStatus
 import org.cqfn.save.entities.AgentStatusDto
+import org.cqfn.save.entities.Execution
 import org.springframework.transaction.annotation.Transactional
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
+import java.time.LocalDateTime
 
 /**
  * Controller to manipulate with Agent related data
@@ -18,7 +19,6 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 class AgentsController(private val agentStatusRepository: AgentStatusRepository,
                        private val agentRepository: AgentRepository,
-                       private val executionRepository: ExecutionRepository,
 ) {
     /**
      * @param agents list of [Agent]s to save into the DB
@@ -42,8 +42,7 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
     @PostMapping("/updateAgentStatusesWithDto")
     fun updateAgentStatusesWithDto(@RequestBody agentStates: List<AgentStatusDto>) {
         agentStates.forEach {
-            val agent = agentRepository.findByContainerId(it.containerId)
-                ?: error("Agent with containerId=${it.containerId} not found in the DB")
+            val agent = getAgentByContainerId(it.containerId)
             agentStatusRepository.save(AgentStatus(it.time, it.state, agent))
         }
     }
@@ -57,9 +56,30 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
      */
     @GetMapping("/getAgentsStatusesForSameExecution")
     @Transactional
-    fun findAllAgentStatusesForSameExecution(@RequestBody agentId: String): List<AgentStatusDto?> = agentRepository
-        .findByExecutionIdOfContainerId(agentId)
-        .map {
-            agentStatusRepository.findTopByAgentContainerIdOrderByTimeDesc(it.containerId)?.toDto()
+    fun findAllAgentStatusesForSameExecution(@RequestBody agentId: String): List<AgentStatusDto?> =
+            agentRepository.findAll { root, cq, cb ->
+                cb.equal(root.get<Execution>("execution"), getAgentByContainerId(agentId).execution)
+            }.map {
+                agentStatusRepository.findAll { root, cq, cb ->
+                    cq.orderBy(cb.desc(root.get<LocalDateTime>("time")))
+                    cb.equal(root.get<Agent>("agent").get<String>("containerId"), it.containerId)
+                }.get(0).toDto()
+            }
+
+    /**
+     * Get agent by containerId.
+     *
+     * @param containerId containerId of an agent.
+     * @return list of agent statuses
+     */
+    private fun getAgentByContainerId(containerId: String): Agent {
+        val agent = agentRepository.findOne { root, query, cb ->
+            cb.equal(root.get<String>("containerId"), containerId)
         }
+        if (agent.isEmpty) {
+            error("Agent with containerId=$containerId not found in the DB")
+        } else {
+            return agent.get()
+        }
+    }
 }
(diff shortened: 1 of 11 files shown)


save-backend/src/test/kotlin/org/cqfn/save/backend/controller/BaseEntityRepositoryTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.cqfn.save.backend.utils.MySqlExtension
+import org.cqfn.save.entities.Execution
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient
+import org.springframework.boot.test.context.SpringBootTest
+
+@SpringBootTest(classes = [SaveApplication::class])
```
petertrr, line 14: I believe `DataJpaTest` will be enough here

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/BaseEntityRepository.kt
```diff
@@ -1,11 +1,26 @@
...
  * JPA repositories for the subclasses of the BaseEntity
  */
 @NoRepositoryBean
-interface BaseEntityRepository<T : BaseEntity> : JpaRepository<T, Long>
+interface BaseEntityRepository<T : BaseEntity> : JpaRepository<T, Long> {
+    /**
+     * @param clazz
+     * @return list of class objects
+     */
+    fun getList(clazz: KClass<T>?): List<T>
```
petertrr, line 18: What's the purpose of this method? How is it different from `findAll`?

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/BaseEntityRepository.kt
```diff
@@ -1,11 +1,26 @@
...
+
+    /**
+     * @param entityClass
+     * @param specification
+     * @return list of class objects
+     */
+    fun getList(entityClass: KClass<T>?, specification: Specification<T>?): List<T>
```
petertrr, line 25: Why is `entityClass` nullable?

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/BaseEntityRepository.kt
Cheshiriks: You can Specification with any conditions: where, order by etc.

petertrr: Yeah, for the second method it's clear. But this method only accepts a single argument.

Cheshiriks: Yes, removed it

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/BaseEntityRepository.kt
```diff
@@ -1,11 +1,20 @@
...
  */
 @NoRepositoryBean
-interface BaseEntityRepository<T : BaseEntity> : JpaRepository<T, Long>
+interface BaseEntityRepository<T : BaseEntity> : JpaRepository<T, Long> {
+    /**
+     * @param entityClass
+     * @param specification
+     * @return list of class objects
+     */
+    fun getList(entityClass: KClass<T>, specification: Specification<T>): List<T>
```
petertrr, line 19: I believe you can extend `JpaSpecificationExecutor<T>` to be able to configure query using `Specification`. Did you take a look at [this section of docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#specifications)?

Cheshiriks: Yes, I think we can use it :)

petertrr (changes requested):
I believe we don't need `JpaSpecificationExecutor` for all repositories, because it leads to generation of more complex runtime proxies, which we won't use. I suggest to add this interface only when needed; e.g. after we merge #93 you can change one method from native query to `Specification` there

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/JpaSpecificationExecutorRepository.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor
+
+/**
+ * Extension interface for JpaSpecificationExecutor
+ */
+interface JpaSpecificationExecutorRepository<T> : JpaSpecificationExecutor<T> {
+    /**
+     * @param spec Specification.
+     */
+    fun getFirstResultOrNull(spec: Specification<T>) = this.findAll(spec).getOrNull(0)
```
petertrr, line 13: Will it perform `get` operation on client side, with all data fetched from DB? If so, I believe limiting results in DB will be more preferable, e.g. by passing a `Pageable` of single element to `findAll`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -57,9 +56,22 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
      */
     @GetMapping("/getAgentsStatusesForSameExecution")
     @Transactional
-    fun findAllAgentStatusesForSameExecution(@RequestBody agentId: String): List<AgentStatusDto?> = agentRepository
-        .findByExecutionIdOfContainerId(agentId)
-        .map {
-            agentStatusRepository.findTopByAgentContainerIdOrderByTimeDesc(it.containerId)?.toDto()
-        }
+    fun findAllAgentStatusesForSameExecution(@RequestBody agentId: String): List<AgentStatusDto?> = agentRepository.findAll { root, cq, cb ->
+        cb.equal(root.get<Execution>("execution"), getAgentByContainerId(agentId).execution)
```
petertrr, line 60: Is `"execution"` here a name of column or a name of a class property? In latter case, is there any overload for this method for property reference?

petertrr: Also, is there any difference if we compare entire `Execution`s or only their IDs?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -57,9 +56,22 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
+    }.map {
+        agentStatusRepository.getFirstResultOrNull { root, cq, cb ->
```
petertrr, line 62: When we expect a single result, you can just use `findOne(_: Specification)`

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
Cheshiriks: No difference. When building a request, it still uses the ID.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -57,9 +56,30 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
+     * Get agent by containerId.
+     *
+     * @param containerId containerId of an agent.
+     * @return list of agent statuses
+     */
+    private fun getAgentByContainerId(containerId: String): Agent {
+        val agent = agentRepository.findOne { root, query, cb ->
+            cb.equal(root.get<String>("containerId"), containerId)
+        }
+        if (agent.isEmpty) {
```
petertrr, line 79: You can use `orElseThrow` instead of `if`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -57,9 +56,30 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
+            }
+
+    /**
+     * Get agent by containerId.
+     *
+     * @param containerId containerId of an agent.
+     * @return list of agent statuses
+     */
+    private fun getAgentByContainerId(containerId: String): Agent {
+        val agent = agentRepository.findOne { root, query, cb ->
```
petertrr, line 76: You can substitute unused parameters with `_`, as kotlin compiler ususally suggests

petertrr: Also, do you think there is a benefir in using criteria API here instaead of `findByContainderId` method?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -57,9 +56,30 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
-        .map {
-            agentStatusRepository.findTopByAgentContainerIdOrderByTimeDesc(it.containerId)?.toDto()
+    fun findAllAgentStatusesForSameExecution(@RequestBody agentId: String): List<AgentStatusDto?> =
+            agentRepository.findAll { root, cq, cb ->
+                cb.equal(root.get<Execution>("execution"), getAgentByContainerId(agentId).execution)
+            }.map {
+                agentStatusRepository.findAll { root, cq, cb ->
+                    cq.orderBy(cb.desc(root.get<LocalDateTime>("time")))
+                    cb.equal(root.get<Agent>("agent").get<String>("containerId"), it.containerId)
+                }.get(0).toDto()
```
petertrr, line 66: You should limit the number of query results inside the query, not using clinet-side `get(0)`

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -59,4 +59,17 @@ class AgentsControllerTest {
                 Assertions.assertEquals(AgentState.BUSY, statuses[1]!!.state)
             }
     }
+
+    @Test
+    fun `should return error not found container id`() {
```
petertrr, line 64: I thought you'd add a test in `JpaSpecificationTest`, just to test that a specific exception is thrown

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
Cheshiriks: I think we do not need to create methods in the repository classes that are suitable for only one entity class

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/JpaSpecificationTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.cqfn.save.backend.utils.MySqlExtension
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest
+
+@DataJpaTest
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
```
petertrr, line 14: I think this annotation is redundant

petertrr (approved):

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/JpaSpecificationTest.kt
petertrr: Sorry, my bad, it is actually required to use mysql

PR #99, opened 2021-04-21
Feature. Store time of first and last heartbeat with a particular state

### LINKED ISSUE #92: Backend & Database: Store time of first and last heartbeat with a particular state
E.g.,
13:15 - IDLE
13:30 - IDLE
13:45 - RUNNING

Should produce 2 and not 3 entries in the DB

So in response for heartbeats an existing entry should be updated instead of a new one

aktsay6:
### What's done:
  * Added feature with updating states

This PR closes #92


petertrr (commented):
AgentStatus should have two fields for time: start and end of a particular status. It is redundant for the most part, but for the very first heartbeat we will otherwise lose information about it's start time

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -41,10 +41,17 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
     @PostMapping("/updateAgentStatusesWithDto")
     fun updateAgentStatusesWithDto(@RequestBody agentStates: List<AgentStatusDto>) {
-        agentStates.forEach {
-            val agent = agentRepository.findByContainerId(it.containerId)
-                ?: error("Agent with containerId=${it.containerId} not found in the DB")
-            agentStatusRepository.save(AgentStatus(it.time, it.state, agent))
+        agentStates.forEach { dto ->
+            val agent = agentRepository.findByContainerId(dto.containerId)
+                ?: error("Agent with containerId=${dto.containerId} not found in the DB")
+            val agentStatus = agentStatusRepository.findTopByAgentIdAndStateOrderByTimeDesc(agent.id!!, dto.state)
```
petertrr, line 47: Can we use `findTopByAgentContainerIdAndStateOrderByTimeDesc` instead and not make a query to get `agent`?

petertrr, line 47:
```
IDLE
BUSY
IDLE
```
For this data the query with `dto.state == BUSY` will return the second entry and update it's timestamp, won't it? This is incorrect, we should get the last entry and compare its state with current

petertrr (changes requested):

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -16,14 +18,19 @@ import org.springframework.test.web.reactive.server.WebTestClient
 import org.springframework.test.web.reactive.server.expectBody
 import org.springframework.web.reactive.function.BodyInserters
 import java.time.LocalDateTime
+import javax.transaction.Transactional
 
 @SpringBootTest(classes = [SaveApplication::class])
 @AutoConfigureWebTestClient
 @ExtendWith(MySqlExtension::class)
+@Transactional
```
petertrr, line 26: Is it still needed? Also, I believe it's better to place Transactional on controller itself (though, it needs manual/integration testing)

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -40,9 +40,17 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
      */
     @PostMapping("/updateAgentStatusesWithDto")
     fun updateAgentStatusesWithDto(@RequestBody agentStates: List<AgentStatusDto>) {
-        agentStates.forEach {
-            val agent = getAgentByContainerId(it.containerId)
-            agentStatusRepository.save(AgentStatus(it.time, it.state, agent))
+        agentStates.forEach { dto ->
+            val agent = getAgentByContainerId(dto.containerId)
```
petertrr, line 44: This can be moved to the second if bracnh

db/changelog/v-1/sqlRequests/agent-status.csv
```diff
@@ -1,10 +1,10 @@
...
-3;"2021-01-01 00:00:30";"BUSY";1
-4;"2021-01-01 00:00:45";"BUSY";1
-5;"2021-01-01 00:01:00";"BUSY";1
-6;"2021-01-01 00:01:15";"BUSY";1
-7;"2021-01-01 00:01:30";"FINISHED";1
-8;"2021-01-01 00:01:45";"IDLE";1
-12;"2021-01-01 01:00:00";"BUSY";2
\ No newline at end of file
+id;start_time;end_time;state;agent_id
+1;"2021-01-01 00:00:00";"2021-01-01 00:00:00";"IDLE";1
```
petertrr, line 2: I believe end time should be at least 15 seconds further in time than start; better make these intervals different throughout test data

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/AgentStatus.kt
```diff
@@ -32,16 +34,18 @@ class AgentStatus(
...
- * @property time time of update
+ * @property startTime staring time of status
+ * @property endTime time of update
  * @property state current state of the agent
  * @property containerId id of the agent's container
  */
 class AgentStatusDto(
-    val time: LocalDateTime,
+    val startTime: LocalDateTime,
+    val endTime: LocalDateTime,
```
petertrr, line 48: It doesn't make much sense for me to store both timestamps here. Dto is a thing that orchestrator send to save information on a current heartbeat; start_time should be inserted in DB on state change and not kept on orchestrator

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -33,12 +40,65 @@ class AgentsControllerTest {
...
+                    AgentStatusDto(LocalDateTime.now(), LocalDateTime.now(), AgentState.BUSY, "container-2")
                 )
             ))
             .exchange()
             .expectStatus()
             .isOk
+
+        assertTrue(
+            agentStatusRepository
+                .findAll()
```
petertrr, line 98: Please also add an assertion that first two IDLE statuses are in the same entry

petertrr (changes requested):

db/changelog/v-1/sqlRequests/agent-status.csv
```diff
@@ -1,10 +1,10 @@
...
-6;"2021-01-01 00:01:15";"BUSY";1
-7;"2021-01-01 00:01:30";"FINISHED";1
-8;"2021-01-01 00:01:45";"IDLE";1
-12;"2021-01-01 01:00:00";"BUSY";2
\ No newline at end of file
+id;start_time;end_time;state;agent_id
+1;"2021-01-01 00:00:00";"2021-01-01 00:00:15";"IDLE";1
+2;"2021-01-01 00:00:15";"2021-01-01 00:00:30";"IDLE";1
+3;"2021-01-01 00:00:30";"2021-01-01 00:00:45";"BUSY";1
+4;"2021-01-01 00:00:45";"2021-01-01 00:01:00";"BUSY";1
```
petertrr, line 5: Please make these intervals more diverse, i.e. some should be greater than 15 sec

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
```diff
@@ -39,10 +40,19 @@ class AgentsController(private val agentStatusRepository: AgentStatusRepository,
...
+        agentStates.forEach { dto ->
+            val agentStatus = agentStatusRepository.findTopByAgentContainerIdOrderByEndTimeDesc(dto.containerId)
+            if (agentStatus != null && agentStatus.state == dto.state) {
+                // updating time
+                agentStatus.endTime = dto.time
+                agentStatusRepository.save(agentStatus)
+            } else {
+                // insert new agent status
+                val agent = getAgentByContainerId(dto.containerId)
+                agentStatusRepository.save(AgentStatus(LocalDateTime.now(), dto.time, dto.state, agent))
```
petertrr, line 54:
```suggestion
                agentStatusRepository.save(AgentStatus(dto.time, dto.time, dto.state, agent))
```
It makes more sense to use timestamps of status receiving as initially provided by agent

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
petertrr: https://github.com/cqfn/save-cloud/pull/99/files#r617588302

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -41,6 +49,72 @@ class AgentsControllerTest {
...
+            ))
+            .exchange()
+            .expectStatus()
+            .isOk
+
+        val firstIdle = agentStatusRepository
+            .findAll()
+            .first { it.state == AgentState.IDLE && it.agent.containerId == "container-2" }
+
+        webTestClient
```
petertrr, line 73: Please extract this to a utility function for this test

petertrr (changes requested):

db/changelog/v-1/sqlRequests/agent-status.csv
aktsay6: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/AgentsController.kt
aktsay6: Fixed

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -41,6 +64,46 @@ class AgentsControllerTest {
             .isOk
     }
 
+    @Test
+    @Suppress("TOO_LONG_FUNCTION")
+    fun `check that agent statuses are updated`() {
+        updateAgentStatuses(
+            listOf(
+                AgentStatusDto(LocalDateTime.now(), AgentState.IDLE, "container-2")
```
petertrr, line 72: Actually, please use fixed dates here too. Once we get to Jun 2022, this test will start failing :)

petertrr (approved):
LGTM, but please fix dates in tests before proceeding

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/AgentsControllerTest.kt
aktsay6: Fixed

PR #100, opened 2021-04-23
Bugfix. In TestService, offset should be stored per Execution

### LINKED ISSUE #97: Backend: In TestService, offset should be stored per Execution
Also, batch of tests should be retrieved for a particular Execution

aktsay6:
### What's done:
  * Fixed bug with last test in TestInitializeControllerTest
  * Fixed bug with offset
This PR closes #97


db/changelog/v-1/10-changeset-project-table-insert.xml
```diff
@@ -14,6 +14,14 @@
             <column name="url" value="huaweiUrl" />
             <column name="description" value="test description" />
         </insert>
+        <insert dbms="mysql" tableName="project">
+            <column name="id" value="2" />
+            <column name="owner" value="Huawei" />
+            <column name="name" value="huaweiName 2" />
+            <column name="type" value="manual" />
+            <column name="url" value="huaweiUrl 2" />
```
petertrr, line 22: Let's use test data that looks like actual. You can always use `example.com` instead of actual urls, but randomly formatted string instead of url can cause some trouble if we use it in tests in the future

db/changelog/v-1/11-changeset-test-agent-insert-data.xml
```diff
@@ -29,6 +29,19 @@
...
             <column name="test_suite_ids" value="1, 2, 3" />
             <column name="resources_root_path" value="resources_root_path" />
+            <column name="offset" value="0" />
+            <column name="execution_limit" value="20" />
+        </insert>
+        <insert dbms="mysql" tableName="execution">
+            <column name="id" value="2" />
+            <column name="project_id" value="2" />
+            <column name="start_time" value="2021-01-01 00:00:00" />
+            <column name="end_time" value="2021-02-01 00:00:00" />
```
petertrr, line 39: Can a running execution have end_tie?

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ExecutionRepository.kt
```diff
@@ -1,10 +1,20 @@
...
  */
 @Repository
-interface ExecutionRepository : BaseEntityRepository<Execution>
+interface ExecutionRepository : BaseEntityRepository<Execution> {
+    /**
+     * @param offset
+     * @param id
+     */
+    @Modifying
+    @Query("update execution set execution.offset = ?1 where execution.id = ?2", nativeQuery = true)
```
petertrr, line 18: We were trying to use `Sepcification`-based queries instead of native sql, have you considered using it here?

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
```diff
@@ -14,7 +14,9 @@ import javax.persistence.ManyToOne
  * @property endTime
  * @property status
  * @property testSuiteIds
+ * @property offset
  * @property resourcesRootPath path to test resources, relative to shared volume mount point
+ * @property executionLimit
```
petertrr, line 19: Pls write a couple of words here, this name is not obvious

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ExecutionRepository.kt
Cheshiriks: The save () and saveAll() methods update the entity in the database.

db/changelog/v-1/10-changeset-project-table-insert.xml
aktsay6: Fixed

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
aktsay6: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ExecutionRepository.kt
aktsay6: No, but now it's remade with typical save, so the method is not needed anymore

db/changelog/v-1/11-changeset-test-agent-insert-data.xml
aktsay6: As we are not considering end time in running state, the column can not be null

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -27,13 +36,21 @@ class TestService(private val configProperties: ConfigProperties) {
...
+        val agent = agentRepository.findByContainerId(agentId) ?: error("The specified agent does not exist")
+        log.debug("Agent found: $agent")
+        val execution = agent.execution
+        log.debug("Retrieving tests")
+        val tests = testRepository.retrieveBatches(execution.executionLimit, execution.offset, execution.id!!).map {
             TestDto(it.filePath, it.testSuite.id!!, it.id!!)
         }
-        offset.addAndGet(configProperties.limit)
+        log.debug("Increasing offset of execution - ${agent.execution}")
+        execution.offset += execution.executionLimit
```
petertrr, line 52: Let's call it `batchSize`; this name represents its purpose much better and is not bound to DB

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
```diff
@@ -14,8 +14,10 @@ interface TestRepository : BaseEntityRepository<Test> {
      *
      * @param limit
      * @param offset
+     * @param executionId
      * @return List of Tests
      */
-    @Query(value = "select * from test inner join test_execution on test.id = test_execution.test_id and test_execution.status = 'READY' limit ?1 offset ?2", nativeQuery = true)
-    fun retrieveBatches(limit: Int, offset: Int): List<Test>
+    @Query(value = "select * from test inner join test_execution on test.id = test_execution.test_id and test_execution.status = 'READY' " +
```
petertrr, line 20: What about this giant query? I think we already have entities with proper relations, so all this joins can be generated by hibernate. And there are [several ways](https://www.baeldung.com/jpa-limit-query-results) to control result set size

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
aktsay6: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
aktsay6: Deleted this method and added method to TestExecutionRepository.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
petertrr: I actually meant `executionLimit`, not `offset` :) Offset is fine to keep, but execution limit actually means batch size; the word ``limit` appeared only because this parameter was specific to database queries. And when we will be scaling this parameter. we will be retrieving batch of different size starting from the same offset.

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestExecutionRepository.kt
```diff
@@ -1,10 +1,20 @@
...
 @Repository
-interface TestExecutionRepository : BaseEntityRepository<TestExecution>
+interface TestExecutionRepository : BaseEntityRepository<TestExecution> {
+    /**
+     * @param status
+     * @param id
+     * @param pageable
+     * @return list of test executions
+     */
+    fun findAllByStatusEqualsAndTestSuiteExecutionIdEquals(status: TestResultStatus, id: Long, pageable: Pageable): List<TestExecution>
```
petertrr, line 19: I think, `equals` is the default behavior for spring-data-jpa, so it can be omitted. Also, `findBy...` with return type of collection should be equivalent to `findAllBy...`

petertrr (changes requested):
Please rename parameters as described in https://github.com/cqfn/save-cloud/pull/100#discussion_r620064295

petertrr (approved):

PR #101, opened 2021-04-23
Preprocessor test initializing

### LINKED ISSUE #13: Test-preprocessor: Create a stub for tests preprocessing
* Backend makes POST request to preprocessor to clone repo (#62 )
* Preprocessor generates test IDs and POSTs them to backend
* Backend writes IDs into DB

kentr0w:
### What's done:
Created logic for test initializing
Fixed tests
Fixed some fields and data classes
Closes #13


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
```diff
@@ -16,12 +18,16 @@ class TestController {
...
+
     /**
      * @param tests
      */
     @PostMapping("/initializeTests")
-    fun initializeTests(@RequestBody tests: List<Test>) {
-        testService.saveTests(tests)
+    fun initializeTests(@RequestBody testDtos: List<TestDto>) {
+        testService.saveTests(testDtos)
+        testExecutionService.saveTestExecution(testDtos)
```
petertrr, line 30: You can pass ids of saved tests from the first method into the second to avoid additional select

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -19,19 +22,28 @@ class TestService(private val configProperties: ConfigProperties) {
...
+
     /**
      * @param tests
      */
-    fun saveTests(tests: List<Test>) {
-        testRepository.saveAll(tests)
+    fun saveTests(tests: List<TestDto>) {
+        tests.map { testDto ->
+            testSuiteRepository.findById(testDto.testSuiteId).ifPresent {
+                testRepository.findByHash(testDto.hash).ifPresentOrElse(
```
petertrr, line 34: This should be the outer lambda; if you don't need to make a query to `testSuiteRepository` if the test is present, why make it unconditionally?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -19,19 +22,28 @@ class TestService(private val configProperties: ConfigProperties) {
...
+                    {},
+                    { testRepository.save(Test(testDto.hash, testDto.filePath, LocalDateTime.now(), it)) })
```
petertrr, line 36: Is `it` parameter of the outer lamda here? Please use named parameters, it's not really obvious

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -19,19 +22,28 @@ class TestService(private val configProperties: ConfigProperties) {
...
+            }
+        }
     }
 
     /**
      * @return Test batches
      */
-    fun getTestBatches(): Mono<List<TestDto>> {
+    fun getTestBatches(): Mono<List<TestBatchDto>> {
```
petertrr, line 44: The name `TestBatchDto` assumes that the object itself contains some sort of batch, but that's not true here

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/FileUtil.kt
```diff
@@ -0,0 +1,10 @@
+package org.cqfn.save.preprocessor.utils
+
+import java.io.File
+import java.math.BigInteger
+import java.security.MessageDigest
+
+fun File.toHash(): String {
+    val md = MessageDigest.getInstance("SHA-256")
+    return BigInteger(1, md.digest(this.readBytes())).toString(16)
```
petertrr, line 9: I suggest something like [this](https://stackoverflow.com/questions/304268/getting-a-files-md5-checksum-in-java) in order not to read the whole file into memory

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -37,4 +49,17 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
         }
         return lostTests
     }
+
+    fun saveTestExecution(testDtos: List<TestDto>) {
+        testDtos.map { testDto ->
+            testRepository.findByHash(testDto.hash).ifPresentOrElse(
+                { test ->
+                    testSuiteRepository.findById(testDto.testSuiteId).ifPresent {
+                        testExecutionRepository.save(TestExecution(test, testDto.testSuiteId, null, it.project!!, TestResultStatus.READY, null, null))
```
petertrr, line 58: We don't actualy need a `Project` field in `TestExecution`, I think it's a bug after refactorings. We already have a `Test`, that has `TestSuite`, that has `Project`. So please remove it from `TestExecution`. Then you won't need to make an additional query into testSuites table

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
petertrr, line 36: Also, we aoready have an id of test suite in dto; in table we have a column for id for mapping, so we don't need a whole TestSuite in order to save `Test`

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -115,17 +120,61 @@ class DownloadProjectController(val configProperties: ConfigProperties) {
...
+                                .body(BodyInserters.fromValue(execution.also { it.id = execId }))
+                                .retrieve()
+                                .toEntity(HttpStatus::class.java)
+                                .subscribe()
+                        }.subscribe()
+                }.subscribe()
+            }.subscribe()
+    }
+
+    private fun getAllTestSuits(project: Project): List<TestSuiteDto> {
```
petertrr, line 167:
```suggestion
    private fun getAllTestSuites(project: Project): List<TestSuiteDto> {
```

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -115,17 +120,61 @@ class DownloadProjectController(val configProperties: ConfigProperties) {
...
+        return listOf(TestSuiteDto(TestSuiteType.PROJECT, "test", project))
+    }
+
+    private fun getAllTests(path: String, testSuites: List<TestSuite>): List<TestDto> {
```
petertrr, line 171: Please add a todo comment marking that it's a stub

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
```diff
@@ -17,12 +18,16 @@ class TestController {
...
+
     /**
-     * @param tests
+     * @param testDtos
      */
     @PostMapping("/initializeTests")
-    fun initializeTests(@RequestBody tests: List<Test>) {
-        testService.saveTests(tests)
+    fun initializeTests(@RequestBody testDtos: List<TestDto>) {
+        val testsId = testService.saveTests(testDtos)
```
petertrr, line 29:
```suggestion
        val testIds = testService.saveTests(testDtos)
```
This will mark this variable as a collection more clearly

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -34,19 +39,34 @@ class TestService(private val configProperties: ConfigProperties) {
...
+
     /**
      * @param tests
+     * @return list tests id's
      */
-    fun saveTests(tests: List<Test>) {
-        testRepository.saveAll(tests)
+    fun saveTests(tests: List<TestDto>): List<Long> {
+        val testsId: MutableCollection<Long> = mutableListOf()
+        tests.map { testDto ->
```
petertrr, line 51: You should either use `forEach`, because mapped list is useless here, or return `id` instead of adding it to a mutable list

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -34,19 +39,34 @@ class TestService(private val configProperties: ConfigProperties) {
...
+            testRepository.findByHash(testDto.hash)?.let { testsId.add(it.id!!) } ?: run {
+                testSuiteRepository.findById(testDto.testSuiteId).ifPresent { testSuite ->
+                    Test(testDto.hash, testDto.filePath, LocalDateTime.now(), testSuite).run {
```
petertrr, line 54: This is the same thing that I talked about: you are making an additional query to get test suite, even though it's not updated and we only need it's id in `test` table. Please check generated SQL in logs, whether hibernate optimizes this somehow or just makes all the queries.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestSuitesController.kt
```diff
@@ -16,10 +18,10 @@ class TestSuitesController {
...
+     * @param testSuiteDtos
+     * @return mono list of TestSuite
      */
-    @PostMapping("/saveTestSuite")
-    fun saveTestSuite(@RequestBody testSuite: TestSuite) {
-        testSuitesService.saveTestSuite(testSuite)
-    }
+    @PostMapping("/saveTestSuites")
+    fun saveTestSuite(@RequestBody testSuiteDtos: List<TestSuiteDto>): Mono<List<TestSuite>> =
+            Mono.just(testSuitesService.saveTestSuite(testSuiteDtos))
```
petertrr, line 26: This line is marked as not covered with tests

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/TestInitializeControllerTest.kt
```diff
@@ -49,15 +52,17 @@ class TestInitializeControllerTest {
             .exchange()
             .expectStatus()
             .isOk
+
+        assertTrue(testInitRepository.findByHash("newHash") != null)
```
petertrr, line 56: You can use `assertNotNull`

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/TestInitializeControllerTest.kt
```diff
@@ -104,10 +109,10 @@ class TestInitializeControllerTest {
             .exchange()
             .expectStatus()
             .isOk
-            .expectBody<List<TestDto>>()
+            .expectBody<List<TestDtoForBatch>>()
             .consumeWith {
                 println(it.responseBody)
-                assertTrue(it.responseBody!!.isNotEmpty() && it.responseBody!!.size == 1)
+                assertTrue(it.responseBody!!.isNotEmpty())
```
petertrr, line 115: Why did you remove an assertion on size?

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/TestSuitesControllerTest.kt
```diff
@@ -36,39 +40,48 @@ class TestSuitesControllerTest {
...
 
         webClient.post()
             .uri("/saveTestSuite")
             .contentType(MediaType.APPLICATION_JSON)
-            .body(BodyInserters.fromValue(testSuite))
+            .body(BodyInserters.fromValue(listOf(testSuite)))
             .exchange()
-            .expectStatus()
-            .isOk
+            .toMono()
```
petertrr, line 54: In tests you don't need to work with mono, you can call `expectBody` on `ResponseSpec` itself

save-common/src/commonMain/kotlin/org/cqfn/save/test/TestDtoForBatch.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import kotlinx.serialization.Serializable
+
+/**
+ * @property filePath
+ * @property testSuiteId
+ * @property id
+ */
+@Serializable
+data class TestDtoForBatch(
```
petertrr, line 11: Why do we need two different dtos for tests? Are we using an id somewhere when dealing with batches?

save-common/src/commonMain/kotlin/org/cqfn/save/testsuite/TestSuiteDto.kt
```diff
@@ -0,0 +1,17 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * @property type
+ * @property name
+ * @property project
+ */
+@Serializable
+data class TestSuiteDto(
+    var type: TestSuiteType,
```
petertrr, line 14: Why are these fields mutable? I believe for dtos we are not changing them after object creation?

save-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
```diff
@@ -29,17 +28,13 @@ class TestExecution(
 
     @ManyToOne
     @JoinColumn(name = "agent_id")
-    var agent: Agent,
-
-    @ManyToOne
-    @JoinColumn(name = "project_id")
-    var project: Project,
+    var agent: Agent?,
```
petertrr, line 31: Please add in kdocs that this field is nullable because agents are not created yet by the moment of test execution creation

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -91,34 +97,84 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
-                    .subscribe()
+                    .onStatus({status -> status != HttpStatus.OK }) { clientResponse ->
+                        log.error("Backend internal error: ${clientResponse.statusCode()}")
+                        throw ResponseStatusException(
+                            HttpStatus.INTERNAL_SERVER_ERROR,
+                            "Backend internal error"
+                        )
+                    }
+                    .bodyToMono<List<TestSuite>>()
+                    .doOnNext { testSuiteList ->
```
petertrr, line 139: Please extract these callbacks into separate methods (e.g., returning `Mono`, so that you can call `doOnNext` on returned value)

petertrr (changes requested):

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/TestInitializeControllerTest.kt
kentr0w: I retured it back after @aktsay6 fixed it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestSuitesController.kt
kentr0w: There are two tests (`testConnection`, `checkDataSave`) for this request.

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/FileUtil.kt
```diff
@@ -0,0 +1,21 @@
...
+
+private const val RADIX = 16
+
+/**
+ * @return hash of file content
+ */
+fun File.toHash(): String {
+    val md = MessageDigest.getInstance("MD5")
+    Files.newInputStream(Paths.get(this.path)).use { inputSrteam -> DigestInputStream(inputSrteam, md) }
+    return BigInteger(1, md.digest()).toString(RADIX)
```
petertrr, line 20: Pls add a test, where hash is calculated

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -91,34 +99,68 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
     private fun sendToBackendAndOrchestrator(project: Project, path: String) {
         val execution = Execution(project, LocalDateTime.now(), LocalDateTime.now(),
             ExecutionStatus.PENDING, "1", path, 0, configProperties.executionLimit)
+        var execId: Long
         log.debug("Knock-Knock Backend")
-        webClientBackend
+        makeRequest(BodyInserters.fromValue(execution), "/createExecution") { it.bodyToMono(Long::class.java) }
+            .doOnNext { executionId ->
+                execId = executionId
+                makeRequest(BodyInserters.fromValue(getAllTestSuites(project)), "/saveTestSuites") { it.bodyToMono<List<TestSuite>>() }
```
petertrr, line 116: Good, looks much cleaner with `makeRequest`

petertrr (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/FileUtil.kt
kentr0w: It wasn't covered, because passed wrong path to directory, now it is covered by test

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
kentr0w: Thanks!

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/FileUtil.kt
petertrr: Orchestrator should receive relative path; please change walking method to be called with absolute, if necessary

kentr0w: Fixed

petertrr (approved):
LGTM

PR #108, opened 2021-05-14
Orchestrator save logs

### LINKED ISSUE #104: Orchestrator: Implement logic for /executionLogs endpoint

kentr0w:
### What's done:
- Created logic to save aget's logs
- Closes #104

### DIFF
--- a/docker-compose.yaml.template
+++ b/docker-compose.yaml.template
@@ -13,6 +13,7 @@ services:
     volumes:
       - "/var/run/docker.sock:/var/run/docker.sock"
       - save:/home/cnb
+      - /home/saveu/agentLogs:/home/cnb/agentLogs/
     deploy:
       labels:
         - "prometheus-job=save-orchestrator"
--- a/docker-compose.yaml.template
+++ b/docker-compose.yaml.template
@@ -13,7 +13,7 @@ services:
     volumes:
       - "/var/run/docker.sock:/var/run/docker.sock"
       - save:/home/cnb
-      - /home/saveu/agentLogs:/home/cnb/agentLogs/
+      - /home/saveu/executionLogs:/home/cnb/executionLogs/
     deploy:
       labels:
         - "prometheus-job=save-orchestrator"
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
@@ -15,7 +15,7 @@ import java.io.File
  * @property testResources configuration for test resources
  * @property docker configuration for docker API
  * @property agentsCount a number of agents to start for every [Execution]
- * @property agentLogsRelativePath  relative path to folder to store agent logs
+ * @property agentLogsRelativePath relative path to folder to store agent logs
  */
 @ConstructorBinding
 @ConfigurationProperties(prefix = "orchestrator")
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
@@ -24,7 +24,7 @@ data class ConfigProperties(
     val testResources: TestResources,
     val docker: DockerSettings,
     val agentsCount: Int,
-    val agentLogsRelativePath: String,
+    private val agentLogsRelativePath: String,
 ) {
     /**
      * Correct path to agents logs folder from root dir
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
@@ -6,6 +6,7 @@ package org.cqfn.save.orchestrator.config
 
 import org.springframework.boot.context.properties.ConfigurationProperties
 import org.springframework.boot.context.properties.ConstructorBinding
+import java.io.File
 
 /**
  * Class for properties
@@ -14,7 +15,7 @@ import org.springframework.boot.context.properties.ConstructorBinding
  * @property testResources configuration for test resources
  * @property docker configuration for docker API
  * @property agentsCount a number of agents to start for every [Execution]
- * @property agentLogs path to folder to store agent logs
+ * @property agentLogsRelativePath  relative path to folder to store agent logs
  */
 @ConstructorBinding
 @ConfigurationProperties(prefix = "orchestrator")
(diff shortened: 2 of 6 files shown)


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/AgentsController.kt
```diff
@@ -67,6 +73,22 @@ class AgentsController {
         dockerService.stopAgents(agentIds)
     }
 
+    /**
+     * @param executionLogs ExecutionLogs
+     */
+    @PostMapping("/executionLogs")
+    fun saveAgentsLog(@RequestBody executionLogs: ExecutionLogs) {
+        val logFile = File(configProperties.agentLogs + File.separator + "${executionLogs.agentId}.log")
```
petertrr, line 81: Maybe, we can simply create a dedicated slf4j logger, configure rolling file policy for it and write received events through it, instead of manually configuring everything?

save-orchestrator/src/main/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 orchestrator.backendUrl=http://backend:5000
+orchestrator.agentLogs=/cnb/agentLogs
```
petertrr, line 2: That's a weird path

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/AgentsController.kt
petertrr: Ok, it would be too complicated to configure this for every agent

save-orchestrator/src/test/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 orchestrator.backendUrl=http://backend:5000
+orchestrator.agentLogs=/cnb/agentLogs
```
petertrr, line 2:
```suggestion
orchestrator.agentLogsRelativePath=/cnb/agentLogs
```

petertrr (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/config/ConfigProperties.kt
```diff
@@ -22,7 +23,13 @@ data class ConfigProperties(
     val testResources: TestResources,
     val docker: DockerSettings,
     val agentsCount: Int,
-)
+    val agentLogs: String,
```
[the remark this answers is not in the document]
kentr0w: Yes, it is

save-orchestrator/src/main/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 orchestrator.backendUrl=http://backend:5000
+orchestrator.agentLogsRelativePath=cnb/agentLogs
```
petertrr, line 2: `user.home` will be `/home/cnb`, so full path will be `/home/cnb/cnb/agentLogs`?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/AgentsController.kt
```diff
@@ -67,6 +73,25 @@ class AgentsController {
...
+    fun saveAgentsLog(@RequestBody executionLogs: ExecutionLogs) {
+        val logDir = File(configProperties.agentLogsFolder)
+        if (!logDir.exists()) {
+            log.info("Folder to store logs from agents was created: ${logDir.name}")
+            logDir.mkdirs()
+        }
+        val logFile = File(logDir.path + File.separator + "${executionLogs.agentId}.log")
+        if (!logFile.exists()) {
+            logFile.createNewFile()
+            log.info("File for ${executionLogs.agentId} agent was created")
```
petertrr, line 91:
```suggestion
            log.info("Log file for ${executionLogs.agentId} agent was created")
```

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/AgentsController.kt
```diff
@@ -67,6 +73,25 @@ class AgentsController {
...
+        }
+        logFile.appendText(executionLogs.cliLogs.joinToString(separator = System.lineSeparator(), postfix = System.lineSeparator()))
+        log.info("Logs were wrote")
```
petertrr, line 94: Which logs?

save-orchestrator/src/main/resources/application.properties
kentr0w: Now, it will `/home/{username}/cnb/agentLogs`

petertrr: Yes, and in our normal environment, inside docker, username will be cnb

petertrr (changes requested):
Please also add a docker volume, mounted to the host file system, where logs will be stored

docker-compose.yaml.template
```diff
@@ -13,6 +13,7 @@ services:
     volumes:
       - "/var/run/docker.sock:/var/run/docker.sock"
       - save:/home/cnb
+      - /home/saveu/agentLogs:/home/cnb/agentLogs/
```
petertrr, line 16: Do you see, how the value from `application.properties` leads to inconsistency, as discussed in https://github.com/cqfn/save-cloud/pull/108#discussion_r634480183?

save-orchestrator/src/test/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 orchestrator.backendUrl=http://backend:5000
+orchestrator.agentLogsRelativePath=cnb/agentLogs
 orchestrator.testResources.basePath=/home/cnb/repositories
```
petertrr, line 3: Why do we have an absolute path for repositories, but can't for agent logs?

petertrr (changes requested):

save-orchestrator/src/test/resources/application.properties
kentr0w: Because we override this path. I fixed

docker-compose.yaml.template
kentr0w: Fixed

save-orchestrator/src/main/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 orchestrator.backendUrl=http://backend:5000
+orchestrator.agentLogs=/home/cnb/agentLogs/
```
petertrr, line 2: One last thing - let's rename these directories and call them `executionLogs`, similar to endpoint name. Because the logs won't be from agent, they'll be logs from CLI, provided by agent; and with current folder name it would be confusing.

petertrr (changes requested):

save-orchestrator/src/main/resources/application.properties
kentr0w: Ok, renamed

petertrr (approved):

PR #109, opened 2021-05-17
Save agent version

### LINKED ISSUE #82: Agent: send additional information in the first heartbeat
id, version

kentr0w:
### What's done:
- Created endpoint to save agent version
- Added request in agent to save agent
- Changed field in entities
- Closes #82


db/changelog/v-1/07-changeset-agent-table.xml
```diff
@@ -21,6 +21,9 @@
             <column name="execution_id" type="bigint">
                 <constraints foreignKeyName="fk_agent_execution" references="execution(id)" nullable="false"/>
             </column>
+            <column name="version" type="varchar(10)">
```
petertrr, line 24: 10 is too little, let's make it at least 64

save-agent/build.gradle.kts
```diff
@@ -85,6 +85,24 @@ kotlin {
     }
 }
 
+val agentVersion = "0.0.1"
```
petertrr, line 88: Why? Use project's version

save-agent/src/nativeMain/resources/agent.properties
```diff
@@ -4,4 +4,5 @@ orchestratorUrl=http://host.docker.internal:5100
 heartbeat.interval=15000
 requestTimeoutMillis=1000
 executionDataRetryAttempts=5
-executionDataInitialRetryMillis=1000
\ No newline at end of file
+executionDataInitialRetryMillis=1000
+agentVersion=0.0.1
```
petertrr, line 8: Version shouldn't be set here, it should be added dynamically

petertrr (changes requested):

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -140,6 +141,15 @@ class SaveAgent(private val config: AgentConfiguration,
         body = executionLogs
     }
 
+    /**
+     * Save agent version
+     */
+    private suspend fun sendAgentVersion() = httpClient.post<HttpResponse> {
```
petertrr, line 147: if we are keeping it as a separate query, please change it to something like `saveAdditionalData`, because we may add other info in there

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -68,6 +69,9 @@ class SaveAgent(private val config: AgentConfiguration,
     @Suppress("WHEN_WITHOUT_ELSE")  // when with sealed class
     private suspend fun startHeartbeats() = coroutineScope {
         println("Scheduling heartbeats")
+        runCatching {
+            saveAdditionalData()
```
petertrr, line 73: Maybe, we can create a new inheritor of `Heartbeat` class with additional data? So that all data is submitted in a unified way

kentr0w: I think, we should split this information. `Heartbeat` pass to orchestrator only its status, not additional information.

petertrr: Ok, this makes sense. My concern was about logic that we have for backend queries: retires, etc. You could extract it into a separate method and reuse here

petertrr (changes requested):

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -176,11 +170,16 @@ class SaveAgent(private val config: AgentConfiguration,
...
+        sendExecution: requestExecutionData,
+        sendAdditional: requestAdditionalData,
+        testExecutionDto: TestExecutionDto
+    ) = coroutineScope {
         var retryInterval = config.executionDataInitialRetryMillis
         repeat(config.executionDataRetryAttempts) { attempt ->
             val result = runCatching {
-                postExecutionData(testExecutionDto)
+                sendExecution(testExecutionDto)
+                sendAdditional()
```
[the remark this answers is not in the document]
kentr0w: Yes, fixed

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -163,11 +167,14 @@ class SaveAgent(private val config: AgentConfiguration,
     /**
      * Attempt to send execution data to backend, will retry several times, while increasing delay 2 times on each iteration.
      */
-    private suspend fun sendExecutionData(testExecutionDto: TestExecutionDto): Unit = coroutineScope {
+    private suspend fun sendAdditionalDataToBackend(
+        sendExecution: requestExecutionData,
+        testExecutionDto: TestExecutionDto?,
```
petertrr, line 172: Why not just pass a function, which produces `HttpResponse`, instead of this complicated logic?

save-backend/src/main/resources/application.properties
```diff
@@ -2,4 +2,5 @@ backend.preprocessorUrl=http://preprocessor:5200
 server.port = 5000
 management.endpoints.web.exposure.include=health,info,prometheus
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL5Dialect
-spring.jpa.hibernate.use-new-id-generator-mappings=false
\ No newline at end of file
+spring.jpa.hibernate.use-new-id-generator-mappings=false
+spring.jpa.show-sql=true
```
petertrr, line 6: I think it should be removed, it can spam a lot in logs in production

save-agent/src/nativeMain/resources/agent.properties
petertrr: .

petertrr (changes requested):

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
kentr0w: Because one of these functions has a parameter

petertrr: Can it be captured when creating a lambda?

kentr0w: I suppose, this solution is optimal, at least I don't know how to do this.

save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -163,11 +169,13 @@ class SaveAgent(private val config: AgentConfiguration,
     /**
      * Attempt to send execution data to backend, will retry several times, while increasing delay 2 times on each iteration.
      */
-    private suspend fun sendExecutionData(testExecutionDto: TestExecutionDto): Unit = coroutineScope {
+    private suspend fun sendAdditionalDataToBackend(
```
petertrr, line 172:
```suggestion
    private suspend fun sendDataToBackend(
```

petertrr, line 172:
```suggestion
    private suspend fun sendDataToBackend(
```

petertrr (changes requested):

petertrr (approved):

PR #114, opened 2021-05-19
Preprocessor: Discover test resources

### LINKED ISSUE #105: Preprocessor: Discover tests and test suites using save-core
Now there are two stubs, `getAllTestSuites` and `getAllTests`. This issue depends on https://github.com/cqfn/save/pull/63 (available since `0.1.0-alpha.3`). If the provided project doesn't have valid SAVE configuration (i.e. ConfigDetector or discoverTestFiles throw exceptions), then the whole execution should be postponed and marked as Failed.

petertrr:
### What's done:
* Added dependency on save-common and save-core
* Implemented logic in TestDiscoveringService
* Use snapshot version of SAVE
* Added tests

Closes #105 

Depends on ~https://github.com/cqfn/save/pull/72~

### TODO:
* Snapshot version of SAVE CLI (or we could decide, that we only need snapshot version to integrate with the new API, and actual execution will always use release version) (#134)
* [x] https://github.com/cqfn/save/issues/104
* Incompatible ABI in kotlinx.datetime breaks compilation in Agent; can be worked around by excluding datetime, because we don't use it


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -168,69 +178,94 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
         executionType: ExecutionType,
-        testSuitesDto: List<TestSuiteDto>?
+        testSuiteDtos: List<TestSuiteDto>?
     ) {
-        testSuitesDto?.let {
-            require(executionType == ExecutionType.STANDARD)
-        } ?: require(executionType == ExecutionType.GIT)
+        testSuiteDtos?.let {
+            require(executionType == ExecutionType.STANDARD) { "Test suites shouldn't be provided unless ExecutionType is STANDARD (actual: $executionType)" }
+        } ?: require(executionType == ExecutionType.GIT) { "Test suites are not provided, but should for executionType=$executionType" }
```
kentr0w, line 187: Do we really need this `require`, if we have verification in frontend to pass only one type of project?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -168,69 +178,94 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
     private fun sendToBackendAndOrchestrator(
         project: Project,
         propertiesRelativePath: String,
-        resourcesRootPath: String,
+        resourcesRootRelativePath: String,
```
kentr0w, line 181: Here it isn't path to resources

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -157,7 +161,13 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
 
     /**
-     * We have not null list of TestSuite only, if execute type is STANDARD ()
+     * Note: We have not null list of TestSuite only, if execute type is STANDARD ()
+     *
+     * - Post request to backend to create PENDING executions
+     * - Discover all test suites in the cloned project
+     * - Post request to backend to save all test suites
+     * - Discover all tests in the cloned project
+     * - Post request to backend to save all tests and create TestExecutions for them
```
kentr0w, line 170: What about `initializeAgent`?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -168,69 +178,94 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
+    }
+
+    private fun discoverAndSaveTestSuites(project: Project,
+                                          testResourcesRootPath: String,
+                                          propertiesRelativePath: String): Mono<List<TestSuite>> {
+        val testSuites: List<TestSuiteDto> = try {
+            testDiscoveringService.getAllTestSuites(project, testResourcesRootPath, propertiesRelativePath)
+        } catch (iae: IllegalArgumentException) {
+            log.error("Couldn't discover test suites, aborting: ", iae)
+            throw iae
```
kentr0w, line 222: Why do you throw an exception?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -168,69 +178,94 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
+                    }
+            }
+            .subscribe()
+    }
+
+    @Suppress("UnsafeCallOnNullableType")
+    private fun getTestResourcesRootAbsolutePath(propertiesRelativePath: String,
+                                                 resourcesRootRelativePath: String): String {
+        val saveProperties: SaveProperties = decodeFromPropertiesFile(
+            File(configProperties.repository, resourcesRootRelativePath)
```
kentr0w, line 207: If i correctly understand, `resourcesRootRelativePath` is path to project, you need to exctract property fro project by its path?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -168,69 +178,94 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
...
+                            .then(initializeAgents(execution, executionId))
+                    }
+            }
+            .subscribe()
+    }
+
+    @Suppress("UnsafeCallOnNullableType")
+    private fun getTestResourcesRootAbsolutePath(propertiesRelativePath: String,
+                                                 resourcesRootRelativePath: String): String {
+        val saveProperties: SaveProperties = decodeFromPropertiesFile(
```
kentr0w, line 206: What if there isn't `save.properties` file?

kentr0w (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Then, the `FileNotFoundException` will be thrown and the job in subscription will be stopped. I believe, I should also mark execution as failed on such errors

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: I believe, additional validation won't hurt, since we don't have good integration tests or tests on frontend yet

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Yes, fixed

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: `propertiesRelativePath` is relative to project root, so there is no need to extract anything else

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: фввув

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Removed in favor of common error handler

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Added a `doOnError` block in mono

kentr0w (approved):
lgtm

PR #136, opened 2021-06-08
Infra: Update diktat to 0.6.1, enable code analysis in .kts files

petertrr:
### What's done:
* Enable code analysis in .kts files
* Code style
* Use codecov bash script instead of action since it's broken now


save-orchestrator/build.gradle.kts
```diff
@@ -20,7 +22,7 @@ tasks.withType<KotlinCompile> {
 }
 
 tasks.getByName("processResources").finalizedBy("downloadSaveCli")
-val saveCliVersion: String = if (Versions.saveCore.endsWith("SNAPSHOT")) {
+val saveCliVersion: String = if (Versions.saveCore.endsWith("SNAPSHOT") || Versions.saveCore.endsWith("]")) {
```
kentr0w, line 25: Why ?
```kotlin
Versions.saveCore.endsWith("]"))
```

petertrr: I was experimenting with version ranges to use specific snapshot version and forgot to delete this, I'll delete it now

kentr0w (approved):
lgtm

PR #140, opened 2021-06-17
Show history in frontend

### LINKED ISSUE #61: Frontend: Populate different views with data from backend

kentr0w:
### What's done:
- Changed execution entity
- Added logic to view history
- Added request to backend to take history information

Closes #61

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
@@ -7,7 +7,6 @@ import org.cqfn.save.execution.ExecutionUpdateDto
 
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
-import org.springframework.transaction.annotation.Transactional
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
@@ -50,7 +49,6 @@ class ExecutionController(private val executionService: ExecutionService) {
      * @return list of execution dtos
      */
     @GetMapping("/executionDtoList")
-    @Transactional
     fun getExecutionByProject(@RequestParam name: String, @RequestParam owner: String): ExecutionDtoListResponse =
             ResponseEntity
                 .status(HttpStatus.OK)
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
@@ -44,7 +44,7 @@ class ExecutionController(private val executionService: ExecutionService) {
                 ResponseEntity.status(HttpStatus.OK).body(it)
             } ?: ResponseEntity.status(HttpStatus.NOT_FOUND).build()
 
-    @PostMapping("executionDtoByProject")
+    @PostMapping("/executionDtoByProject")
     fun getExecutionByProject(@RequestBody project: Project): ResponseEntity<List<ExecutionDto>> =
             executionService.getExecutionDtoByProject(project)?.let {
                 ResponseEntity.status(HttpStatus.OK).body(it)
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
@@ -8,6 +8,7 @@ import org.cqfn.save.execution.ExecutionUpdateDto
 
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
+import org.springframework.transaction.annotation.Transactional
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
@@ -48,9 +49,10 @@ class ExecutionController(private val executionService: ExecutionService) {
      * @param project
      * @return list of execution dtos
      */
-    @PostMapping("/executionDtoByProject")
-    fun getExecutionByProject(@RequestBody project: Project): ExecutionDtoListResponse =
-            executionService.getExecutionDtoByProject(project)?.let {
+    @GetMapping("/executionDtoByNameAndOwner")
+    @Transactional
+    fun getExecutionByProject(@RequestParam name: String, @RequestParam owner: String): ExecutionDtoListResponse =
+            executionService.getExecutionDtoByNameAndOwner(name, owner)?.let {
                 ResponseEntity.status(HttpStatus.OK).body(it)
             } ?: ResponseEntity.status(HttpStatus.NOT_FOUND).build()
 }
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
@@ -49,12 +49,12 @@ class ExecutionController(private val executionService: ExecutionService) {
      * @param owner
      * @return list of execution dtos
      */
-    @GetMapping("/executionDtoByNameAndOwner")
+    @GetMapping("/executionDtoList")
     @Transactional
     fun getExecutionByProject(@RequestParam name: String, @RequestParam owner: String): ExecutionDtoListResponse =
-            executionService.getExecutionDtoByNameAndOwner(name, owner)?.let {
-                ResponseEntity.status(HttpStatus.OK).body(it)
-            } ?: ResponseEntity.status(HttpStatus.NOT_FOUND).build()
+            ResponseEntity
+                .status(HttpStatus.OK)
+                .body(executionService.getExecutionDtoByNameAndOwner(name, owner))
 }
 
 typealias ExecutionDtoListResponse = ResponseEntity<List<ExecutionDto>>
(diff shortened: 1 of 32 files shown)


save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/Agent.kt
```diff
@@ -14,7 +14,7 @@ import javax.persistence.ManyToOne
 class Agent(
     var containerId: String,
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.EAGER)
```
petertrr, line 17: Why did you change it? Absence of this field should be fixed by applying `@Transactional` when necessary

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -51,15 +60,28 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
      * @return list of lost tests
      */
+    @Suppress("TOO_MANY_LINES_IN_LAMBDA")
     fun saveTestResult(testExecutionsDtos: List<TestExecutionDto>): List<TestExecutionDto> {
         val lostTests: MutableList<TestExecutionDto> = mutableListOf()
+        val execution = agentRepository
+            .findByIdOrNull(testExecutionsDtos.first().agentId)  // we take agent id only from first element, because all test executions have same execution
+            ?.execution
+            ?: run {
+                log.error("Agent with id=[${testExecutionsDtos.first().agentId}] was not found in the DB")
```
petertrr, line 70:
```suggestion
                log.error("Agent with id=[${testExecutionsDtos.first().agentId}] was not found in the DB or doesn't have valid execution")
```
This is also possible here?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -51,15 +60,28 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
+                return lostTests
+            }
         testExecutionsDtos.forEach { testExecDto ->
             val foundTestExec = testExecutionRepository.findById(testExecDto.id)
             foundTestExec.ifPresentOrElse({
                 it.run {
```
petertrr, line 76: I didn't think about it before, but it's a very strange statement here: we immediately swap `it` for `this` without any obvious improvements. Maybe, we should remove this `run`?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -42,36 +50,45 @@ class HistoryView : RComponent<HistoryProps, RState>() {
                         +"${it.row.index}"
                     }
                 }
-                column("date", "Date") {
+                column("status", "Status") {
```
petertrr, line 53: We should still have a column for date, I think we can use execution end time

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -42,36 +50,45 @@ class HistoryView : RComponent<HistoryProps, RState>() {
...
-                json("date" to "26-Jan-2016", "passed" to "26", "failed" to "17", "skipped" to "36"),
-                json("date" to "76-Jun-2019", "passed" to "67", "failed" to "75", "skipped" to "236"),
+
+            post(
+                url = "${window.location.origin}/executionDtoByProject",
+                headers = Headers().also {
+                    it.set("Accept", "application/json")
+                    it.set("Content-Type", "application/json")
+                },
+                body = JSON.stringify(getProject(props.name, props.owner))
```
petertrr, line 82: Can we make `executionDtoByProject` accept name and owner? This way we will make one query to backend instead of two, and backend can query the whole project on its own

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -42,4 +43,18 @@ class ExecutionController(private val executionService: ExecutionService) {
             executionService.getExecutionDto(executionId)?.let {
                 ResponseEntity.status(HttpStatus.OK).body(it)
             } ?: ResponseEntity.status(HttpStatus.NOT_FOUND).build()
+
+    /**
+     * @param name
+     * @param owner
+     * @return list of execution dtos
+     */
+    @GetMapping("/executionDtoByNameAndOwner")
```
petertrr, line 52:
```suggestion
    @GetMapping("/executionDtoList")
```
I don't think specifying parameters in method name is good here, it's more important that we get several objects, not just one

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/ExecutionRepository.kt
```diff
@@ -1,10 +1,17 @@
...
  * Repository of execution
  */
 @Repository
-interface ExecutionRepository : BaseEntityRepository<Execution>
+interface ExecutionRepository : BaseEntityRepository<Execution> {
+    /**
+     * @param project
+     * @return list of executions
+     */
+    fun getAllByProject(project: Project): List<Execution>
```
petertrr, line 16:
```suggestion
    fun getAllByProjectNameAndProjectOwner(name: String, owner: String): List<Execution>
```
I think spring-data allows to write what we need in one call

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
```diff
@@ -48,11 +52,17 @@ class Execution(
...
+    var failedTests: Long,
+
+    var skippedTests: Long,
 
 ) : BaseEntity() {
     /**
      * @return Execution dto
      */
-    fun toDto() = ExecutionDto(status, type, version)
+    fun toDto() = ExecutionDto(status, type, version, endTime?.format(DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss")), passedTests, failedTests, skippedTests)
```
petertrr, line 67: Why do we have endTime as String in DTO? I would expect it to be Long timestamp, or some serializable class from kx.datetime

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -49,6 +49,7 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
      * @return response
      */
     @PostMapping(value = ["/saveTestResult"])
+    @Transactional
```
petertrr, line 52: Why do we need transactional here? It seems that this controller is not affected in this PR anywhere else

kentr0w: Yes, but I changed tests for this request and forgot to remove it.

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
kentr0w: Because we use it only in front-end and I thought, to avoid problem with cast use `endTime` as `String`

petertrr: Usually timestamp is a more universal representation, which will be converted to the desired format on frontend without the need to parse string in another format

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -42,4 +43,18 @@ class ExecutionController(private val executionService: ExecutionService) {
...
                 ResponseEntity.status(HttpStatus.OK).body(it)
             } ?: ResponseEntity.status(HttpStatus.NOT_FOUND).build()
+
+    /**
+     * @param name
+     * @param owner
+     * @return list of execution dtos
+     */
+    @GetMapping("/executionDtoList")
+    @Transactional
```
petertrr, line 53: I think `@Transactional` is redundant now

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
petertrr: BTW, we already have `org.cqfn.save.utils.LocalDateTimeSerializer` to make it easier to convert `LocalDateTime` to timestamp.

petertrr:
@kentr0w , also there is a problem, related to histroy view, as stated in the issue:

> Also, href from ProjectView, which opens HistoryView on submission, is broken: opens /history instead of #/owner/name/history

Please fix it in this PR as well

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -56,7 +57,11 @@ class HistoryView : RComponent<HistoryProps, RState>() {
                 }
                 column("date", "Date") {
                     td {
-                        +(it.value.endTime ?: "RUNNING")
+                        +(it.value.endTime?.let {
+                            Instant.fromEpochSeconds(it.toString().toLong())  // there is some bug with parse endTime to Instant - that why this crutch
```
petertrr, line 61: How does `Long.toString().toLong()` help? What exactly was the bug?

petertrr (approved):

PR #142, opened 2021-06-22
Fixes & improvements

### LINKED ISSUE #134: Preprocessor: Use snapshot version of save-cli
> Snapshot version of SAVE CLI (or we could decide, that we only need snapshot version to integrate with the new API, and actual execution will always use release version) (originally posted in https://github.com/cqfn/save-cloud/pull/114)

Now we will have SAVE API with snapshot version, while CLI will use latest release version.

For example, artifacts can be downloaded from here: https://github.com/cqfn/save/actions/runs/885403597. But Github's artifact API is limited to actions and requires authentication.

petertrr:
### What's done:
* Allow manual placing of `save-cli` for snapshots
* Chain `Mono`s in `HeartbeatController` (`updateAgentStatusesWithDto` wasn't executed before)
* Provide file names in `cliArgs` for `save-cli`
* Code cleanup
* Renamed rootProject, fixed build
* Updated dependencies

Closes #134

### DIFF
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
@@ -45,32 +45,34 @@ class HeartbeatController(private val agentService: AgentService,
     fun acceptHeartbeat(@RequestBody heartbeat: Heartbeat): Mono<out HeartbeatResponse> {
         logger.info("Got heartbeat state: ${heartbeat.state.name} from ${heartbeat.agentId}")
         // store new state into DB
-        agentService.updateAgentStatusesWithDto(
+        return agentService.updateAgentStatusesWithDto(
             listOf(
                 AgentStatusDto(LocalDateTime.now(), heartbeat.state, heartbeat.agentId)
             )
         )
-        return when (heartbeat.state) {
-            AgentState.IDLE -> agentService.setNewTestsIds(heartbeat.agentId).subscribeOn(scheduler)
-                .doOnNext {
-                    if (it is WaitResponse) {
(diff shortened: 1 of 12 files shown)


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -45,32 +45,34 @@ class HeartbeatController(private val agentService: AgentService,
...
+            .then(
+                when (heartbeat.state) {
+                    AgentState.IDLE -> agentService.getNewTestsIds(heartbeat.agentId).subscribeOn(scheduler)
+                        .doOnSuccess {
+                            if (it is WaitResponse) {
+                                // If agent was IDLE and there are no new tests - we check if the Execution is completed.
+                                agentService.getAgentsAwaitingStop(heartbeat.agentId).subscribeOn(scheduler).subscribe { finishedAgentIds ->
+                                    if (finishedAgentIds.isNotEmpty()) {
+                                        logger.info("Agents ids=$finishedAgentIds have completed execution and will be terminated")
+                                        dockerService.stopAgents(finishedAgentIds)
```
kentr0w, line 62: Shouldn't we `stopAgent` only if all tests was run?

kentr0w (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
petertrr: Yes, this flow is triggered only when `getTestBatch` returns an empty list. This should mean that all tests have been either executed or distributed among agents for execution

kentr0w (approved):
lgtm

PR #148, opened 2021-06-23
Run sh file

kentr0w:
### What's done:
Updated command

### DIFF
--- a/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
+++ b/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
@@ -30,7 +30,7 @@ data class AgentConfiguration(
     val requestTimeoutMillis: Long,
     val executionDataRetryAttempts: Int,
     val executionDataInitialRetryMillis: Long,
-    val cliCommand: String = "./save-$SAVE_CORE_VERSION-linuxX64.kexe",
+    val cliCommand: String = "test -f run.sh && ./run.sh; ./save-$SAVE_CORE_VERSION-linuxX64.kexe",
 )
 
 /**
--- a/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
+++ b/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
@@ -30,7 +30,7 @@ data class AgentConfiguration(
     val requestTimeoutMillis: Long,
     val executionDataRetryAttempts: Int,
     val executionDataInitialRetryMillis: Long,
-    val cliCommand: String = "test -f run.sh && ./run.sh; ./save-$SAVE_CORE_VERSION-linuxX64.kexe",
+    val cliCommand: String = "test -f run.sh && ./run.sh || ./save-$SAVE_CORE_VERSION-linuxX64.kexe",
 )
 
 /**
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -109,7 +109,7 @@ class DockerService(private val configProperties: ConfigProperties) {
             execution.resourcesRootPath,
         )
         val runCmd = if (File(resourcesPath, "run.sh").exists()) {
-                "./run.sh || ./$SAVE_AGENT_EXECUTABLE_NAME"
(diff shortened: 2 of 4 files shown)


save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
```diff
@@ -30,7 +30,7 @@ data class AgentConfiguration(
     val requestTimeoutMillis: Long,
     val executionDataRetryAttempts: Int,
     val executionDataInitialRetryMillis: Long,
-    val cliCommand: String = "./save-$SAVE_CORE_VERSION-linuxX64.kexe",
+    val cliCommand: String = "test -f run.sh && ./run.sh; ./save-$SAVE_CORE_VERSION-linuxX64.kexe",
```
petertrr, line 33: Are you sure it will work? It will execute `kexe` file too, but we expect the situation, when it's called from inside the script. So this will be more suitable:
```suggestion
    val cliCommand: String = "test -f run.sh && ./run.sh || ./save-$SAVE_CORE_VERSION-linuxX64.kexe",
```

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -103,11 +103,16 @@ class DockerService(private val configProperties: ConfigProperties) {
...
             }
 
-    private fun buildBaseImageForExecution(execution: Execution): String {
+    private fun buildBaseImageForExecution(execution: Execution): Pair<String, String> {
         val resourcesPath = File(
             configProperties.testResources.basePath,
             execution.resourcesRootPath,
         )
+        val runCmd = if (File(resourcesPath, "run.sh").exists()) {
+            "./run.sh || ./$SAVE_AGENT_EXECUTABLE_NAME"
```
petertrr, line 112:
```suggestion
            "./run.sh
```
No need to add fallback behavior, if we are already sure that file exists

kentr0w: Fixed

petertrr: You still have this `||` operator, though.

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -103,11 +103,12 @@ class DockerService(private val configProperties: ConfigProperties) {
                 false
             }
 
-    private fun buildBaseImageForExecution(execution: Execution): String {
+    private fun buildBaseImageForExecution(execution: Execution): Pair<String, String> {
         val resourcesPath = File(
             configProperties.testResources.basePath,
             execution.resourcesRootPath,
         )
+        val runCmd = "${if (File(resourcesPath, "run.sh").exists()) "./run.sh || " else ""}./$SAVE_AGENT_EXECUTABLE_NAME"
```
petertrr, line 111: Do you think `||` is useful here? This way we won't be able to pass additional arguments to the script, and if we ave checked that the file exists, there is no need for this safe fallback

petertrr (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kentr0w: Corrected, misunderstood at first

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -103,11 +103,16 @@ class DockerService(private val configProperties: ConfigProperties) {
...
-    private fun buildBaseImageForExecution(execution: Execution): String {
+    private fun buildBaseImageForExecution(execution: Execution): Pair<String, String> {
         val resourcesPath = File(
             configProperties.testResources.basePath,
             execution.resourcesRootPath,
         )
+        val runCmd = if (File(resourcesPath, "run.sh").exists()) {
+            "./run.sh"
+        } else {
+            "./$SAVE_AGENT_EXECUTABLE_NAME\""
```
petertrr, line 114:
```suggestion
            "./$SAVE_AGENT_EXECUTABLE_NAME"
```
An accidental quote?

kentr0w: Forgot to remove

petertrr (approved):

PR #151, opened 2021-06-23
Feature/history link execution

### LINKED ISSUE #146: Frontend: Rows of HistoryView should be links to ExecutionView
* Rows of HistoryView should be links to ExecutionView with correspondihng Execution
* Indexing the table should start from 1

kentr0w:
### What's done:
- Updated dto
- Added logic to link
- Closes #146

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
@@ -48,21 +48,21 @@ class HistoryView : RComponent<HistoryProps, RState>() {
             columns = columns {
                 column("index", "#") {
                     td {
-                        a(href = "${window.location}/${it.value.id}") {
+                        a(href = getHrefToExecution(it.value.id)) {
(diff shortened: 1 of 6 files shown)


petertrr:
Also, keeping in mind that history has more than one page, I think we should sort executions in reverse

kentr0w:
Yes, agree. There should be a later `execution` first.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -41,42 +42,54 @@ external interface HistoryProps : RProps {
...
-                        +"${it.row.index}"
+                        a(href = "${window.location}/${it.value.id}") {
+                            +"${it.row.index + 1}"
+                        }
                     }
                 }
                 column("status", "Status") {
                     td {
-                        +"${it.value.status}"
+                        a(href = "${window.location}/${it.value.id}") {
```
petertrr, line 58: Please create a function for this url, so that if we change it's format, we change it in one place

petertrr (approved):
lgtm

PR #152, opened 2021-06-24
Orchestrator: Fixes for agents shutdown logic

### LINKED ISSUE #145: Orchestrator: Agent statuses and Execution status are not updated after execution end
* [x] Execution is not marked after shutdown.
* [x] AgentState isn't updated: agents are still `IDLE` after shutdown.
* [x] Logging is not clear, when atomic flag for agents stop is already set

petertrr:
### What's done:
* Execution and AgentsStatus are marked as FINISHED after agents shutdown
* Enabled debug for save-agent
* Fix property file path in backend saveTestSuite
* Added inputs for code generation tasks in gradle
* Added more logging for TestDiscoveringService, enabled debug logging in application.properties
* Publish agent's version instead of cli version

Closes #145

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/AgentStatus.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/AgentStatus.kt
@@ -42,8 +42,19 @@ class AgentStatus(
  * @property containerId id of the agent's container
  * @property time
  */
-class AgentStatusDto(
+data class AgentStatusDto(
     val time: LocalDateTime,
     val state: AgentState,
     val containerId: String,
+)
+
+/**
+ * Statuses of a group of agents for a single Execution
+ *
+ * @property executionId id of Execution
+ * @property agentStatuses list of [AgentStatusDto]s
+ */
+data class AgentStatusesForExecution(
+    val executionId: Long,
+    val agentStatuses: List<AgentStatusDto>,
 )
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
@@ -81,9 +81,10 @@ class HeartbeatController(private val agentService: AgentService,
         agentService.getAgentsAwaitingStop(agentId).doOnSuccess { (executionId, finishedAgentIds) ->
             scheduler.schedule {
                 if (finishedAgentIds.isNotEmpty()) {
-                    logger.info("Agents ids=$finishedAgentIds have completed execution, will make an attempt to terminate them")
+                    logger.debug("Agents ids=$finishedAgentIds have completed execution, will make an attempt to terminate them")
                     val areAgentsStopped = dockerService.stopAgents(finishedAgentIds)
                     if (areAgentsStopped) {
+                        logger.info("Agents have been stopped, will mark execution id=$executionId and agents $finishedAgentIds as FINISHED")
                         agentService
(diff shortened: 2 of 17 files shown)


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -52,16 +52,10 @@ class HeartbeatController(private val agentService: AgentService,
         )
             .then(
                 when (heartbeat.state) {
-                    AgentState.IDLE -> agentService.getNewTestsIds(heartbeat.agentId).subscribeOn(scheduler)
+                    AgentState.IDLE -> agentService.getNewTestsIds(heartbeat.agentId)
```
kentr0w, line 55: Why did you remove `subscribeOn`?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -80,17 +81,27 @@ class DockerService(private val configProperties: ConfigProperties) {
...
+                try {
+                    agentIds.forEach {
+                        log.info("Stopping agent with id=$it")
+                        containerManager.dockerClient.stopContainerCmd(it).exec()
+                        log.info("Agent with id=$it has been stopped")
+                    }
+                    true
+                } catch (dex: DockerException) {
+                    log.error("Error while stopping agents $agentIds", dex)
+                    false
```
kentr0w, line 97: Do we need to do something else to stop container?

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/AgentStatus.kt
```diff
@@ -42,8 +42,19 @@ class AgentStatus(
...
     val containerId: String,
 )
+
+/**
+ * Statuses of a group of agents for a single Execution
+ *
+ * @property executionId id of Execution
+ * @property agentStatuses list of [AgentStatusDto]s
+ */
+data class AgentStatusesForExecution(
```
kentr0w, line 57: May be just add `executionId` to `AgentStatusDto` to avoid new class?

kentr0w (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
petertrr: Because it's a little refactored now: before we were adding a blocking handler for the returned Mono, which was executed on `sheduler`. Now we make another request inside `doOnSuccess` and subscribe on *that* request on `scheduler`.

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
petertrr: There are two documented types of `DockerException` here: 
```java
     * @throws NotFoundException
     *             No such container
     * @throws NotModifiedException
     *             Container already stopped
```
Both of them seem to indicate that we shouldn't try to stop it again. And even if the container haven't been stopped, it will continue to send heartbeats and we will continue to try to stop it.

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/AgentStatus.kt
petertrr: I think this class can more clearly indicate, that all agents have the same `executionId` here, and also removes a little of data duplication.

kentr0w (approved):
lgtm

PR #171, opened 2021-07-09
WIP: deploy via github actions

petertrr:
### What's done:
* Added deploy.yml


akuleshov7:
I think it will never be done

Issue #156, opened 2021-06-24
Orchestrator: Agents shutdown logic has problems


petertrr:
* [x] (done in #174) Shutdown occurs too soon, when other agents haven't sent first heartbeat yet (~actually, it's a corner-case, because usually the first agent will receive non-empty batch~ - if one agent receives an empty batch after the first heartbeat, it will be enough to initiate stop of all agents):
 ```
 14:14:04.749 [reactor-http-epoll-5] INFO o.c.s.o.service.DockerService - Agent with id=61ca029dfa9640581b637daf0e8107b361174e3bce8e517b069e8559f48480c2 has been stopped
 14:14:04.352 [reactor-http-epoll-5] INFO o.c.s.o.service.DockerService - Stopping agent with id=61ca029dfa9640581b637daf0e8107b361174e3bce8e517b069e8559f48480c2
 14:14:04.351 [reactor-http-epoll-5] INFO o.c.s.o.service.DockerService - Agent with id=d24bf4f55a03e08c04b64fdd97981ae205dfec7fb7d6c2454bb55e0e1832208e has been stopped
 14:14:04.048 [reactor-http-epoll-1] INFO o.c.s.o.c.HeartbeatController - Agents ids=[5e09cb9c76570853c92d89e1bd740dccd277aa74df52c16e655c9fe2a74b8a24, d24bf4f55a03e08c04b64fdd97981ae205dfec7fb7d6c2454bb55e0e1832208e, 61ca029dfa9640581b637daf0e8107b361174e3bce8e517b069e8559f48480c2] have completed execution and will be terminated
 14:14:04.012 [reactor-http-epoll-6] INFO o.c.s.o.service.AgentService - Next test batch for agentId=d24bf4f55a03e08c04b64fdd97981ae205dfec7fb7d6c2454bb55e0e1832208e is empty, setting it to wait
 14:14:03.978 [reactor-http-epoll-8] INFO o.c.s.o.c.HeartbeatController - Got heartbeat state: IDLE from d24bf4f55a03e08c04b64fdd97981ae205dfec7fb7d6c2454bb55e0e1832208e
 14:14:03.969 [reactor-http-epoll-5] INFO o.c.s.o.service.DockerService - Stopping agent with id=d24bf4f55a03e08c04b64fdd97981ae205dfec7fb7d6c2454bb55e0e1832208e
[...]

PR #174, opened 2021-07-13
Add AgentState.STARTING for initial agent's heartbeat

petertrr:
### What's done:
* Added state STARTING
* Added tests

Previousely we had the following case:
- Agents are started
- The first agent sends IDLE, gets a batch of tests
- The second agent sends IDLE, there are no more tests
- Shutdown sequence is initiated; because all agents have IDLE in DB they all are subjects for termination
- First agent is stopped whether it has finished execution or not

This PR proposes a new state for the very first heartbeat, and for this state orchestrator doesn't check shutdown conditions.

This is part of #156

### DIFF
--- a/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/nativeMain/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -57,7 +57,7 @@ class SaveAgent(private val config: AgentConfiguration,
     /**
      * The current [AgentState] of this agent
      */
-    val state = AtomicReference(AgentState.IDLE)
+    val state = AtomicReference(AgentState.STARTING)
     private var saveProcessJob: Job? = null
 
     /**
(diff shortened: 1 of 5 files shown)


kentr0w (commented):
Maybe it's better to change logic in orchestrator? I mean orchestrator should initiate shutdown sequence only if all heartbeats is `IDLE` and only after there are no more tests.

petertrr:
> Maybe it's better to change logic in orchestrator? I mean orchestrator should initiate shutdown sequence only if all heartbeats is `IDLE` and only after there are no more tests.

Yes, that's what it is currently doing. However, now we have no means to separate agents with tests, who haven't updated their state yet, from completely idle agents. I also thought about how it could be done in the orchestrator. We could add additional collection in the orchestrator to store current agent states and their assigned jobs, however, it will make orchestrator stateful and harder to restart. We could add additional data in the DB, but it will require additional queries on every idle heartbeat. So I decided, that adding this responsibility into agent will be easier.

Do you have other ideas about orchestrator?

kentr0w:
Yes, I thought about additional collection, but now I agree, this is easier way.

PR #175, opened 2021-07-14
Save execution before all

### LINKED ISSUE #155: Execution should be created right after request submission
Now execution is created only after successful downloading of repo => user doesn't see it after submission of frontend. Also, if preprocessor fails to download resources, it is not stored in history in any way.

kentr0w:
### What's done:
- Changed logic to save execution before all
- Changed logic in preprocessor to update execution
- Added tests
- Closes #155

### DIFF
--- a/db/changelog/v-1/06-chageset-execution-table.xml
+++ b/db/changelog/v-1/06-chageset-execution-table.xml
@@ -22,28 +22,28 @@
                 <constraints nullable="false"/>
             </column>
             <column name="end_time" type="timestamp">
-                <constraints nullable="false" />
+                <constraints nullable="true" />
             </column>
             <column name="status" type="varchar(100)">
                 <constraints nullable="false" />
             </column>
             <column name="test_suite_ids" type="varchar(255)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="resources_root_path" type="varchar(255)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="page" type="integer">
                 <constraints nullable="false"/>
             </column>
             <column name="batch_size" type="integer">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="type" type="varchar(100)">
                 <constraints nullable="false" />
             </column>
             <column name="version" type="varchar(255)">
-                <constraints nullable="false" />
+                <constraints nullable="true" />
             </column>
             <column name="passed_tests" type="bigint">
                 <constraints nullable="false" />
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
@@ -2,9 +2,13 @@ package org.cqfn.save.backend.controllers
 
 import org.cqfn.save.backend.StringResponse
 import org.cqfn.save.backend.configs.ConfigProperties
+import org.cqfn.save.backend.service.ExecutionService
 import org.cqfn.save.backend.service.ProjectService
+import org.cqfn.save.entities.Execution
 import org.cqfn.save.entities.ExecutionRequest
 import org.cqfn.save.entities.ExecutionRequestForStandardSuites
+import org.cqfn.save.execution.ExecutionStatus
+import org.cqfn.save.execution.ExecutionType
 
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
@@ -20,6 +24,7 @@ import org.springframework.web.reactive.function.BodyInserters
 import org.springframework.web.reactive.function.client.WebClient
 import reactor.core.publisher.Mono
 import reactor.kotlin.core.publisher.toMono
+import java.time.LocalDateTime
 
 /**
  * Controller to save project
@@ -31,6 +36,7 @@ import reactor.kotlin.core.publisher.toMono
 @RestController
 class CloneRepositoryController(
     private val projectService: ProjectService,
+    private val executionService: ExecutionService,
     configProperties: ConfigProperties,
 ) {
     private val log = LoggerFactory.getLogger(CloneRepositoryController::class.java)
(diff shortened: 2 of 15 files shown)


save-cloud-common/src/commonMain/kotlin/org/cqfn/save/execution/ExecutionUpdateCreationDto.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @property project
+ * @property testSuiteIds
+ * @property resourcesRootPath
+ * @property batchSize
+ * @property version
+ */
+@Serializable
+data class ExecutionUpdateCreationDto(
```
petertrr, line 14:
```suggestion
data class ExecutionInitializationDto(
```
or someting similar? Execution is already created by the time this class is used

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -214,10 +217,9 @@ class DownloadProjectController(private val configProperties: ConfigProperties)
             require(executionType == ExecutionType.STANDARD) { "Test suites shouldn't be provided unless ExecutionType is STANDARD (actual: $executionType)" }
         } ?: require(executionType == ExecutionType.GIT) { "Test suites are not provided, but should for executionType=$executionType" }
 
-        val execution = Execution(project, LocalDateTime.now(), null, ExecutionStatus.PENDING, "ALL",
-            projectRootRelativePath, 0, configProperties.executionLimit, executionType, executionVersion, 0, 0, 0)
-        webClientBackend.makeRequest(BodyInserters.fromValue(execution), "/createExecution") { it.bodyToMono<Long>() }
-            .flatMap { executionId ->
+        val executionUpdate = ExecutionUpdateCreationDto(project, "ALL", projectRootRelativePath, configProperties.executionLimit, executionVersion)
+        webClientBackend.makeRequest(BodyInserters.fromValue(executionUpdate), "/updateNewExecution") { it.bodyToMono<Execution>() }
+            .flatMap { exec ->
```
petertrr, line 222:
```suggestion
            .flatMap { execution ->
```

db/changelog/v-1/06-chageset-execution-table.xml
```diff
@@ -22,28 +22,28 @@
                 <constraints nullable="false"/>
             </column>
             <column name="end_time" type="timestamp">
-                <constraints nullable="false" />
+                <constraints nullable="true" />
```
petertrr, line 25: @aktsay6 insisted very much, that this field should be not nullable, because we will not read it until execution status changes to `finish` :) But, I guess, since we are making all other fields nullable, it doesn't matter

petertrr (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -42,11 +48,16 @@ class CloneRepositoryController(
      * @param executionRequest information about project
      * @return mono string
      */
+    @Suppress("TOO_MANY_LINES_IN_LAMBDA")
     @PostMapping(value = ["/submitExecutionRequest"])
     fun submitExecutionRequest(@RequestBody executionRequest: ExecutionRequest): Mono<StringResponse> {
         val projectExecution = executionRequest.project
         val project = projectService.getProjectByNameAndOwner(projectExecution.name, projectExecution.owner)
         return project?.let {
+            val execution = Execution(project, LocalDateTime.now(), null, ExecutionStatus.PENDING, null,
```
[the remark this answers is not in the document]
kentr0w: I moved it into another method to save execution in backend.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -75,6 +84,7 @@ class CloneRepositoryController(
         val projectExecution = executionRequestForStandardSuites.project
         val project = projectService.getProjectByNameAndOwner(projectExecution.name, projectExecution.owner)
         project?.let { proj ->
+            saveExecution(proj, ExecutionType.STANDARD)
```
petertrr, line 87: You can discard lambda's argument completely and just use `project`, which is smart-casted to non-nullable type inside this scope. I think it improves readability compared to introducing a new name for the same variable. Or just use `it`, likein the first method.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/ExecutionService.kt
```diff
@@ -77,4 +80,18 @@ class ExecutionService(private val executionRepository: ExecutionRepository) {
...
     fun getLatestExecutionByProjectNameAndProjectOwner(name: String, owner: String): Execution? =
             executionRepository.findTopByProjectNameAndProjectOwnerOrderByStartTimeDesc(name, owner)
+
+    /**
+     * @param executionInitializationDto execution dto to update
+     * @return execution
+     */
+    fun updateNewExecution(executionInitializationDto: ExecutionInitializationDto) =
+            executionRepository.findTopByProjectOrderByStartTimeDesc(executionInitializationDto.project)?.let {
+                assert(it.version == null) { "Execution was already updated" }
```
petertrr, line 90: Use `require` on something similar for preconditions; it's better suited for this than JVM-specific `assert`. Also, you still need `-ea` flag even for newer JVMs to enable asserts

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/execution/ExecutionInitializationDto.kt
```diff
@@ -0,0 +1,20 @@
+package org.cqfn.save.execution
+
+import org.cqfn.save.entities.Project
+import kotlinx.serialization.Serializable
+
+/**
```
petertrr, line 6: Any useful docs for this?

petertrr (approved):
LGTM, just fix a couple of minor comments

PR #178, opened 2021-07-14
Backend: when saving new tests, check all fields, not only hash

### LINKED ISSUE #149: Tests should be compared not only by hash, but also by other fields
Now it is as folows:
1. Test has the same hash - don't insert
2. Test suite has changed, but test is still ound to the old one
3. Agent receives old parameters

* [x] What should we do, if suite with same project, name and ... changes it's additional data? - Already covered in #159, we are now saving a new test suite if comparison by example has failed
* [ ] What should we do, if a test with same hash changes something? E.g., if test suite for it changes? - save a new test with updated data

petertrr:
### What's done:
* Changed logic
* Code cleanup

Closes #149

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
@@ -3,14 +3,15 @@ package org.cqfn.save.backend.repository
 import org.cqfn.save.entities.Test
 import org.springframework.stereotype.Repository
 
+import java.util.Optional
+
 /**
  * Repository of tests
  */
 @Repository
+@Suppress("MISSING_KDOC_CLASS_ELEMENTS", "MISSING_KDOC_ON_FUNCTION")
 interface TestRepository : BaseEntityRepository<Test> {
-    /**
-     * @param hash
-     * @return Test was founded by hash
-     */
-    fun findByHash(hash: String): Test?
+    fun findByHashAndFilePath(hash: String, filePath: String): Optional<Test>
+
+    fun findByHashAndFilePathAndTestSuiteId(hash: String, filePath: String, testSuiteId: Long): Optional<Test>
 }
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -10,7 +10,7 @@ object Versions {
     const val coroutines = "1.5.0"
     const val serialization = "1.1.0"
     const val kotlinxDatetime = "0.2.0"
-    const val saveCore = "0.1.0"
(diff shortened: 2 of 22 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
```diff
@@ -3,14 +3,15 @@ package org.cqfn.save.backend.repository
...
+@Suppress("MISSING_KDOC_CLASS_ELEMENTS", "MISSING_KDOC_ON_FUNCTION")
 interface TestRepository : BaseEntityRepository<Test> {
-    /**
-     * @param hash
-     * @return Test was founded by hash
-     */
-    fun findByHash(hash: String): Test?
+    fun findByHashAndFilePath(hash: String, filePath: String): Optional<Test>
+
+    fun findByHashAndFilePathAndTestSuiteId(hash: String, filePath: String, testSuiteId: Long): Optional<Test>
```
kentr0w, line 16: Is `TestSuiteId` necessary here?

kentr0w: Why not just `Test?` instead of `Optional`

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
```diff
@@ -3,14 +3,15 @@ package org.cqfn.save.backend.repository
...
  */
 @Repository
+@Suppress("MISSING_KDOC_CLASS_ELEMENTS", "MISSING_KDOC_ON_FUNCTION")
 interface TestRepository : BaseEntityRepository<Test> {
-    /**
-     * @param hash
-     * @return Test was founded by hash
-     */
-    fun findByHash(hash: String): Test?
+    fun findByHashAndFilePath(hash: String, filePath: String): Optional<Test>
```
kentr0w, line 14: This method is used only in one test, is it needed?

kentr0w (changes requested):

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
petertrr: Because the same file can be a test in different test suites. Also, using `find(...).orElseGet { ... }.id` seemed nicer than `(find(...) ?: run { ... }).id`

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
petertrr: No, removed and updated the test

kentr0w (approved):
lgtm

PR #185, opened 2021-07-15
Fixed NPE in execution

kentr0w:
###What's done:
Fixed bug


petertrr:
So, the problem is, that we create an empty execution with only project, status and start time and then query it from frontend, and dto had the error with nullability? I guess we'll need some sort of functional test for this flow in the future, but simple fix is ok for now.

petertrr (approved):

PR #191, opened 2021-07-16
Change execution in all cases

### LINKED ISSUE #180: Frontend: after new project creation, should redirect to the main page
Now it stays on `/creation`.

### LINKED ISSUE #189: Follow-up on execution creation: set status to ERROR on earlier stages too
Now we only set ERROR after preprocessor tries to discover test resources. But if project cloning fails, the execution will be forever stuck at PENDING.

kentr0w:
###What's done:
- Changed logic
- Closes #189 
- Closes #180


save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/ExecutionRequest.kt
```diff
@@ -8,10 +8,12 @@ import kotlinx.serialization.Serializable
  * @property project project
  * @property gitDto github repository
  * @property propertiesRelativePath location of save.properties file to start the execution, relative to project's root directory
+ * @property executionId
```
petertrr, line 11: Any docs, specifically about nullability?

petertrr (approved):

PR #192, opened 2021-07-19
Save test results to database

### LINKED ISSUE #179: Backend: implement /executionData endpoint
Need to implement an endpoint to accept data from agent, which is implemented in #173 
After that, we can use all this data to set execution status in #157

petertrr:
### What's done:
* Changed target endpoint
* Build save-agent only for linuxX64
* Use --test-root-path instead of --properties-file

We actually had an endpoint that should be receiving test results, but agent was querying a different endpoint. I have changed this and adjusted logic of saving.

Closes #179


save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
```diff
@@ -40,3 +40,17 @@ data class AgentConfiguration(
...
+ * Configuration for connection to backend service
+ *
+ * @property url URL of backend
+ * @property additionalDataEndpoint endpoint to post additional data (version etc.) to
+ * @property executionDataEndpoint endpoint to post execution data to
+ */
+@Serializable
+data class BackendConfig(
+    val url: String,
+    val additionalDataEndpoint: String = "saveAgentVersion",
```
kentr0w, line 54: Why might endpoints in backend it be customizable?

kentr0w (changes requested):
And according to RP name there isn't logic to save execution

petertrr:
> And according to RP name there isn't logic to save execution

It's logic to save execution data, not execution itself. I'll rename it to test results

save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/AgentConfiguration.kt
petertrr: It's mostly not about customization, but rather about keeping the whole configuration in a single place, because we actually had an endpoint in backend, but in agent a different endpoint was hardcoded somewhere deep in logic.

kentr0w (approved):
LGTM, but anyway, I don't understand why to put the endpoint that won't be changed in the configuration)

PR #202, opened 2021-07-27
Bugfixes around save-agent

petertrr:
### What's done:
* Make some HeartbeatResponses classes instead of objects
* Fix `serializersModule` in save-agent to workaround kx.serialization bug
* Serialize HeartbeatResponses explicitly in orchestrator
* Fix bug in logic of `run.sh` execution
* Save initial agent state with status STARTING
* Log heartbeats in agent
* Bump gradle to 7.1.1

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -204,7 +204,7 @@ class SaveAgent(private val config: AgentConfiguration,
      * @return a [HeartbeatResponse] from Orchestrator
      */
     internal suspend fun sendHeartbeat(executionProgress: ExecutionProgress): HeartbeatResponse {
-        // log.trace("Sending heartbeat to $orchestratorUrl")
+         logDebug("Sending heartbeat to ${config.orchestratorUrl}")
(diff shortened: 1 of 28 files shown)


save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -213,7 +204,7 @@ class SaveAgent(private val config: AgentConfiguration,
      * @return a [HeartbeatResponse] from Orchestrator
      */
     internal suspend fun sendHeartbeat(executionProgress: ExecutionProgress): HeartbeatResponse {
-        // log.trace("Sending heartbeat to $orchestratorUrl")
+        logDebug("Sending heartbeat to ${config.orchestratorUrl}")
```
kentr0w, line 207: Is this method still will return `HeartbeatResponse`

petertrr: Yes, it will try to deserialize server response as HeartbeatResponse and then use polymorphic serializer, that we have registered when creating httpClient

kentr0w: Got it

kentr0w (approved):
LGTM

PR #204, opened 2021-07-27
Fixed wrong encode/decode

### LINKED ISSUE #203: Bad request during submit request
`Bad request` alert appears when submitted request

kentr0w:
###What's done:
- Fixed bug
- Closes #203

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
@@ -107,8 +107,9 @@ class CloneRepositoryController(
     @Suppress("UnsafeCallOnNullableType")
     private fun saveExecution(project: Project, type: ExecutionType, sdk: Sdk): Long {
         val execution = Execution(project, LocalDateTime.now(), null, ExecutionStatus.PENDING, null,
-            null, 0, null, type, null, 0, 0, 0, sdk.toString())
-        execution.id = executionService.saveExecution(execution)
+            null, 0, null, type, null, 0, 0, 0, sdk.toString()).apply {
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -104,10 +104,13 @@ class CloneRepositoryController(
...
+    @Suppress("UnsafeCallOnNullableType")
     private fun saveExecution(project: Project, type: ExecutionType, sdk: Sdk): Long {
         val execution = Execution(project, LocalDateTime.now(), null, ExecutionStatus.PENDING, null,
-            null, 0, null, type, null, 0, 0, 0, sdk.toString())
+            null, 0, null, type, null, 0, 0, 0, sdk.toString()).apply {
+            id = executionService.saveExecution(this)
+        }
         log.info("Creating a new execution id=${execution.id} for project id=${project.id}")
-        return executionService.saveExecution(execution)
+        return execution.id!!
```
petertrr, line 114: Was there an issue here? I believe, spring-data returns the saved object with id. What's the difference with old version here?

petertrr (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
kentr0w: Yes, you're right. But in `log.info ()`, before saving, GHHs id is still `null`

petertrr: Oh, ok

PR #207, opened 2021-07-28
Restore native targets for agent

kgevorkyan:
### What's done:
* Restore native targets for agent


petertrr:
As discussed, it doesn't make sense to compile mingw target, because agent isn't supposed for execution on Windows machine, it will be immediately packed into docker image by orchestrator and executed there.
However, we indeed can fix compilation from windows for linux, because the only thing missing is `libcurl.so`. I tried a small experiment and was able to succesfully link exeuctable by placing `libcurl.so` into `.konan\dependencies\x86_64-unknown-linux-gnu-gcc-8.3.0-glibc-2.19-kernel-4.9-2\x86_64-unknown-linux-gnu\sysroot\usr\lib`. It could be *relatively* easy automated by gradle by creating a download task (from http://security.ubuntu.com/ubuntu/pool/main/c/curl/libcurl4_7.58.0-2ubuntu3.14_amd64.deb) and extract it using standard [`Copy`](https://docs.gradle.org/current/userguide/working_with_files.html#sec:unpacking_archives_example) task and copy `libcurl.so.4.5.0` into the above path.

petertrr:
Also, to fix orchestrator tests on windows we need the following:
* Use correct `orchestrator.docker.host` on windows (`tcp://localhost:2375`)
* Don't use `runsc` in tests on windows

kgevorkyan:
Unfortunately, `gradle` supports unpacking only of `tar` and `zip` files by `tarTree` and `zipTree` correspondingly.
But extension `deb` is actually `ar` format, which is not supported by gradle
However, in java it is possible to use `commons compress`, to unpack such files. java code could be created in `buildSrc` dir, and called from `build.gradle` of `save-agent`

akuleshov7 (changes requested):
empty PR? :)

PR #213, opened 2021-07-30
Fix multiple re-renders in TableComponent

### LINKED ISSUE #212: Frontend: Infinite calls `Querying test executions for page 0 with size 10`
From ExecutionView

petertrr:
### What's done:
* Correctly unpack array for useEffect (it accepts `vararg`, so array needs to be unpacked)
* Minor changes in TableComponent

Closes #212


kentr0w (approved):
lgtm

PR #218, opened 2021-08-04
Change project view

kentr0w:
## What's done:
Made project view prettier


petertrr (changes requested):
Let's have something like
- 1st column: choice between two execution modes
- middle column: setup execution for a chosen mode, run button
- 3rd column: project info, history link

petertrr (approved):
looks OK to me, but I'd suggest making left column narrower and aligning 1st and 2nd columns to the left

PR #225, opened 2021-08-06
Minor fixes

kgevorkyan:
### What's done:
* Minor fixes: remove temp files, change `projectRootRelativePath` in `saveBinaryFile()` to more correct, logging polishing

### DIFF
--- a/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
+++ b/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
@@ -127,7 +127,7 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
     /**
      * Controller to download standard test suites
      */
-    @ExperimentalFileSystem
+    @OptIn(ExperimentalFileSystem::class)
     @Suppress("TOO_LONG_FUNCTION")
     @PostMapping("/uploadStandardTestSuite")
     fun uploadStandardTestSuite() {
@@ -139,7 +139,7 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
                     log.info("Starting to discover root test config in test root path: $testRootPath")
                     val testResourcesRootAbsolutePath = tmpDir.resolve(testRootPath).absolutePath
                     val rootTestConfig = testDiscoveringService.getRootTestConfig(testResourcesRootAbsolutePath)
-                    log.info("Starting to discover standard test suites for root test config ${rootTestConfig.location.name} in $testResourcesRootAbsolutePath")
+                    log.info("Starting to discover standard test suites for root test config ${rootTestConfig.location}")
                     val propertiesRelativePath = "${rootTestConfig.directory.toFile().relativeTo(tmpDir)}${File.separator}save.properties"
                     val tests = testDiscoveringService.getAllTestSuites(null, rootTestConfig, propertiesRelativePath)
                     log.info("Test suites size = ${tests.size}")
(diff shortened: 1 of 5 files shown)


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -229,11 +232,13 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
...
         val pathToProperties = tmpDir.resolve(propertyFile.name)
         propertyFile.copyTo(pathToProperties)
+        propertyFile.delete()
         binFile.copyTo(tmpDir.resolve(binFile.name))
+        binFile.delete()
         val project = executionRequestForStandardSuites.project
         val propertiesRelativePath = pathToProperties.relativeTo(tmpDir).name
         // todo: what's with version?
-        return updateExecution(executionRequestForStandardSuites.project, propertiesRelativePath, binFile.name).flatMap { execution ->
+        return updateExecution(executionRequestForStandardSuites.project, tmpDir.name, binFile.name).flatMap { execution ->
```
petertrr, line 241: It still should be a part of path, not only directory name, Ithink

petertrr (changes requested):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -126,6 +127,7 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
     /**
      * Controller to download standard test suites
      */
+    @ExperimentalFileSystem
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -134,13 +136,14 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
             val tmpDir = generateDirectory(testSuiteUrl)
             cloneFromGit(GitDto(testSuiteUrl), tmpDir)?.use {
                 Flux.fromIterable(testSuitePaths).flatMap { testRootPath ->
-                    log.info("Starting to discover root test config for test root $testRootPath")
+                    log.info("Starting to discover root test config in test root path: $testRootPath")
                     val testResourcesRootAbsolutePath = tmpDir.resolve(testRootPath).absolutePath
                     val rootTestConfig = testDiscoveringService.getRootTestConfig(testResourcesRootAbsolutePath)
-                    log.info("Starting to discover standard test suites for config test root $testRootPath in $testResourcesRootAbsolutePath")
-                    val tests = testDiscoveringService.getAllTestSuites(null, rootTestConfig, "stub")
+                    log.info("Starting to discover standard test suites for root test config ${rootTestConfig.location.name} in $testResourcesRootAbsolutePath")
```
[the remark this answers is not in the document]
kgevorkyan: This message already contain `testResourcesRootAbsolutePath`, however I changed this

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
kgevorkyan: This path used like `execution.resourcesRootPath` in `buildBaseImageForExecution()` in this way:
```kotlin
        val resourcesPath = File(
            execution.resourcesRootPath,
        )
```

So we take the `configProperties.testResources.basePath` and then appends the relative path, which is name of directory, so I think all is correct.
Or what did you mean?

petertrr: I actually meant, that there can be several directories, and `name` returns the name of the last one only. But as I see now, in this function we don't really pay attention to test root directory and just copy everything into root. Then, `tmpDir.name` works, but we will need to change general logic later.

petertrr (approved):

PR #229, opened 2021-08-13
Upload multiple files 

### LINKED ISSUE #211: Frontend: Upload multiple files
Not only binary and properties, but arbitrary number of files, that will be put into root direcotry

petertrr:
### What's done:
* Accept arbitrary number of files in `/submitExecutionRequestBin` and `/uploadBin`
* Accept files in `/submitExecutionRequest` and `/upload`
* Frontend: Accept multiple files from ProjectView, minor refactorings
* Frontend: extract `CheckBoxComponent`, `SdkSelection` and `FileUploader` into separate files
* Frontend: Fix missing column in CollectionView
* Increase `client_max_body_size` in nginx configuration
* Introduce `ExecutionRequestBase`

In `/rerunExecution` for GIT flow should also restore binaries (created #231)

Closes #211

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
@@ -56,7 +56,7 @@ class CloneRepositoryController(
     @PostMapping(value = ["/submitExecutionRequest"], consumes = ["multipart/form-data"])
     fun submitExecutionRequest(
         @RequestPart(required = true) executionRequest: ExecutionRequest,
-        @RequestPart("file", required = true) files: Flux<FilePart>,
+        @RequestPart("file", required = false) files: Flux<FilePart>,
     ): Mono<StringResponse> {
         val projectExecution = executionRequest.project
         val project = projectService.getProjectByNameAndOwner(projectExecution.name, projectExecution.owner)
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
@@ -66,12 +66,7 @@ class CloneRepositoryController(
             val bodyBuilder = MultipartBodyBuilder().apply {
                 part("executionRequest", executionRequest.copy(executionId = newExecutionId))
             }
-            files.map {
-                log.info("Appending a file ${it.filename()} to multipart")
-                bodyBuilder.part("file", it)
-            }
-                .collectList()
-                .switchIfEmpty(Mono.just(emptyList()))
+            files.collectToMultipart(bodyBuilder)
                 .flatMap {
                     preprocessorWebClient
                         .post()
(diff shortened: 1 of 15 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -92,20 +108,21 @@ class CloneRepositoryController(
...
-            }.flatMap {
-                preprocessorWebClient
-                    .post()
-                    .uri("/uploadBin")
-                    .contentType(MediaType.MULTIPART_FORM_DATA)
-                    .body(BodyInserters.fromMultipartData(bodyBuilder.build()))
-                    .retrieve()
-                    .toEntity(String::class.java)
-                    .toMono()
+            return files.map {
```
kgevorkyan, line 111: This logic is almost the same as in `submitExecutionRequest`, didn't you think about to move it in some function?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
```diff
@@ -197,19 +181,11 @@ class ProjectView : RComponent<ProjectExecutionRouteProps, ProjectViewState>() {
...
-            state.binaryFile ?: run {
-                setState {
-                    isErrorOpen = true
-                    errorLabel = "No binary file has been selected"
-                    errorMessage = "Please select binary file"
-                }
-                return
-            }
-            state.propertyFile ?: run {
+            if (state.files.isEmpty()) {
```
kgevorkyan, line 184: Now we don't check, if user uploaded only one file, for example, just `save.properties`? But we won't be able to work in this case, since bin file is absent

Will it be handled in some other place?

petertrr: I think it becomes really hard to determine which set of files is correct, because for some projects `save.properties` will be optional, for others it would be required; the executable may be a binary or a script (i.e. plaintext) etc. I think it'll be easier to allow arbitrary files for now.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
petertrr: Extracted

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
kgevorkyan: Ok, agree

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
kgevorkyan: And what about `preprocessorWebClient.post()...` chain? They differs only in `uri`. Or I missed something?

petertrr: You are right, I did some more refactoring

kgevorkyan: Now it's looks even more better, than I expect.

Just one small question, why do you decide to create `ExecutionRequestBase` class? Previously we use ExecutionRequest and had no problems

petertrr: Because one version of the method uses `ExecutionRequest` and the other - `ExecutionRequestForStandardSuites`. So instead of passing a number of different arguments, I decided to introduce a base class, which could contain all those values.

kgevorkyan (approved):
In general lgtm, however, I remember, that @akuleshov7 also wanted to review  your PR's, so let's wait some time for him

akuleshov7 (approved):

PR #232, opened 2021-08-18
Check image and container existence before removing

### LINKED ISSUE #230: Orchestrator: Don't cleanup if stuff is already removed
In cleanup method should correctly handle missing containers

petertrr:
### What's done:
* Update logic

Closes #230


kgevorkyan (approved):
lgtm

PR #242, opened 2021-08-27
Persistent file storage

### LINKED ISSUE #210: Upload binary: store persistently, allow to choose from existing
* [x] #237
* [ ] #242 Frontend: Add selection from variants from p.1
* [ ] Rerun support: Store file info in the execution

petertrr:
### What's done:
* Upload file after submission from frontend
* Return FileInfo after uploading file
* Select files from the list of available
* Accept Flux of `FileInfo`s on backend in `/submit*` handlers
* Added devServer proxy for local development of save-frontend
* Make fonts in ProjectView on the right side smaller

### TODO:
* [x] Re-render after upload finishes (check, if new files are added to availableFiles)
* [x] Wait for /upload to complete, disable submission while it's in progress
* [x] Remove already selected files from the list of available
* [ ] Don't upload same files (check on backend in `/upload`: if name, content hash and modified timestamps are equal, return 4xx code or something) (maybe it's better to check on frontend?) for now 409 can be returned if file hasn't been saved because it already exists
* [x] Add FileSystemRepository storage to docker volume
* [x] Add `uploadDate` to FileInfo

Closes #210

### DIFF
--- a/save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
+++ b/save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
@@ -59,8 +59,8 @@ import kotlin.io.path.writeLines
 class DownloadFilesTest {
     @Autowired
     lateinit var webTestClient: WebTestClient
-    @Autowired private lateinit var fileSystemRepository: FileSystemRepository
-
+    @Autowired
+    private lateinit var fileSystemRepository: FileSystemRepository
     @Autowired
     private lateinit var configProperties: ConfigProperties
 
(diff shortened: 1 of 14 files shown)


save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
```diff
@@ -58,19 +59,21 @@ import kotlin.io.path.writeLines
 class DownloadFilesTest {
     @Autowired
     lateinit var webTestClient: WebTestClient
+    @Autowired private lateinit var fileSystemRepository: FileSystemRepository
```
kgevorkyan, line 62: Dose it done intentionally?

```suggestion
    @Autowired
    private lateinit var fileSystemRepository: FileSystemRepository
```

kgevorkyan (approved):

save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
petertrr: Maybe, we should support unified style for annotations in diktat? Currently we are only checking several annotations, not single ones.

kgevorkyan: Good idea

PR #254, opened 2021-09-15
Preprocessor: Close `Git` object

petertrr:
### What's done:
* Changed logic
* Minor refactoring

We didn't close `Git` object after it has been used, which sometimes lead to inability to delete directory. It may be the reason, why tests in preprocessor were flaky.

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -141,7 +141,7 @@ class SaveAgent(private val config: AgentConfiguration,
     }
 
     private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, logFilePath.toPath())
+        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, ".", logFilePath.toPath())
 
     @Suppress("TOO_MANY_LINES_IN_LAMBDA")
     private fun readExecutionResults(jsonFile: String): List<TestExecutionDto> {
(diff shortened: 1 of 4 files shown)


save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -141,7 +141,7 @@ class SaveAgent(private val config: AgentConfiguration,
     }
 
     private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, logFilePath.toPath())
+        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, ".", logFilePath.toPath())
```
kgevorkyan, line 144: May be we can put just `""`? In this case Pb won't change directory at all, of course 'changing dir to current' does not give a large load, however it's still an extra action

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -441,7 +439,7 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
         val saveProperties: SaveProperties = decodeFromPropertiesFile<SaveProperties>(propertiesFile)
             .mergeConfigWithPriorityToThis(defaultConfig())
         return propertiesFile.parentFile
-            .resolve(saveProperties.testRootPath!!)
+            .resolve(saveProperties.testFiles!!.firstOrNull() ?: ".")
```
kgevorkyan, line 442: Why for example just `testFiles!!.first()` is not enough?

kgevorkyan (approved):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Because `SaveProperties` don't have default value for `testFiles`, and here we are using only properties file, while users would usually specify root dir via CLI. Therefore in this case we will very likely face an NPE.

save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
petertrr: OK, changed

PR #258, opened 2021-09-17
Orchestrator: Chown copied files to `cnb:cnb`

petertrr:
### What's done:
* Updated logic

Orchestrator is executed as root (to access docker socket), but files are in a shared volume, so they need to have correct ownership.

### DIFF
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -20,6 +20,9 @@ import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.bodyToMono
 
 import java.io.File
+import java.nio.file.Files
+import java.nio.file.LinkOption
+import java.nio.file.attribute.PosixFileAttributeView
 import java.util.concurrent.atomic.AtomicBoolean
 
 import kotlin.io.path.ExperimentalPathApi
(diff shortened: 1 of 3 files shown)


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -203,14 +206,24 @@ class DockerService(private val configProperties: ConfigProperties) {
...
                 )
+            log.debug("Copying suite ${it.name} from $standardTestSuiteAbsolutePath into $destination/...")
             standardTestSuiteAbsolutePath.copyRecursively(destination.resolve("${it.name}_${it.propertiesRelativePath.hashCode()}_${Random.nextInt()}"))
         }
+        // orchestrator is executed as root (to access docker socket), but files are in a shared volume
+        val lookupService = destination.toPath().fileSystem.userPrincipalLookupService
+        destination.walk().forEach {
+            Files.getFileAttributeView(it.toPath(), PosixFileAttributeView::class.java, LinkOption.NOFOLLOW_LINKS).apply {
+                setGroup(lookupService.lookupPrincipalByGroupName("cnb"))
+                owner = lookupService.lookupPrincipalByName("cnb")
```
kgevorkyan, line 224: May be use one style here?

```suggestion
                setOwner(lookupService.lookupPrincipalByName("cnb"))
```

Strange, that we can't use `group = ....` here, like we does for `owner`

kgevorkyan (approved):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
petertrr: Yeah, that's because `owner` has both setter and getter, while `group` has only a setter. So Kotlin can't implement syntactic sugar with property access for group.

petertrr: I thought it would be a compiler warning, but it's only IDEA's warning, so I'll change both to explicit setters

PR #259, opened 2021-09-20
Frontend: Display more info in ExecutionView.kt

### LINKED ISSUE #245: Save: results list visualization
We should visualize more fields on the test result page:
1. Type of test
2. Type of plugin
3. Extra info (think about it)

petertrr:
### What's done:
* Added ExecutionStatistics.kt
* Added more columns
* Added shutdown logging in save-agent

Closes #245 

Now in looks like this (central object changes color based on execution status):
image

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
@@ -1,0 +1,50 @@
+@file:Suppress("FILE_NAME_MATCH_CLASS")
+
+package org.cqfn.save.frontend.components.basic
+
+import org.cqfn.save.execution.ExecutionDto
+import org.cqfn.save.execution.ExecutionStatus
+import react.Props
+import react.dom.button
+import react.dom.div
+import react.fc
+
+/**
+ * [Props] for execution statistics component
+ */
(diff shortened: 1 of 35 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
```diff
@@ -0,0 +1,50 @@
...
+ * A component which displays statistics about an execution from its props
+ *
+ * @param classes HTML classes for the enclosing div
+ * @return a functional react component
+ */
+fun executionStatistics(classes: String = "") = fc<ExecutionStatisticsProps> { props ->
+    val totalTests = props.executionDto?.run {
+        passedTests + failedTests + skippedTests
+    } ?: 0
+    val isInProgress = props.executionDto?.run { status == ExecutionStatus.RUNNING || status == ExecutionStatus.PENDING } ?: true
```
kgevorkyan, line 32: May be `?: false` ?

petertrr: I thought that if `executionDto` is not available (i.e. we haven't etched if from backend yet) it makes more sense to use the same logic as for in-progress executions.

akuleshov7:
Can we add tags here?

akuleshov7 (approved):

PR #260, opened 2021-09-21
Dependency updates 

petertrr:
### What's done:
* Update kotlin-wrappers: update usages of kotlin-react-router API in App.kt
* Fixed react warning
* Kotlin 1.5.31 and other dependency updates

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
@@ -81,9 +81,10 @@ fun fileUploader(
                 }
                 li("list-group-item d-flex justify-content-between align-items-center") {
                     select(classes = "form-control") {
+                        attrs.value = "default"
                         option {
+                            attrs.value = "default"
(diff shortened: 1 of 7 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
```diff
@@ -81,9 +81,10 @@ fun fileUploader(
                 }
                 li("list-group-item d-flex justify-content-between align-items-center") {
                     select(classes = "form-control") {
+                        attrs.value = "default"
```
kgevorkyan, line 84: Why do we need this twice?

kgevorkyan (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
petertrr: It's for different tags - for `select` and `option`. React encourages to use `value` on `select` instead of `selected` on `option`.

PR #261, opened 2021-09-21
Fix timestamp in `/files/list`

### LINKED ISSUE #248: `NoSuchFileException` after selecting a file from frontend
```
/home/cnb/files/1630487894369/ktlint; nested exception is java.nio.file.NoSuchFileException: /home/cnb/files/1630487894369/ktlint
```
But this file has been returned by `/files/list`.

petertrr:
### What's done:
* Changed logic
* Skip `logging:` lines in docker-compose.yaml in `dev` profile
* Add `FileSystemRepository.delete` method

Closes #248

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/FileSystemRepository.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/repository/FileSystemRepository.kt
@@ -20,6 +20,7 @@ import kotlin.io.path.copyTo
 import kotlin.io.path.createDirectories
 import kotlin.io.path.createDirectory
 import kotlin.io.path.createFile
+import kotlin.io.path.deleteExisting
 import kotlin.io.path.exists
 import kotlin.io.path.fileSize
 import kotlin.io.path.getLastModifiedTime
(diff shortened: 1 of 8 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/repository/FileSystemRepository.kt
```diff
@@ -101,4 +101,21 @@ class FileSystemRepository(configProperties: ConfigProperties) {
...
     }
+
+    /**
+     * Delete a file described by [fileInfo]
+     *
+     * @param fileInfo a [FileInfo] describing a file to be deleted
+     * @return true if file has been deleted successfully, false otherwise
+     */
+    fun delete(fileInfo: FileInfo) = try {
+        Files.delete(fileInfo.getPath())
```
akuleshov7, line 112: why do we need to delete the file first and after that to delete a directory?

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/FileSystemRepository.kt
petertrr: Fixed

PR #262, opened 2021-09-21
Remove usages of `run.sh` from SAVE repo

### LINKED ISSUE #255: Remove temporary added logic for `run.sh`
We eventually decided not to develop this direction in SAVE further and to improve resources uploading flow instead. So this https://github.com/cqfn/save-cloud/blob/42005d4ff2b5140f640e026d994eda2afeef6a6c/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt#L249-L251
logic should be updated, and probably in some other places

petertrr:
### What's done:
* Removed code
* Minor refactoring

Closes #255

### DIFF
--- a/save-agent/src/linuxX64Main/resources/agent.properties
+++ b/save-agent/src/linuxX64Main/resources/agent.properties
@@ -6,3 +6,4 @@ requestTimeoutMillis=1000
 executionDataRetryAttempts=5
 executionDataInitialRetryMillis=1000
 debug=true
+cliCommand=echo Provide CLI command here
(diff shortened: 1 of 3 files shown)


save-agent/src/linuxX64Main/resources/agent.properties
```diff
@@ -5,4 +5,5 @@ heartbeat.intervalMillis=15000
 requestTimeoutMillis=1000
 executionDataRetryAttempts=5
 executionDataInitialRetryMillis=1000
-debug=true
\ No newline at end of file
+debug=true
+cliCommand=echo Provide CLI command here
```
kgevorkyan, line 9: It is just stub line, but why with `echo`?

kgevorkyan (approved):

save-agent/src/linuxX64Main/resources/agent.properties
petertrr: No actual reason; just so that it looks more like a command

kgevorkyan: Ok

PR #263, opened 2021-09-22
Display description of test suites

### LINKED ISSUE #247: Display additional info about standard test suites in ProjectView
When user selects test suties for execution, they should be able to see additional info, e.g. in hover or in modal, opened by click. These modals can also contain a link to tests repo or more detailed description.

petertrr:
### What's done:
* Save description in the DB
* Save only distinct test suites
* Display description on hover on frontend

### TODO:
* [x] Trim description for tooltip, if it's too long
* [x] Add popover with the following info: link to suites repo, name of folder in the repo, full (untrimmed) description
* [x] Add a small `?` sign near every suite name, display info if clicked

image

Closes #247

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
@@ -51,7 +51,7 @@ class TestSuitesService {
                         }
                     }
             }
-            testSuiteRepository.saveAll(testSuites)
+        testSuiteRepository.saveAll(testSuites)
         return testSuites.toList()
     }
 
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
@@ -27,22 +27,31 @@ class TestSuitesService {
     @Suppress("UnsafeCallOnNullableType")
     fun saveTestSuite(testSuitesDto: List<TestSuiteDto>): List<TestSuite> {
         val testSuites = testSuitesDto
-            .map { TestSuite(it.type, it.name, it.project, null, it.propertiesRelativePath, it.testSuiteRepoUrl) }
+            .distinctBy {
+                // Same suites may be declared in different directories, we unify them here.
+                // We allow description of existing test suites to be changed.
+                it.copy(description = null)
+            }
+            .map {
+                TestSuite(it.type, it.name, it.description, it.project, null, it.propertiesRelativePath, it.testSuiteRepoUrl)
+            }
             .map { testSuite ->
                 // try to find TestSuite in the DB based on all non-null properties of `testSuite`
                 // NB: that's why `dateAdded` is null in the mapping above
-                testSuiteRepository.findOne(
-                    Example.of(testSuite)
-                )
-                    // if testSuite is not present in the DB, we will save it with current timestamp
+                val description = testSuite.description
+                testSuiteRepository
+                    .findOne(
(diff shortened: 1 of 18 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
```diff
@@ -37,22 +37,31 @@ class TestSuitesService {
     @Suppress("UnsafeCallOnNullableType")
     fun saveTestSuite(testSuitesDto: List<TestSuiteDto>): List<TestSuite> {
         val testSuites = testSuitesDto
-            .map { TestSuite(it.type, it.name, it.project, null, it.propertiesRelativePath, it.testSuiteRepoUrl) }
+            .distinctBy {
```
akuleshov7, line 40: don't understand the case for distinct

I understand when we could have the duplicate by name and description, but what's with the path of suite

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/CheckBoxGrid.kt
```diff
@@ -56,4 +71,22 @@ fun checkBoxGrid(options: List<String>) = fc<CheckBoxGridProps> { props ->
                 }
             }
     }
+    useEffect(emptyList<dynamic>()) {
+        js("var jQuery = require(\"jquery\")")
```
akuleshov7, line 75: why not unite it with the next line?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
petertrr: There may be suites with same names in the same repo, but with different test roots. Then we should save them separately. But if everything is same, other than description, it's probably the same test suite (do we have any checks in SAVE that description shouldn't be overriden if test suite name is the same? I think it would make sense).

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/service/TestDiscoveringService.kt
```diff
@@ -48,18 +48,30 @@ class TestDiscoveringService {
...
         project: Project?,
         rootTestConfig: TestConfig,
         propertiesRelativePath: String,
         testSuiteRepoUrl: String) = rootTestConfig
         .getAllTestConfigs()
-        .mapNotNull { it.getGeneralConfigOrNull()?.suiteName }
-        .map { suiteName ->
+        .asSequence()
+        .mapNotNull { it.getGeneralConfigOrNull() }
+        .filterNot { it.suiteName == null }
```
akuleshov7, line 60: Good check, but do we have suites with null name?

petertrr: Maybe in topmost configs, that already have general section, but expect suite names to be filled in child configs?

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
akuleshov7: so-so, but ok

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/CheckBoxGrid.kt
petertrr: I decided to put different expressions into different calls, in order not to make it even more of a mess.

PR #265, opened 2021-09-22
Create empty TestExecutions in prepareExecutionForStandard

kgevorkyan:
### What's done:
* Provide logic for `prepareExecutionForStandard`: save empty TestExecutions for standard test suites
* Update tests
* Increase block duration in standard test suites scheduler
* Temporary disable kotlin native cache

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
@@ -5,6 +5,8 @@ import org.cqfn.save.backend.service.TestService
 import org.cqfn.save.test.TestDto
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
@@ -40,6 +42,10 @@ class TestController {
      */
     @GetMapping("/getTestBatches")
     fun testBatches(@RequestParam agentId: String) = testService.getTestBatches(agentId)
+
+    @GetMapping("/getTestsWithTestSuiteId")
+    fun getAllTestsByTestSuiteId(@RequestParam testSuiteId: Long) =
+        ResponseEntity.status(HttpStatus.OK).body(testService.findTestByTestSuiteId(testSuiteId))
 
     companion object {
         private val log = LoggerFactory.getLogger(TestController::class.java)
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
@@ -28,7 +28,8 @@ class TestController {
      */
     @PostMapping("/initializeTests")
     fun initializeTests(@RequestBody testDtos: List<TestDto>, @RequestParam(required = false) executionId: Long?) {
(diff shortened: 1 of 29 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
```diff
@@ -33,6 +33,18 @@ class TestController {
         executionId?.let { testExecutionService.saveTestExecution(executionId, testsIds) }
     }
 
+    /**
+     * @param executionId ID of the [Execution], during which these tests will be executed
+     * @param testSuiteId ID of the [TestSuite], for which there will be created execution in DB
+     */
+    @Suppress("UnsafeCallOnNullableType")
+    @PostMapping("/getAllTestsByTestSuiteIdAndSaveExecution")
+    fun getAllTestsByTestSuiteIdAndSaveExecution(@RequestBody executionId: Long, @RequestParam testSuiteId: Long) {
```
petertrr, line 42:
```suggestion
    fun getAllTestsByTestSuiteIdAndSaveExecution(@RequestParam executionId: Long, @RequestParam testSuiteId: Long) {
```
Why not param?

petertrr, line 42: Also, why is endpoint called `get...`, if it's `POST` and returns nothing?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
```diff
@@ -57,7 +57,7 @@ class TestSuitesService {
      * @return all standard test suites with specific name
      */
     fun findStandardTestSuitesByName(name: String) =
-            testSuiteRepository.findAllByNameAndType(name, TestSuiteType.STANDARD).map { it.toDto() }
+            testSuiteRepository.findAllByNameAndType(name, TestSuiteType.STANDARD)
```
petertrr, line 60: Why did you change return type?

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestRepository.kt
```diff
@@ -12,4 +12,6 @@ import java.util.Optional
 @Suppress("MISSING_KDOC_CLASS_ELEMENTS", "MISSING_KDOC_ON_FUNCTION")
 interface TestRepository : BaseEntityRepository<Test> {
     fun findByHashAndFilePathAndTestSuiteId(hash: String, filePath: String, testSuiteId: Long): Optional<Test>
+
+    fun findAllByTestSuiteIdIs(testSuiteId: Long): List<Test>
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestController.kt
kgevorkyan: Because I'd like to pass to the uri only the `testSuiteId` and `executionId` by `BodyInserters` as more low level and static information

kgevorkyan: Because it also get tests from DB by `testSuiteId`, however it could be confusing, renamed

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestSuitesService.kt
kgevorkyan: For now I use this method also from preprocessor, but `dto` is not enough for him, since we need the `id`, which located only in `TesSuite`. So I change the return type and what about `orchestrator` - he called `it.toDto()` by itself

petertrr (approved):

PR #275, opened 2021-09-29
Preprocessor: Don't crash if save.properties doesn't exist in the repo

### LINKED ISSUE #274: Preprocessor: Don't crash if save.properties doesn't exist in the repo
In `getTestResourcesRootAbsolutePath` we are reading `save.properties`, which may not exist and it is valid. We should use defaults in this case.

This issue is related to #205, but that one is more general, while this one focuses on a specific bug.

Cheshiriks:
### What's done:
* Added check on save.properties exist
Close #274

### DIFF
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -187,7 +187,7 @@ class DockerService(private val configProperties: ConfigProperties) {
                     |RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
                     |RUN chmod +x $executionDir/$SAVE_AGENT_EXECUTABLE_NAME
                     |RUN chmod +x $executionDir/$SAVE_CLI_EXECUTABLE_NAME
-                    |RUN chmod +x $executionDir/diktat-rules/src/test/resources/test/smoke/src/ktlint
+                    |RUN chmod +x $executionDir/diktat-rules/src/test/resources/test/smoke/ktlint
                 """
         )
         saveAgent.delete()
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -188,7 +188,7 @@ class DockerService(private val configProperties: ConfigProperties) {
                     |RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
                     |RUN chmod +x $executionDir/$SAVE_AGENT_EXECUTABLE_NAME
                     |RUN chmod +x $executionDir/$SAVE_CLI_EXECUTABLE_NAME
-                    |RUN find "$executionDir/diktat-rules/src/test/resources/test/smoke/ktlint" -type f -exec chmod +x {} \;
+                    |RUN find $executionDir/diktat-rules/src/test/resources/test/smoke/ktlint -type f -exec chmod +x {} \;
                 """
         )
         saveAgent.delete()
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -183,6 +183,7 @@ class DockerService(private val configProperties: ConfigProperties) {
             imageName = imageName(execution.id!!),
             baseDir = resourcesPath,
             resourcesPath = executionDir,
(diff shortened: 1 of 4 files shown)


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -449,11 +449,14 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
     private fun getTestResourcesRootAbsolutePath(propertiesRelativePath: String,
                                                  projectRootRelativePath: String): String {
         // TODO: File should be provided without explicit naming of `save.propeties`, also
```
petertrr, line 451:
```suggestion
        // TODO: File should be provided without explicit naming of `save.propeties`
```

Cheshiriks: Fixed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -187,6 +187,7 @@ class DockerService(private val configProperties: ConfigProperties) {
                     |RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
                     |RUN chmod +x $executionDir/$SAVE_AGENT_EXECUTABLE_NAME
                     |RUN chmod +x $executionDir/$SAVE_CLI_EXECUTABLE_NAME
+                    |RUN chmod +x $executionDir/diktat-rules/src/test/resources/test/smoke/src/ktlint
```
akuleshov7: may be we will hardcode it for all binaries?

Cheshiriks: We are running the Save on this branch for testing

petertrr: Yes, we are testing a workaround. Will be implemented properly under #277

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -187,6 +187,7 @@ class DockerService(private val configProperties: ConfigProperties) {
                     |RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
                     |RUN chmod +x $executionDir/$SAVE_AGENT_EXECUTABLE_NAME
                     |RUN chmod +x $executionDir/$SAVE_CLI_EXECUTABLE_NAME
+                    |RUN chmod +x $executionDir/diktat-rules/src/test/resources/test/smoke/ktlint
```
petertrr, line 190: Pls change it to safe call (e.g. `if [ -f <path> ]` or `find -exec`), so that if there is no ktlint file, build doesn't fail. Also, pls add a comment that this is a temporary workaround link to #277 . This way we will be able to run diktat tests while #277 is still WIP.

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -183,10 +183,12 @@ class DockerService(private val configProperties: ConfigProperties) {
             imageName = imageName(execution.id!!),
             baseDir = resourcesPath,
             resourcesPath = executionDir,
+            // TODO: find ktlint this is a temporary workaround link to #277
             runCmd = """RUN apt-get update && env DEBIAN_FRONTEND="noninteractive" apt-get install -y libcurl4-openssl-dev tzdata && rm -rf /var/lib/apt/lists/*
                     |RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
                     |RUN chmod +x $executionDir/$SAVE_AGENT_EXECUTABLE_NAME
                     |RUN chmod +x $executionDir/$SAVE_CLI_EXECUTABLE_NAME
+                    |RUN find $executionDir/diktat-rules/src/test/resources/test/smoke/ktlint -type f -exec chmod +x {} \;
```
petertrr, line 191:
```suggestion
                    |RUN find . -name $executionDir/diktat-rules/src/test/resources/test/smoke/ktlint -type f -exec chmod +x {} \;
```
I tried your command, and it returns 1 if no files have been found. This will cause the entire docker build to fail

petertrr (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
Cheshiriks: fixed

petertrr (approved):

PR #287, opened 2021-10-04
Display tags in ExecutionView

kgevorkyan:
### What's done:
* Display tags in Execution view
* Add new column in the `Test` table
* Update corresponding tests

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
+++ b/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
@@ -23,5 +23,5 @@ data class TestExecutionDto(
     val startTimeSeconds: Long?,
     val endTimeSeconds: Long?,
     val testSuiteName: String? = null,
-    val tags: String? = null,
+    val tags: List<String>? = null,
 )
--- a/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
+++ b/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
@@ -11,6 +11,7 @@ import kotlinx.serialization.Serializable
  * @property endTimeSeconds
  * @property filePath
  * @property pluginName name of a plugin which will execute test at [filePath]
+ * @property tags list of tags of current test
  * @property testSuiteName a name of test suite, a test from which has been executed
  */
 @Serializable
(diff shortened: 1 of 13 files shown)


save-cloud-common/src/commonMain/kotlin/org/cqfn/save/test/TestDto.kt
```diff
@@ -11,13 +11,15 @@ import kotlinx.serialization.Serializable
...
  * @property pluginName name of a plugin which this test belongs to
+ * @property tags list of tags of current test
  */
 @Serializable
 data class TestDto(
     val filePath: String,
     val pluginName: String,
     val testSuiteId: Long,
     val hash: String,
+    val tags: String?,
```
petertrr, line 22: Maybe, we can store tags as lists in DTO and split when in `toDto` methods in entities?

petertrr (approved):

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
```diff
@@ -22,4 +23,5 @@ data class TestExecutionDto(
     val startTimeSeconds: Long?,
     val endTimeSeconds: Long?,
     val testSuiteName: String? = null,
+    val tags: List<String>? = null,
```
petertrr, line 26:
```suggestion
    val tags: List<String> = emptyList(),
```
One of the reasons I suggested collection is that you can make it non-nullable

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/test/TestDto.kt
```diff
@@ -11,13 +11,15 @@ import kotlinx.serialization.Serializable
...
  * @property pluginName name of a plugin which this test belongs to
+ * @property tags list of tags of current test
  */
 @Serializable
 data class TestDto(
     val filePath: String,
     val pluginName: String,
     val testSuiteId: Long,
     val hash: String,
+    val tags: List<String>?,
```
petertrr, line 22:
```suggestion
    val tags: List<String> = emptyList(),
```

petertrr (approved):

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/TestExecutionDto.kt
kgevorkyan: Yes, good advise, fixed

PR #376, opened 2021-11-01
Fix count tests for execution

Cheshiriks:
What's done:
* fixed count tests fot execution


petertrr:
Why did you change `gradlew.bat`?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
```diff
@@ -55,14 +55,11 @@ fun executionStatistics(classes: String = "") = fc<ExecutionStatisticsProps> { p
...
  * @return a functional react component
  */
-fun executionTestsNotFound() = fc<ExecutionStatisticsProps> { props ->
-    val totalTests = props.executionDto?.run {
-        runningTests + passedTests + failedTests + skippedTests
-    } ?: 0
-
-    if (totalTests == 0L) {
+fun executionTestsNotFound(count: Int?) = fc<ExecutionStatisticsProps> {
+    if (count == 0) {
```
petertrr, line 62: Are we doing it as a workaround, while coutners in execution are unreliable?

Cheshiriks: Counters in execution do not take into account many statuses

Cheshiriks:
> Why did you change `gradlew.bat`?

I didn't change gradlew.bat

petertrr:
> > Why did you change `gradlew.bat`?
> 
> I didn't change gradlew.bat

Well, take a look at `Files changed' tab then, it's there

kgevorkyan:
> > > Why did you change `gradlew.bat`?
> > 
> > 
> > I didn't change gradlew.bat
> 
> Well, take a look at `Files changed' tab then, it's there

I think it's just permissions, when you did `chmod +777` several days ago, you can restore them

Cheshiriks:
> > > > Why did you change `gradlew.bat`?
> > > 
> > > 
> > > I didn't change gradlew.bat
> > 
> > 
> > Well, take a look at `Files changed' tab then, it's there
> 
> I think it's just permissions, when you did `chmod +777` several days ago, you can restore them

Fixed

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
petertrr: Then we should also use this value instead of `totalTests` in `fun executionStatistics`, because the same problem applies there: even if all four counters are correct, their sum may be different from what's actually displayed in the table, because we currently have 7 statuses and only 4 counters

Cheshiriks: Fixed

petertrr (approved):

PR #377, opened 2021-11-01
Hot fix for gitDto

akuleshov7:
### What's done:
- if git url was provided, the creadentials were not passed to the backend. Now fixed


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
```diff
@@ -193,7 +193,7 @@ class ProjectView : RComponent<ProjectExecutionRouteProps, ProjectViewState>() {
 
         if (state.isFirstTypeUpload == true) {
             gitUrlFromInputField?.let {
-                val newGitDto = GitDto(url = it)
+                val newGitDto = GitDto(url = it, username = gitDto?.username, password = gitDto?.password)
```
petertrr, line 196:
```suggestion
                val newGitDto = gitDto?.copy(url = it) ?: GitDto(url = it)
```
Why not use `copy` in case we add other important fields in the future?

petertrr (approved):

PR #381, opened 2021-11-02
Hotfix for the "create test project" button

akuleshov7:
### What's done:
- as git credentials are optional, we should not fail while creating it

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
@@ -1,3 +1,4 @@
+
 /**
  * A view with project creation details
  */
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
```diff
@@ -224,23 +225,20 @@ class CreationView : RComponent<PropsWithChildren, ProjectSaveViewState>() {
         val gitUser = fieldsMap[InputTypes.GIT_USER]
         if (gitUser.isNullOrBlank() || gitUser.trim().matches(".*\\s.*")) {
             setState { isValidGitUser = false }
-            valid = false
```
akuleshov7, line 227: this validation is not needed as it is optional, will make a better validation in a different issue

petertrr (approved):

Issue #183, opened 2021-07-15
Display execution logs on frontend


petertrr:
Now we save them in filesystem, but they should be accessible via frontend

PR #384, opened 2021-11-02
Frontend: Add view with verbose test execution data

### LINKED ISSUE #368: Frontend: Add page with verbose debug info
Continuation of #183. All required info is already available in `TestResultDebugInfo`, should be displayed

petertrr:
### What's done:
* Added endpoint in TestExecutionController
* Created Requests.kt for common HTTP requests
* Added TestExecutionDetailsView

image

Continuation of #183 , closes #368


kgevorkyan (approved):
May be we could also make this page colorized in the future to make it more `visual`

petertrr:
> May be we could also make this page colorized in the future to make it more `visual`

Sure, this is the initial version to be able to test it on some real data

PR #385, opened 2021-11-02
Bugfix: Illegal argument exception in rerun mode

kgevorkyan:
### What's done:
* Save the test suite ids in execution record, reuse only them in rerun mode
* Update tests


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -515,7 +515,13 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
             discoverAndSaveTestSuites(project, rootTestConfig, testRootPath, gitUrl)
         }
         .flatMap { (rootTestConfig, testSuites) ->
-            initializeTests(testSuites, rootTestConfig, executionId)
+            val testSuiteIds = testSuites.map { it.id!! }.toMutableList()
+            testSuiteIds.sort()
```
petertrr, line 519:
```suggestion
            val testSuiteIds = testSuites.map { it.id!! }.sorted()
```

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -140,9 +140,8 @@ class ExecutionController(private val executionService: ExecutionService,
             "Can't rerun execution $id, project ${execution.project.name} has no associated git address"
         }
         val testRootPath = if (executionType == ExecutionType.GIT) {
-            execution.testSuiteIds?.let {
-                require(it == "ALL") { "Only executions with \"ALL\" tests suites from a GIT project are supported now" }
-                testSuitesService.findTestSuitesByProject(execution.project)
+            execution.testSuiteIds?.split(", ")?.map {
+                testSuitesService.findTestSuiteById(it.toLong()).get()
```
petertrr, line 144: Please add `orElseThrow` with description, which IDs don't exists. If we accidentally clean up something and it will simply start failing here, it'll be hard to debug.

petertrr (commented):
Do we also need to update our testing data (`db/v-1/insert...`)?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
kgevorkyan: `get` method throws exceptions by itself in case when there is no such element,
however I added `orElseThrow`, it will be easy to find error by our message

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
kgevorkyan: Since all tests were passed, I didn't think about it. Also the records in DB with ids `ALL`  don't bother me, at least.

Any other reasons to make changeset?

kgevorkyan: About code suggestion - fixed

kgevorkyan: Actually, during the execution, we update record twice, in `upload` function and now in `prepareForGit`, in first case we actually will save `ALL`, but now it's basically the stub, and will be rewritten by actual ids, when we will discover tests.
So 'ALL' could be hold in case of some exceptions between `upload` and  `prepareForGit`

Tests data updated

petertrr (approved):

PR #388, opened 2021-11-02
Backend: Fix for overlapping batches and counters update

petertrr:
### What's done:
* Use explicit programmatic transaction in `TestService` and `TestExecutionService` for operations under `synchronized`

Apparently, I was using `@Transactional` incorrectly, when trying to wrap batch retrieval into a dedicated transaction. This PR fixes that by using `TransactionTemplate` for one operation. Tested  it several times on data, which caused incorrect counters behavior before, and it looks like all good.

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
@@ -3,7 +3,6 @@ package org.cqfn.save.backend.service
 import org.cqfn.save.agent.TestExecutionDto
 import org.cqfn.save.backend.repository.AgentRepository
 import org.cqfn.save.backend.repository.ExecutionRepository
-import org.cqfn.save.backend.repository.TestDataFilesystemRepository
 import org.cqfn.save.backend.repository.TestExecutionRepository
 import org.cqfn.save.backend.repository.TestRepository
 import org.cqfn.save.backend.utils.secondsToLocalDateTime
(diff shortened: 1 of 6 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestService.kt
```diff
@@ -79,7 +77,9 @@ class TestService {
         val execution = agent.execution
         val lock = locks.computeIfAbsent(execution.id!!) { Any() }
         return synchronized(lock) {
-            val testExecutions = getTestExecutionsBatchByExecutionIdAndUpdateStatus(execution)
+            val testExecutions = transactionTemplate.execute {
+                getTestExecutionsBatchByExecutionIdAndUpdateStatus(execution)
+            }!!
```
akuleshov7, line 82: If no executions found we will return null here, no?

petertrr: Should be an empty list

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -25,10 +28,13 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
                            private val testRepository: TestRepository,
                            private val agentRepository: AgentRepository,
                            private val executionRepository: ExecutionRepository,
-                           private val testDataFilesystemRepository: TestDataFilesystemRepository,
+                           transactionManager: PlatformTransactionManager,
 ) {
     private val log = LoggerFactory.getLogger(TestExecutionService::class.java)
     private val locks: ConcurrentHashMap<Long, Any> = ConcurrentHashMap()
+    private val transactionTemplate = TransactionTemplate(transactionManager).apply {
+        propagationBehavior = TransactionDefinition.PROPAGATION_REQUIRES_NEW
```
akuleshov7, line 36: Why not REQUIRED?

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
petertrr: Because I want to perform this operation in a transaction, that will be completed while lock is acquired. Otherwise, there will be a condition, when different threads access test executions with incorrect statuses.

PR #390, opened 2021-11-03
Hotfix: Capture filePath as repeated parameter

petertrr:
Other than `exact = false`, react-router requires regex-like setting for repeated params group

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
@@ -135,7 +135,7 @@ class App : RComponent<PropsWithChildren, AppState>() {
                         }
                         Route {
                             attrs {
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
```diff
@@ -135,7 +135,7 @@ class App : RComponent<PropsWithChildren, AppState>() {
                         }
                         Route {
                             attrs {
-                                path = arrayOf("/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath")
+                                path = arrayOf("/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath+")
```
akuleshov7, line 138: Why + not *?

petertrr: There should be at least file name, so at least one part in the repeated group

akuleshov7 (approved):

PR #394, opened 2021-11-03
Refactoring for the project view

akuleshov7:
### What's done:
- fixed an issue with validation of git URL with tests (with the default value)
- refactored project view

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/ExecutionRequest.kt
+++ b/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/ExecutionRequest.kt
@@ -37,8 +37,7 @@ sealed class ExecutionRequestBase {
 data class ExecutionRequest(
     override val project: Project,
     val gitDto: GitDto,
-    // empty testRootPath is not valid and user needs to pass "." instead. Let's make it as a default by for now.
-    val testRootPath: String = ".",
+    val testRootPath: String,
     override val sdk: Sdk,
     val executionId: Long?,
 ) : ExecutionRequestBase()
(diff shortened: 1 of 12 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
```diff
@@ -104,7 +104,7 @@ enum class GitConnectionStatusEnum {
 @OptIn(ExperimentalJsExport::class)
 class CreationView : RComponent<PropsWithChildren, ProjectSaveViewState>() {
     private val fieldsMap: MutableMap<InputTypes, String> = mutableMapOf()
-    private lateinit var responseFromCreationProject: Response
+    private var responseFromCreationProject: Response? = null
```
akuleshov7, line 107: lateinit here caused a clash with a courutines, I do not have any ideas why by for now

petertrr: My idea - after execution of the main coroutine, `invokeOnCompletion` somehow still had a stale copy of `responseFromCreationProject` variable, which was unitialized. I think `invokeOnCompletion` is not needed here at all, because we can get response and immediately perform redirect from within coroutine body.

akuleshov7: no, that does not help :)
there is a failure on the "post" part

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/ExecutionRequest.kt
```diff
@@ -37,8 +37,7 @@ sealed class ExecutionRequestBase {
 data class ExecutionRequest(
     override val project: Project,
     val gitDto: GitDto,
-    // empty testRootPath is not valid and user needs to pass "." instead. Let's make it as a default by for now.
-    val testRootPath: String = ".",
+    val testRootPath: String,
```
akuleshov7, line 40: I moved this logic to the FE

PR #395, opened 2021-11-11
Frontend: Display start and end time of execution

### LINKED ISSUE #391: Frontend: Display start and end time of execution
On both HistoryView (now 'Date' is ambiguous) and ExecutionView

Cheshiriks:
What's done:
* added start time
Closes #391

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
@@ -225,7 +225,7 @@ class HistoryView : RComponent<HistoryProps, HistoryViewState>() {
                         }
                     }
                 }
-                column("checkBox", "") { cellProps ->
+                column("check-box", "") { cellProps ->
                     buildElement {
                         td {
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -86,7 +86,20 @@ class HistoryView : RComponent<HistoryProps, State>() {
                         }
                     }
                 }
-                column("date", "Date") {
+                column("startDate", "Start time") {
+                    buildElement {
+                        td {
+                            a(href = getHrefToExecution(it.value.id)) {
+                                +(it.value.startTime.let {
+                                    Instant.fromEpochSeconds(it, 0)
```
petertrr, line 94: Please extract date formatting into a function, because it's the same for two columns

Cheshiriks: Fixed

petertrr (approved):

PR #402, opened 2021-11-19
Frontend: After rerun request is submitted, refresh the page

### LINKED ISSUE #386: Frontend: After rerun request is submitted, refresh the page

Cheshiriks:
What's done:
* added refresh after rerun
Closes #386

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
@@ -136,6 +136,7 @@ class ExecutionView : RComponent<ExecutionProps, ExecutionState>() {
                                 undefined
                             )
                         }.invokeOnCompletion {
+                            window.location.reload()
                             window.alert("Rerun request successfully submitted")
                         }
                     }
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
```diff
@@ -136,6 +136,7 @@ class ExecutionView : RComponent<ExecutionProps, ExecutionState>() {
                                 undefined
                             )
                         }.invokeOnCompletion {
+                            window.location.reload()
                             window.alert("Rerun request successfully submitted")
```
petertrr, line 140: I think this should be the other way around: at first user sees an alert, and once alert is dismissed, only then is the page reloaded

petertrr (changes requested):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
Cheshiriks: Agree. Fixed

petertrr (approved):

PR #423, opened 2021-11-25
Fix for standard mode: move additional files into proper place

kgevorkyan:
### What's done:
* Move additional files during standard mode into proper place
* Change links to actual repo over the project
* Update save-cli version and it's API

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
@@ -117,7 +117,8 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
         val executionId = agent.execution.id!!
         val lostTests: MutableList<TestExecutionDto> = mutableListOf()
         val counters = Counters()
-        testExecutionsDtos.forEach { testExecDto ->
+        testExecutionsDtos.forEach { //testExecDto ->
+            val testExecDto = it.copy(filePath = if (it.filePath.startsWith("STANDARD_")) it.filePath.dropWhile { it != '/' }.drop(1) else it.filePath)
             val foundTestExec = testExecutionRepository.findByExecutionIdAndTestPluginNameAndTestFilePath(
                 executionId,
                 testExecDto.pluginName,
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
@@ -114,7 +114,7 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
         val agent = requireNotNull(agentRepository.findByContainerId(agentContainerId)) {
             "Agent with containerId=[$agentContainerId] was not found in the DB"
         }
-        println("\n\n\nCurr agent ${agentContainerId}")
+
         val executionId = agent.execution.id!!
         val lostTests: MutableList<TestExecutionDto> = mutableListOf()
         val counters = Counters()
@@ -125,6 +125,7 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
                 testExecDto.pluginName,
                 testExecDto.filePath
             )
+            val testExecutionId = foundTestExec.get().id
             foundTestExec.also {
                 if (it.isEmpty) {
                     log.error("Test execution $testExecDto for execution id=$executionId was not found in the DB")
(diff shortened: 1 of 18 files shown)


kgevorkyan:
The standard mode won't work until https://github.com/diktat-static-analysis/save-cloud/issues/333

However, I remove manually the extra command line args for save-cli and it worked fine, both for standard mode and rerun of standard mode

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -114,15 +120,24 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
+                it.filePath
+            }
+            val testExecDto = it.copy(filePath = filePathWithDroppedPrefixForStandardMode)
             val foundTestExec = testExecutionRepository.findByExecutionIdAndTestPluginNameAndTestFilePath(
                 executionId,
                 testExecDto.pluginName,
                 testExecDto.filePath
             )
+            var testExecutionId: Long? = null
+            foundTestExec.ifPresent { testExecutionId = it.id }
```
petertrr, line 140:
```suggestion
            val testExecutionId: Long? = foundTestExec.map { testExecutionId = it.id }.orElse(null)
```

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -114,15 +120,24 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
             "Agent with containerId=[$agentContainerId] was not found in the DB"
         }
+
         val executionId = agent.execution.id!!
         val lostTests: MutableList<TestExecutionDto> = mutableListOf()
         val counters = Counters()
-        testExecutionsDtos.forEach { testExecDto ->
+        testExecutionsDtos.forEach {
+            val filePathWithDroppedPrefixForStandardMode = if (it.filePath.startsWith(PREFIX_FOR_SUITES_LOCATION_IN_STANDARD_MODE)) {
+                it.filePath.dropWhile { it != '/' }.drop(1)
```
petertrr, line 129: Please add some comments, that the first branch is for `STANDARD` mode and the secobd - for `GIT` mode

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -175,10 +177,19 @@ class DockerService(private val configProperties: ConfigProperties) {
         val testSuitesDir = resourcesPath.resolve(standardTestSuiteDir)
         // copy corresponding standard test suites to resourcesRootPath dir
         copyTestSuitesToResourcesPath(testSuitesForDocker, testSuitesDir)
+
+        // move additional files, which were downloaded into the root dit to the execution dir for standard suites
+        execution.additionalFiles?.split(";")?.filter { it.isNotBlank() }?.forEach {
+            val additionalFilePath = Paths.get(resourcesPath.resolve(File(it).name).absolutePath)
```
petertrr, line 183: Do we always have additional files in the root of resourcesPath? Maybe, having `resolve(it)` will be more reliable?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -175,10 +177,19 @@ class DockerService(private val configProperties: ConfigProperties) {
...
+            val destination = Paths.get(testSuitesDir.absolutePath).resolve(File(it).name)
+            log.info("Move additional file $additionalFilePath into $destination")
+            Files.move(additionalFilePath, destination, StandardCopyOption.REPLACE_EXISTING)
```
petertrr, line 186: When are we copying those files in `resourcesPath`? Maybe, we can move them in the proper directory at that point already?

kgevorkyan: We download files into `resourcesPath` during `uploadBin`, and at those moment  we doesn't know anything about `standardTestSuiteDir`, so I decide to provide logic of movement here

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kgevorkyan: Actually, in `execution.additionalFiles` we have the full paths to the files from our storage `/home/cnb/files/...`, but we also know, that during `uploadBin` execution we have already this files in the `resourcesPath`, with the same `.name` as in `execution.additionalFiles`

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
kgevorkyan: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
kgevorkyan: Added

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
petertrr: OK

petertrr (approved):

Issue #417, opened 2021-11-24
Secure backend's endpoints with correct permission checks


petertrr:
Should be done after #470 and #495
- Only some endpoints are accessible for unauthorized users (is configurable from `WebSecurityConfig` like in #498)
- Only project's owner can delete stuff and change description and submit executions

List of required roles:
- `ROLE_ADMIN` - can delete anything, can view anything
- `ROLE_PROJECT_OWNER` - this is the user, that created the project and is linked to it in the DB
- `ROLE_PROJECT_ADMIN` - can be added by project owner, can start and manage executions, can't delete execution and project

Implementation:
- Create roles in DB
- Create roles hierarchy for spring security
- Protect methods according to roles: https://www.baeldung.com/spring-security-method-security#5-method-security-meta-annotation, https://www.baeldung.com/spring-security-create-new-custom-security-expression

petertrr:
Thoughts from #424:
[* Convert backend into a resource server to properly consume tokens? https://docs.spring.io/spring-security/site/docs/5.2.1.RELEASE/reference/htmlsingle/#webflux-oauth2-resource-server. Then, hardcoded endpoints can be removed from gateway's security config, and backend can check if endpoint is available for a current user. After this PR is merged, need to be implemented in a separate issue. Also can use CircuitBreaker filter to redirect to login page on status 403

PR #424, opened 2021-11-25
Add api-gateway component

### LINKED ISSUE #415: Support OAuth tokens

### LINKED ISSUE #419: Add API Gateway component

petertrr:
### What's done:
* Add new component
* User details are passed from GW to backend via `TokenRelay` filter
* Frontend: prefix all backend queries with `/api`; update `dev-server.js` accordingly
* Prepare docs about deployment of the new component
* Configuration for OAuth providers, including secrets, should be provided via particular deployment system

### TODO:
* Make part of backend's API `/internal` and proxy only `/api` (separate issue, or part of migration to OpenAPI specification)
* Metrics from gateway are not being scraped, because endpoint is protected too - we can ignore it for now.
* #443
* #417

Closes #419 , closes #415


gradle/libs.versions.toml
```diff
@@ -9,6 +9,7 @@ kotlinx-coroutines = "1.5.2"
 kotlin-wrappers = "0.0.1-pre.247-kotlin-1.5.31"
 spring-boot = "2.6.0"
 spring-security = "5.6.0"
+spring-cloud = "3.1.0-M3"
```
petertrr, line 12: This weird version is because spring-cloud with support for spring-boot 2.6.x is not released yet 🤷

save-deploy/reverse-proxy.conf
```diff
@@ -11,7 +11,7 @@ server {
     error_log /var/log/nginx/reverse-error.log;
 
     location / {
-        proxy_pass http://127.0.0.1:5000;
+        proxy_pass http://127.0.0.1:5300;
```
petertrr, line 14: I won't remove nginx config completely just yet, need to figure out if gateway can replace it completely

kgevorkyan (approved):

api-gateway/src/main/resources/application.yml
```diff
@@ -0,0 +1,43 @@
...
+  cloud:
+    gateway:
+      routes:
+        - id: api_route
+          uri: http://backend:5000
+          predicates:
+            - Path=/api/**
+          filters:
+            - StripPrefix=1
+            # If SESSION cookie is passed to downstream, it is then removed, because downstream discards it
```
petertrr, line 23: It was needed for credentials-based authentication, need to check if it can be removed for OAuth

gradle/libs.versions.toml
petertrr: It's released

api-gateway/src/main/resources/application.yml
petertrr: Spring still sets `SESSION` cookie, so I'll leave it here at least till #417

PR #429, opened 2021-11-29
Fix: Compatibility of standard mode with actual version of SAVE-cli

kgevorkyan:
### What's done:
* Fix: Compatibility of standard mode with actual version of SAVE-cli

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -20,6 +20,7 @@ import org.cqfn.save.domain.TestResultLocation
 import org.cqfn.save.domain.TestResultStatus
 import org.cqfn.save.plugins.fix.FixPlugin
 import org.cqfn.save.reporter.Report
+import org.cqfn.save.utils.STANDARD_TEST_SUITE_DIR
 
 import generated.SAVE_CLOUD_VERSION
 import io.ktor.client.HttpClient
@@ -156,7 +157,14 @@ class SaveAgent(internal val config: AgentConfiguration,
 
     @Suppress("MagicNumber")
     private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, "", config.logFilePath.toPath(), 100_000L)
+        .exec(
+            config.cliCommand.let {
+                if (!it.contains(STANDARD_TEST_SUITE_DIR)) "$it $cliArgs" else it
+            } + " --report-type json --result-output file",
+            "",
+            config.logFilePath.toPath(),
+            100_000L
+        )
 
     @Suppress("TOO_MANY_LINES_IN_LAMBDA")
     private fun CoroutineScope.readExecutionResults(jsonFile: String): List<TestExecutionDto> {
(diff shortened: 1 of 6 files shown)


save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -156,7 +157,14 @@ class SaveAgent(internal val config: AgentConfiguration,
 
     @Suppress("MagicNumber")
     private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, "", config.logFilePath.toPath(), 100_000L)
+        .exec(
+            config.cliCommand.let {
+                if (!it.contains(STANDARD_TEST_SUITE_DIR)) "$it $cliArgs" else it
```
kgevorkyan, line 162: `$cliArgs` here could be not empty only in the Git mode and this variable contain the `testRootPath` + set of tests in those case

in standard mode just use command, which we created in `DockerService.kt`

save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -156,7 +157,16 @@ class SaveAgent(internal val config: AgentConfiguration,
 
     @Suppress("MagicNumber")
     private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(config.cliCommand.let { if (cliArgs.isNotEmpty()) "$it $cliArgs" else it }, "", config.logFilePath.toPath(), 100_000L)
+        .exec(
+            config.cliCommand.let {
+                // cliArgs could be not empty only in the Git mode and this variable contain the `testRootPath` + set of tests in this case
+                // in standard mode just use command, which we created in DockerService, it already contain all necessary configuration
+                if (!it.contains(STANDARD_TEST_SUITE_DIR)) "$it $cliArgs" else it
```
petertrr, line 164: Maybe, we can keep construction of `cliArgs` in orchestrator and concat them here unconditionally? At least they will always have reporter, so will never be empty. Or it there some additional advantage in moving this logic here?

kgevorkyan: We still have 3 places with construction of cli command:

* `TestService` - takes `testRootPath`
* `AgentService` - concatenate  `testRootPath` with set of tests
* `SaveAgent` - read execution cmd == `./$SAVE_CLI_EXECUTABLE_NAME` and additional flags from `agentPropertiesFile`, which was constructed in `DockerService`.  In particular, this file contain the cli command for standard test suites, in standard mode.

In orchestrator we basically doesn't read `agentPropertiesFile` anywhere more, and the first place, when we are reading this file is `SaveAgent`, so I use this check here.

Ideally, our logic should be reworked, and don't use 3 different places for construction of one command,
but it looks as a quite big work, which probably could be done in future

petertrr: Oh, I've forgot about agent.properties

> Ideally, our logic should be reworked, and don't use 3 different places for construction of one command,
but it looks as a quite big work, which probably could be done in future

Totally agree

petertrr (approved):

PR #434, opened 2021-11-30
Support processing of uploaded zip archives

kgevorkyan:
### What's done:
* Support processing of uploaded zip archives - unzip zip archives into docker image among provided additional files

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,7 +6,6 @@ okio = "3.0.0-alpha.10"
 serialization = "1.3.1"
 kotlinx-datetime = "0.3.1"
 kotlinx-coroutines = "1.5.2"
-kotlin-zip4j = "2.9.1"
 kotlin-wrappers = "0.0.1-pre.265-kotlin-1.5.31"
 spring-boot = "2.6.0"
 spring-security = "5.6.0"
@@ -33,6 +32,7 @@ okhttp3 = "4.9.3"
 reckon = "0.13.0"
 commons-compress = "1.21"
 picocli = "4.6.1"
+zip4j = "2.9.1"
 
 [plugins]
 kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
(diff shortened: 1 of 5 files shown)


gradle/libs.versions.toml
```diff
@@ -6,6 +6,7 @@ okio = "3.0.0-alpha.10"
 serialization = "1.3.1"
 kotlinx-datetime = "0.3.1"
 kotlinx-coroutines = "1.5.2"
+kotlin-zip4j = "2.9.1"
```
petertrr, line 9:
```suggestion
zip4j = "2.9.1"
```
This is pure Java library? Why prefix it with `kotlin-`?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -240,6 +254,63 @@ class DockerService(private val configProperties: ConfigProperties) {
...
+
+    @Suppress("MAGIC_NUMBER", "LONG_NUMERICAL_VALUES_SEPARATED", "MagicNumber")
+    private fun File.isValidZipArchive(): Boolean {
+        var fileSignature: Int
+        try {
+            RandomAccessFile(this, "r").use { raf -> fileSignature = raf.readInt() }
+        } catch (e: IOException) {
+            log.error("Error during checking file ${this.name} for being an archive: ${e.message}")
+            fileSignature = 0
+        }
```
petertrr, line 300:
```suggestion
        val fileSignature: Int = try {
            RandomAccessFile(this, "r").use { raf -> raf.readInt() }
        } catch (e: IOException) {
            log.error("Error during checking file ${this.name} for being an archive: ${e.message}")
            0
        }
```

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -240,6 +254,63 @@ class DockerService(private val configProperties: ConfigProperties) {
...
+        // check for 'magic numbers', i.e. the data used to identify or verify the content of a file
+        return fileSignature == 0x504B0304 || fileSignature == 0x504B0506 || fileSignature == 0x504B0708
```
petertrr, line 302: Doesn't the library handle such checks by itself? Why is this check needed?

gradle/libs.versions.toml
kgevorkyan: Fixed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kgevorkyan: Removed this method

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kgevorkyan: Yeah, let it be covered by library

petertrr (approved):

PR #439, opened 2021-12-01
If the test execution is in the PENDING status we should show "Loading"

### LINKED ISSUE #418: If the test execution is in the PENDING status we should not show our awesome gif with cat
Right now when tests are not discovered yet, but the status of run is PENDING we show the gif "TESTS NOT FOUND!"
We should write someting like "Loading" instead

Cheshiriks:
What's done:
* changes logic for show info to execution in the PENDING status
Closes #418

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
@@ -61,14 +61,20 @@ fun executionStatistics(classes: String = "") = fc<ExecutionStatisticsProps> { p
  * @param count tests for execution
  * @return a functional react component
  */
-fun executionTestsNotFound(count: Int?) = fc<ExecutionStatisticsProps> {
-    if (count == 0) {
+fun executionTestsNotFound(count: Int?) = fc<ExecutionStatisticsProps> { props ->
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
```diff
@@ -71,5 +71,11 @@ fun executionTestsNotFound(count: Int?) = fc<ExecutionStatisticsProps> {
                 +"Tests not found!"
             }
         }
+    } else if (count == 0 && props.executionDto?.status == ExecutionStatus.PENDING) {
+        div("d-sm-flex align-items-center justify-content-center mb-4 mt-2") {
+            h1("h3 mb-0 text-gray-800") {
+                +"Loading ..."
```
petertrr, line 77:
```suggestion
                +"Execution is starting..."
```
It's not exactly 'loading', because while execution is still pending, there is nothing to load :)

petertrr (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ExecutionStatistics.kt
Cheshiriks: Agree, fixed)

PR #440, opened 2021-12-01
Minor: Improve performance of uploading files in standard mode and other small fixes

kgevorkyan:
### What's done:
* Improve performance of uploading files in standard mode and other minor fixes

### DIFF
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
@@ -37,7 +37,7 @@ fun getHostIp(hostname: String): String? {
  */
 fun copyRecursivelyWithAttributes(sourceDir: File, targetDir: File) {
     if (!sourceDir.exists()) {
-        throw FileNotFoundException("Source directory doesn't exists!")
+        throw FileNotFoundException("Source directory $sourceDir doesn't exist!")
     }
     sourceDir.walkTopDown().forEach { source ->
         val target = targetDir.resolve(source.relativeTo(sourceDir)).canonicalFile
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
@@ -5,6 +5,7 @@
 package org.cqfn.save.orchestrator
 
 import java.io.File
+import java.io.FileNotFoundException
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 
@@ -32,8 +33,12 @@ fun getHostIp(hostname: String): String? {
  *
  * @param sourceDir source directory
  * @param targetDir target directory
+ * @throws FileNotFoundException if source dir doesn't exists
  */
 fun copyRecursivelyWithAttributes(sourceDir: File, targetDir: File) {
+    if (!sourceDir.exists()) {
+        throw FileNotFoundException("Source directory doesn't exists!")
+    }
     sourceDir.walkTopDown().forEach { source ->
         val target = targetDir.resolve(source.relativeTo(sourceDir)).canonicalFile
         if (source.isDirectory) {
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
@@ -23,6 +23,7 @@ import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.core.io.ClassPathResource
 import org.springframework.http.MediaType
 import org.springframework.stereotype.Service
+import org.springframework.util.FileSystemUtils
 import org.springframework.web.reactive.function.BodyInserters
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.bodyToMono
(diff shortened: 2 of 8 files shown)


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -313,8 +313,10 @@ class DockerService(private val configProperties: ConfigProperties) {
                     .resolve(it.testRootPath)
                 )
             val currentSuiteDestination = destination.resolve("$PREFIX_FOR_SUITES_LOCATION_IN_STANDARD_MODE${it.testSuiteRepoUrl.hashCode()}_${it.testRootPath.hashCode()}")
-            log.info("Copying suite ${it.name} from $standardTestSuiteAbsolutePath into $currentSuiteDestination/...")
-            copyRecursivelyWithAttributes(standardTestSuiteAbsolutePath, currentSuiteDestination)
+            if (!currentSuiteDestination.exists()) {
```
kgevorkyan, line 316: Previously for test suites from one standard repository and with similar testRootPath we overwrite data any time, but 
at first it were extra actions (however for now our standard repos not really large, so it was not critical)
at second, damn `Files.copy` even with replace existing flag raise exceptions if it will try copy into not empty dir


[REPLACE_EXISTING – Performs the copy even when the target file already exists. If the target is a symbolic link, the link itself is copied (and not the target of the link). If the target is a non-empty directory, the copy fails with the DirectoryNotEmptyException exception.](https://www.codejava.net/java-se/file-io/java-nio-copy-file-or-directory-examples)

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -138,14 +139,15 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
...
+    @OptIn(ExperimentalFileSystem::class)
     @PostMapping(value = ["/uploadBin"], consumes = ["multipart/form-data"])
     fun uploadBin(
         @RequestPart executionRequestForStandardSuites: ExecutionRequestForStandardSuites,
         @RequestPart("file", required = true) files: Flux<FilePart>,
         @RequestPart("fileInfo", required = true) fileInfos: Flux<FileInfo>,
     ) = Mono.just(ResponseEntity("Clone pending", HttpStatus.ACCEPTED))
         .doOnSuccess { _ ->
-            files.zipWith(fileInfos).download(File("."))
+            files.zipWith(fileInfos).download(File(FileSystem.SYSTEM_TEMPORARY_DIRECTORY.toString()))
```
kgevorkyan, line 150: Performance improvements

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
```diff
@@ -32,8 +33,12 @@ fun getHostIp(hostname: String): String? {
  *
  * @param sourceDir source directory
  * @param targetDir target directory
+ * @throws FileNotFoundException if source dir doesn't exists
  */
 fun copyRecursivelyWithAttributes(sourceDir: File, targetDir: File) {
+    if (!sourceDir.exists()) {
+        throw FileNotFoundException("Source directory doesn't exists!")
```
petertrr, line 40:
```suggestion
        throw FileNotFoundException("Source directory $sourceDir doesn't exist")
```

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kgevorkyan: Yes, I think in some way, it was the lazy solution (if some files weren't removed before new execution, by this overwrite we were be able to not care about it). However, for these reasons it's better to use some manual deletion, which I added in the latest commit

I've tested git and standard modes with reruns, also I tested different executions, which used the same directories in docker image (because of the same initial data, provided from frontend), and all worked fine without any clashes

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/LinuxUtils.kt
kgevorkyan: Fixed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
kgevorkyan: > `so we won't always have !exists() == true here?`

Despite the fact, that I didn't met it before review commit too, now it shouldn't happened for sure

petertrr: Ok, great

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -302,8 +302,9 @@ class DockerService(private val configProperties: ConfigProperties) {
             .block()!!
     }?.map { it.toDto() } ?: emptyList()
 
-    @Suppress("UnsafeCallOnNullableType")
+    @Suppress("UnsafeCallOnNullableType", "TOO_MANY_LINES_IN_LAMBDA")
     private fun copyTestSuitesToResourcesPath(testSuitesForDocker: List<TestSuiteDto>, destination: File) {
+        destination.deleteRecursively()
```
petertrr, line 307: I suggest to use something like spring's `FileSystemUtils.deleteRecursively`, because it uses NIO under the hood

kgevorkyan: Changed

petertrr (approved):

PR #441, opened 2021-12-02
Make executable all files in archive, which marked executable

kgevorkyan:
### What's done:
* Create logic of marking files from archive executable
* Fixes: Save attributes during copying/movement files

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/utils/FileUtils.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/utils/FileUtils.kt
@@ -1,0 +1,23 @@
+package org.cqfn.save.utils
+
+import java.io.File
+import java.io.FileNotFoundException
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+
+/**
+ * Move [source] into [destinationDir], while also copying original file attributes
+ *
+ * @param source source file
+ * @param destinationDir target file
+ * @throws FileNotFoundException if source doesn't exists
+ */
+fun moveFileWithAttributes(source: File, destinationDir: File) {
+    if (!source.exists()) {
+        throw FileNotFoundException("Source file $source doesn't exist!")
+    }
+    val destination = destinationDir.resolve(source.name)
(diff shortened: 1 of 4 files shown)


akuleshov7 (approved):

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/utils/FileUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+ * @param destinationDir destination directory
+ * @throws FileNotFoundException if source doesn't exists
+ */
+fun moveFileWithAttributes(source: File, destinationDir: File) {
+    if (!source.exists()) {
+        throw FileNotFoundException("Source file $source doesn't exist!")
+    }
+
+    Files.copy(source.toPath(), destinationDir.resolve(source.name).toPath(), StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.COPY_ATTRIBUTES)
+    Files.delete(source.toPath())
```
petertrr, line 23: `Files.move`?

kgevorkyan: Unfortunately, `Files.move` doesn't support `StandardCopyOption.COPY_ATTRIBUTE` option, so I used this workaround method

petertrr: Hm, that's strange. They even have it described in doc: `An implementation may also attempt to copy other file attributes but is not required to fail if the file attributes cannot be copied.`

petertrr (approved):

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/utils/FileUtils.kt
kgevorkyan: Besides that I met exceptions `unsupported option`, here it's also can be seen https://docs.oracle.com/javase/tutorial/essential/io/move.html, that they support only `REPLACE_EXISTING` and `ATOMIC_MOVE` options

PR #450, opened 2021-12-06
Use Gradle JVM toolchains

### LINKED ISSUE #449: Configure JVM toolchain in gradle
https://blog.jetbrains.com/kotlin/2021/11/gradle-jvm-toolchain-support-in-the-kotlin-plugin/

save-cloud requires JDK >11, so we can add this into configuration to make it easier to perform a first-time build.

petertrr:
### What's done:
* Use gradle JVM toolchains
* Add gateway to `deploy_images.yaml`
* Add `RequiresOptIn` for `buildSrc`
* Enable `useOverride` by default and log a warning

Closes #449

### DIFF
--- a/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
+++ b/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
@@ -10,6 +10,7 @@ import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.withType
 import org.springframework.boot.gradle.tasks.bundling.BootBuildImage
 import java.io.ByteArrayOutputStream
+import java.io.File
 import java.nio.file.Files
 import java.nio.file.Paths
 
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -67,8 +68,14 @@ fun Project.createStackDeployTask(profile: String) {
             }
         }
 
-        val useOverride = (findProperty("useOverride") as String?).toBoolean()
         val configsDir = Paths.get("${System.getProperty("user.home")}/configs")
+        val useOverride = (properties.getOrDefault("useOverride", "true") as String).toBoolean()
+        val composeOverride = File("$configsDir/docker-compose.override.yaml")
+        if (useOverride && !composeOverride.exists()) {
+            logger.warn("Can't use override configuration, because ${composeOverride.canonicalPath} doesn't exist")
```
kgevorkyan, line 75: ?
```suggestion
            logger.warn("`useOverride` option is set to true, but can't use override configuration, because ${composeOverride.canonicalPath} doesn't exist")
```

kgevorkyan (approved):

PR #456, opened 2021-12-06
Fix sorting in TableComponent

petertrr:
Columns require proper accessor functions for react-table to sort them.

### What's done:
* Add explicit accessor functions in all tables
* Add polyfill for `path-browserify`

Caveats:
* I can't find a way to disable sorting on specific columns, that don't (at least for now) have valid accessors. E.g., testFilePath - I don't think we need it to be sortable. Or tags - tags is a collection, and filtering/sorting by tags should be a separate logic.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,7 +6,7 @@ okio = "3.0.0"
 serialization = "1.3.1"
 kotlinx-datetime = "0.3.1"
 kotlinx-coroutines = "1.5.2"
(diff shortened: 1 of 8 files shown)


akuleshov7:
This one is ready?

petertrr:
> This one is ready?

Not completely, I wanted to figure how to disable sorting for columns that shouldn't support it

akuleshov7 (approved):

PR #459, opened 2021-12-07
File upload progress bar

### LINKED ISSUE #301: Frontend: display file upload progress

sanyavertolet:
This PR closes #301 

## What's done:
* [ ] Added bar that shows file upload progress

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
@@ -14,8 +14,10 @@ import org.cqfn.save.frontend.externals.fontawesome.faUpload
 import org.cqfn.save.frontend.externals.fontawesome.fontAwesomeIcon
 import org.cqfn.save.frontend.utils.toPrettyString
 
+import csstype.Width
 import org.w3c.dom.HTMLInputElement
 import org.w3c.dom.HTMLSelectElement
+import react.CSSProperties
 import react.Props
 import react.PropsWithChildren
 import react.dom.attrs
@@ -66,6 +68,21 @@ external interface UploaderProps : PropsWithChildren {
      * state for the creation of unified confirmation logic
      */
     var confirmationType: ConfirmationType
+
+    /**
+     * General size of test suite in bytes
+     */
+    var suiteByteSize: Long
+
+    /**
+     * Bytes received by server
+     */
+    var bytesReceived: Long
+
+    /**
+     * Flag to handle uploading a file
+     */
+    var isUploading: Boolean
 }
 
 /**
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
```diff
@@ -80,7 +97,7 @@ fun fileUploader(
     onFileSelect: (HTMLSelectElement) -> Unit,
     onFileRemove: (FileInfo) -> Unit,
     onFileInput: (HTMLInputElement) -> Unit,
-    onExecutableChange: (file: FileInfo, checked: Boolean) -> Unit,
+    onExecutableChange: (file: FileInfo, checked: Boolean) -> Unit
```
petertrr, line 100: Why? Trailing comma is useful here (btw, we should configure diktat rule for it already)

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
```diff
@@ -152,6 +169,20 @@ fun fileUploader(
                         strong { +"Upload files:" }
                     }
                 }
+
+                div("progress") {
+                    attrs.hidden = !props.isUploading
+                    div("progress-bar progress-bar-striped progress-bar-animated") {
+                        attrs["style"] = kotlinext.js.jsObject<CSSProperties> {
+                            width = if (props.suiteByteSize != 0.toLong()) {
```
petertrr, line 177:
```suggestion
                            width = if (props.suiteByteSize != 0L) {
```

petertrr (changes requested):
If we upload one file and then upload another one, will the second progress bar start from `first.size` or from zero?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
sanyavertolet: lol, sorry, fully agree

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/FileUploader.kt
sanyavertolet: ok, fixed

sanyavertolet:
> If we upload one file and then upload another one, will the second progress bar start from `first.size` or from zero?

@petertrr it will start from first.size.
To my mind, it is not that important. It just seems nicer when we understand the full size of files (for example to approximate the time we need to wait till it is fully uploaded)

akuleshov7 (approved):

PR #463, opened 2021-12-08
Update liquibase-core to 4.6.2

### LINKED ISSUE #448: `LiquibaseException` on versions >4.5.0
```
Caused by: liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set db/v-1/insert-test-data/test-execution-insert.xml::test_execution-insert::frolov:
     Reason: liquibase.exception.DatabaseException: java.sql.SQLException: Parameter index out of range (7 > number of parameters, which is 6).
```

petertrr:
### What's done:
* Update dependency
* Fix changesets

Closes #448


petertrr:
@Cheshiriks please take a look at this. It seems, starting with liquibase 4.5.0 they had some change with `SEQUENCE` type, because now it's crashing on your `APP_SEQ` placeholders. Maybe, you know, what could've changed there?
Also, do we really need them? Without `id` column in CSV data is still inserted with autoincrementing IDs.

Cheshiriks (commented):
SEQUENCE can be configured to create a unique identifier across the entire database, rather than a single table.
This is convenient, but not necessary, you can remove

Cheshiriks (approved):

PR #465, opened 2021-12-08
Use `GITHUB_TOKEN` to authenticate requests for resources access in CI

### LINKED ISSUE #462: CI: Use `GITHUB_TOKEN` in CI to download resources from Github
download-plugin in Gradle can now hit API requests limit and it leads to things like
```
Egress is over the account limit. (HTTP status code: 503, URL: https://github.com/cqfn/diKTat/releases/download/v1.0.0-rc.2/diktat-1.0.0-rc.2.jar
```

petertrr:
### What's done:
* Add configuration

Closes #462


petertrr:
@akuleshov7 this fix doesn't work (credentials are not passed due to redirections) and is not required - the issue is only with github infrastructure

PR #472, opened 2021-12-09
Add columns for match and missing warnings

### LINKED ISSUE #454: Add columns for match and mismatch warnings
Add columns in test

Cheshiriks:
What's done:
* added columns for match and mismatch warnings
Closes #454

missing

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
@@ -18,8 +18,11 @@ import javax.persistence.ManyToOne
  * @property status status of test execution
  * @property startTime start time
  * @property endTime finish time
+ * @property missing missing warnings
+ * @property matched matched warnings
  */
 @Entity
+@Suppress("LongParameterList")
 class TestExecution(
 
     @ManyToOne
@@ -41,6 +44,10 @@ class TestExecution(
 
     var endTime: LocalDateTime?,
 
+    var missing: Int?,
+
+    var matched: Int?,
+
 ) : BaseEntity() {
     /**
      * Converts `this` to [TestExecutionDto]
@@ -57,5 +64,7 @@ class TestExecution(
         endTime?.toEpochSecond(ZoneOffset.UTC),
         test.testSuite.name,
         test.tags?.split(";")?.filter { it.isNotBlank() } ?: emptyList(),
+        missing,
+        matched,
     )
 }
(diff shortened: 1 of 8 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
```diff
@@ -199,6 +199,20 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
                         }
                     }
                 }
+                column(id = "missing", header = "Missing") {
+                    buildElement {
+                        td {
+                            +"${it.value.missing}"
```
petertrr, line 205: What if it's null? Also, it will always be null for fix-plugin - it needs to be handled correctly.

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
```diff
@@ -41,6 +44,10 @@ class TestExecution(
 
     var endTime: LocalDateTime?,
 
+    var missing: Int?,
```
petertrr, line 47:
```suggestion
    var missingWarnings: Int?,
```

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
```diff
@@ -41,6 +44,10 @@ class TestExecution(
...
+
+    var matched: Int?,
```
petertrr, line 49:
```suggestion
    var matchedWarnings: Int?,
```

petertrr (changes requested):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ExecutionView.kt
Cheshiriks: Yes. Instead of a null, we will output an empty string

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
Cheshiriks: Fixed

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/TestExecution.kt
Cheshiriks: Fixed

petertrr (approved):

PR #476, opened 2021-12-09
Improvements in standard mode - execute tests directly instead of execution by test suites

kgevorkyan:
### What's done:
* Improvements in standard mode - execute tests directly instead of execution by test suites
* Updates in tests

### DIFF
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
@@ -59,7 +59,7 @@ class AgentService {
                 .uri("/getTestBatches?agentId=$agentId")
                 .retrieve()
                 .bodyToMono<TestBatch>()
-                .map { batch -> batch.toHeartbeatResponse(agentId) }
+                .flatMap { batch -> batch.toHeartbeatResponse(agentId) }
                 .doOnSuccess {
                     if (it is NewJobResponse) {
                         updateAssignedAgent(agentId, it)
@@ -238,9 +238,9 @@ class AgentService {
                     .uri("/testSuite/${tests.first().testSuiteId}")
                     .retrieve()
                     .bodyToMono<TestSuite>()
-                    .map {
+                    .map { testSuite ->
                         val testPaths: MutableList<String> = mutableListOf()
-                        val isStandardMode = it.type == TestSuiteType.STANDARD
+                        val isStandardMode = testSuite.type == TestSuiteType.STANDARD
                         tests.forEach { test ->
                             if (isStandardMode) {
                                 webClientBackend.get()
@@ -259,10 +259,13 @@ class AgentService {
                                 testPaths.add(test.filePath)
                             }
                         }
+                        println("\n\n\nTESTS PATHS ${testPaths}")
+                        println("\nCMD ${suitesToArgs.values.first() +
+                                " " + testPaths.joinToString(" ")}")
                         NewJobResponse(
                             tests,
                             suitesToArgs.values.first() +
-                                    " " + testPaths)
+                                    " " + testPaths.joinToString(" "))
                     }
 
             } else {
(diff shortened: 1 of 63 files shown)


petertrr (changes requested):
We should set correct paths when returning a batch of tests from orchestrator

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
```diff
@@ -230,15 +234,62 @@ class AgentService {
...
+        // first, need to check the current mode, it could be done by looking of type of any test suite for current tests
+        return webClientBackend.get()
+            .uri("/testSuite/${tests.first().testSuiteId}")
+            .retrieve()
+            .bodyToMono<TestSuite>()
+            .flatMap { testSuite ->
+                isStandardMode = testSuite.type == TestSuiteType.STANDARD
+                if (isStandardMode) {
+                    // in standard mode for each test get proper prefix location, since we created extra directories
+                    Flux.fromIterable(tests).flatMap { test ->
```
petertrr, line 262: Group `tests` by `testSuiteId` to avoid a lot of equivalent requests?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
```diff
@@ -230,15 +234,62 @@ class AgentService {
...
+                        webClientBackend.get()
+                            .uri("/testSuite/${test.testSuiteId}")
+                            .retrieve()
+                            .bodyToMono<TestSuite>()
+                            .map {
+                                val testFilePathInStandardDir =
+                                        Paths.get(getLocationInStandardDirForTestSuite(it.toDto()))
+                                            .resolve(Paths.get(test.filePath))
+                                testPaths.add(testFilePathInStandardDir.toString())
```
petertrr, line 271: You could return the value from here instead of adding into a collection and then get the collection after `collectList`. This way the list is still available for the next operation in chain, but is fully contained in scope of the lambda

petertrr (changes requested):

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
```diff
@@ -230,15 +234,62 @@ class AgentService {
...
                 // fixme: do we still need suitesToArgs, since we have execFlags in save.toml?
-                NewJobResponse(
-                    tests,
-                    suitesToArgs.values.first() +
-                            " " + tests.joinToString(separator = " ") { it.filePath }
-                )
+                constructCliCommand(tests, suitesToArgs).flatMap { cliArgs ->
+                    Mono.fromCallable {
+                        NewJobResponse(tests, cliArgs)
+                    }
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
```diff
@@ -230,15 +234,62 @@ class AgentService {
...
+                }
             } else {
                 log.info("Next test batch for agentId=$agentId is empty, setting it to wait")
-                WaitResponse
+                Mono.fromCallable {
+                    WaitResponse
+                }
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
kgevorkyan: Make sense, changed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/AgentService.kt
kgevorkyan: Thanks, fixed

petertrr (approved):

PR #477, opened 2021-12-09
Preprocessor: Initialize tests in batches

### LINKED ISSUE #474: Support batching from preprocessor
Right now we have a ONE SINGLE POST request from preprocessor to backend.
This causes problems with Juliet tests

petertrr:
### What's done:
* Initialize tests in batches
* Fix: Save language for new test suites

Closes #474

### DIFF
--- a/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
+++ b/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
@@ -54,9 +54,9 @@ import org.springframework.web.reactive.function.client.bodyToMono
 import org.springframework.web.reactive.function.client.toEntity
 import org.springframework.web.server.ResponseStatusException
 import reactor.core.publisher.Flux
-import reactor.core.publisher.Flux.fromIterable
 import reactor.core.publisher.Mono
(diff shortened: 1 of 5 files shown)


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -554,13 +551,13 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
         execution: Execution
     ): Mono<ResponseEntity<HttpStatus>> {
         val testSuiteIds: MutableList<Long> = mutableListOf()
-        return fromIterable(testSuiteDtos).flatMap<List<TestSuite>?> {
+        return Flux.fromIterable(testSuiteDtos).flatMap<List<TestSuite>?> {
```
kgevorkyan, line 554: But it is redundant?

kgevorkyan (approved):

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
petertrr: Yes, but for me it's more clear with explicit class name.

PR #478, opened 2021-12-10
Enable multithreading in save-agent

### LINKED ISSUE #315: Agent: Try to use coroutines native-mt version
Now agent is signle-threaded and heartbeats are not sent while SAVE is executed.
Otherwise, we'll need to make processBuilder's `exec` method properly `suspend`able - i.e. it should yield and poll process status.

petertrr:
### What's done:
* More logging: log thread id in save-agent
* Run save in a separate single-threaded context (should lead to regular heartbeats while save is being executed)
* Add logs of save-cli into stdout of save-agent

Closes #315

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
@@ -8,9 +8,6 @@ import org.cqfn.save.agent.utils.logInfoCustom
 import org.cqfn.save.agent.utils.readFile
 import org.cqfn.save.agent.utils.sendDataToBackend
 import org.cqfn.save.core.logging.describe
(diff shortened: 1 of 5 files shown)


sanyavertolet (approved):
AFAICS everything is fine

save-agent/src/linuxX64Main/kotlin/org/cqfn/save/agent/SaveAgent.kt
```diff
@@ -70,51 +74,53 @@ class SaveAgent(internal val config: AgentConfiguration,
      * @return Unit
      */
     suspend fun start() = coroutineScope {
-        logInfo("Starting agent")
+        logInfoCustom("Starting agent")
```
kgevorkyan, line 77: Do we have some additional information at this moment? Simple message looks little bit senselessly

akuleshov7 (approved):

PR #480, opened 2021-12-13
Redirect unauthorized requests from FE to login page

### LINKED ISSUE #443: Frontend: support 403 responses, redirects to login page

petertrr:
### What's done:
* Gateway: return 401 for unathorized requests instead of redirect to login
* Handle 401 on frontend - if any request fails with 401, redirect to main page
* After authentication, redirect to `/#/projects` instead of main page

Closes #443

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
@@ -81,7 +81,14 @@ suspend fun request(url: String,
         credentials = credentials,
     )
 )
-    .await()
+    .await().also {
+        if (it.status == 401.toShort()) {
+            // if 401 - change current URL to /login
+            // fixme: this should happen if there is no logged in user on FE
+            // otherwise 401 means that the user has insufficient permissions
+            window.location.href = "${window.location.origin}/login"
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
```diff
@@ -81,7 +81,14 @@ suspend fun request(url: String,
         credentials = credentials,
     )
 )
-    .await()
+    .await().also {
+        if (it.status == 401.toShort()) {
+            // if 401 - change current URL to the main page (with login screen)
+            // fixme: this should happen if there is no logged in user on FE
+            // otherwise 401 means that the user has insufficient permissions
```
kgevorkyan, line 88: but insufficient permissions is more appropriate for 403, no?

kgevorkyan (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
petertrr: Yeah, I believe you're right. I meant that 401 might mean something different and not always cause redirect, but it probably always should.

PR #482, opened 2021-12-14
Frontend: Update react-router to v6

petertrr:
### What's done:
* Bump kotlin wrappers to pre.268
* Change API usages of react-router: https://reactrouter.com/docs/en/v6/upgrading/v5


save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
```diff
@@ -79,91 +79,93 @@ class App : RComponent<PropsWithChildren, AppState>() {
...
                                 }
                             }
                         }
 
                         Route {
                             attrs {
-                                path =
-                                        arrayOf("/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath+")
-                                exact = false  // all paths parts under testFilePath should be captured
+                                path = "/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath/*"
```
petertrr, line 156: react-router v6 has dropped support for regexes in patterns, but trailing `*` now indicates that the entire ending of path should be captured. I tested it manually and it works as expected.

akuleshov7:
What is the need of this update?

petertrr:
> What is the need of this update?

kotlin-wrappers have updated react-router and it was blocking us from updates. And since kotlin-wrappers release only for current version of Kotlin and generally change quite fast, it's worth for us to update too

kgevorkyan (approved):
but let's also wait for @akuleshov7 for final decision, I remember, he had some doubts

akuleshov7 (changes requested):
don't merge this week

akuleshov7 (approved):

Issue #331, opened 2021-10-19
Dedicated descriptions for preprocessing errors


petertrr:
E.g., in service logs we can see `Error during preprocessing, will mark execution.id=<N> as failed; error: 
java.lang.IllegalArgumentException: Provided directory [/home/cnb/repositories/<hash>] doesn't contain save.toml configuration file`, but execution is simply marked as `ERROR` and no additional information is displayed on the frontend.

petertrr:
Partially done in https://github.com/saveourtool/save-cloud/commit/1bdd7f47db818e03b843ceabf6cea2a4f32489fd

PR #483, opened 2021-12-14
Support possibility of switching to the concrete branches and commits in git/standard modes

kgevorkyan:
### What's done:
* Support possibility of switching to the  concrete branches and commits in git/standard modes
Capture


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -334,6 +340,16 @@ class DownloadProjectController(private val configProperties: ConfigProperties,
             }
     }
 
+    // Notice: branches should contain explicit `origin/` prefix
+    private fun switchBranch(git: Git, repoUrl: String, branchOrCommit: String) {
+        log.debug("For $repoUrl switching to the $branchOrCommit")
+        try {
+            git.checkout().setName(branchOrCommit).call()
+        } catch (ex: RefNotFoundException) {
+            log.warn("Provided branch/commit $branchOrCommit wasn't found, will use default branch")
```
petertrr, line 349: I think we should crash execution completely in this case. Why spend additional resources on something, that user hasn't requested?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -728,8 +745,10 @@ fun readStandardTestSuitesFile(name: String) =
             .filter { it.isNotBlank() }
             .associate {
                 val splitRow = it.split("\\s".toRegex())
-                require(splitRow.size == 2)
-                splitRow.first() to splitRow[1].split(";")
+                require(splitRow.size == 3) {
+                    "Follow the format for each line: (Gir url) (branch or commit hash) (testRootPath1;testRootPath2;...)"
+                }
+                (splitRow.first() to splitRow[1]) to splitRow[2].split(";")
```
petertrr, line 751: It's time to have a data class for this :)

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
kgevorkyan: If I remember correctly, we discussed with @akuleshov7, that in this case we will use default branch

kgevorkyan: May be in some cases it will be unrequested, but since jgit require `/origin` prefix and since we are not able to check whether the branch name contain typos like, `maste` or `mmaster` etc, probably it will be also useful for such cases

petertrr: Let's leave it as a workaround until we can display preprocessing errors on FE (#331), but it shouldn't be default behavior IMO.

petertrr (approved):
Please merge after #501

PR #498, opened 2021-12-17
Security: Store new logged in users, pass user info downstream

### LINKED ISSUE #495: Store users in the DB
After user logs in via OAuth, we should store username (and auth provider?) into the DB to be able to set permissions and roles for this user.

petertrr:
### What's done:
* When the user logs in, store user info in the database (BE: add UsersController, GW: add StoringServerAuthenticationSuccessHandler)
* More user info: unique name, identity source
* Default username attribute for github
* GW: Convert authorized tokens into username
* BE: Authorize users in backend by username
* DB: Add mapping project - project creator (user id), add list of project admins
* Make POST endpoint for debug-info internal

Closes #495, preparation for #417


save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/User.kt
```diff
@@ -6,10 +6,12 @@ import javax.persistence.Entity
...
  * @property password *in plain text*
  * @property role role of this user
+ * @property source where the user identity is coming from, e.g. "github"
  */
 @Entity
 class User(
     var name: String?,
     var password: String?,
     var role: String?,
+    var source: String,
```
akuleshov7, line 16: It will be a part of primary key for uniqueness?

save-backend/src/main/kotlin/org/cqfn/save/backend/utils/ConvertingAuthenticationManager.kt
```diff
@@ -0,0 +1,28 @@
...
+class ConvertingAuthenticationManager : ReactiveAuthenticationManager {
+    @Autowired
+    private lateinit var userDetailsService: UserDetailsService
+
+    override fun authenticate(authentication: Authentication): Mono<Authentication> = if (authentication is UsernamePasswordAuthenticationToken) {
+        userDetailsService.findByUsername(authentication.name).map {
+            UsernamePasswordAuthenticationToken(it.username, null, it.authorities)
+        }
+    } else {
+        Mono.error { IllegalStateException("Unsupported authentication type ${authentication::class}") }
```
akuleshov7, line 26: why do we need it? What is the scenario when we go here? 
Isn't it not reachable in SAVE?

petertrr: This should not happen , as long as only gateway sends requests. However, to have safe cast to `UsernamePasswordAuthenticationToken` I added explicit branch

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/User.kt
petertrr: Now I'm considering only user name to be unique. Do you mean we should allow users with the same name from different systems, like from Github and Google? I think it makes sense

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
```diff
@@ -34,8 +44,16 @@ class WebSecurityConfig {
...
-        .formLogin()
-        .and().build()
+        .httpBasic()
+        .authenticationManager(authenticationManager)
+        .and().exceptionHandling {
+            it.authenticationEntryPoint(
+                HttpStatusServerEntryPoint(HttpStatus.UNAUTHORIZED)
+            )
+        }
+        .logout().disable()
```
akuleshov7, line 54: why logout disabled?

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
```diff
@@ -4,25 +4,35 @@
...
     ): SecurityWebFilterChain = http.run {
         // `CollectionView` is a public page
+        // all `/internal/**` requests should be sent only from internal network
+        // they are not proxied from gateway
         authorizeExchange()
-            .pathMatchers("/", "/projects/not-deleted")
+            .pathMatchers("/", "/api/projects/not-deleted", "/internal/**")
+            .permitAll()
+            // resources for frontend
+            .pathMatchers("/*.html", "/*.js*", "img/**", "*.gif", "*.svg")
```
akuleshov7, line 35: gif and svg are not in img?

petertrr: They actually are, I'll change it

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
petertrr: Because in backend spring-security is only used to populate context with users, all actual login/logout logic is only on gateway. So logout option is not needed here

db/v-2/tables/project.xml
```diff
@@ -13,4 +13,17 @@
...
 
+    <changeSet id="project-add-user" author="petertrr" context="dev or prod">
+        <addColumn tableName="project">
+            <column name="user_id" type="bigint">
+<!-- Shouldn't be nullable, but mappings should be done at first -->
+                <constraints foreignKeyName="fk_project_user" references="user(id)" nullable="true"/>
+            </column>
+        </addColumn>
+        <addColumn tableName="project">
+            <column name="admin_ids" type="varchar(256)">
```
akuleshov7, line 24: what about superusers? where do you plan to store them?

petertrr: These will be users with `ROLE_ADMIN` in authorities. USers, that are project admins, will have something like `ROLE_PROJECT_ADMIN`

api-gateway/src/main/kotlin/org/cqfn/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+        .findAndRegisterModules()
+        .registerModule(CoreJackson2Module())
+        .registerModule(OAuth2ClientJackson2Module())
+    private val webClient = WebClient.create(configurationProperties.backend.url)
+
+    override fun onAuthenticationSuccess(
+        webFilterExchange: WebFilterExchange,
+        authentication: Authentication
+    ): Mono<Void> {
+        logger.info("Authenticated user with authentication type ${authentication::class}")
```
akuleshov7, line 35: logged it in WebFilter already?

api-gateway/src/main/kotlin/org/cqfn/save/gateway/utils/ConvertAuthorizationHeaderGatewayFilterFactory.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+@Component
+class ConvertAuthorizationHeaderGatewayFilterFactory : AbstractGatewayFilterFactory<Any>() {
+    override fun apply(config: Any?): GatewayFilter = GatewayFilter { exchange: ServerWebExchange, chain: GatewayFilterChain ->
+        exchange.getPrincipal<Principal>()
+            .map { it.userName() }
+            .map { name ->
+                exchange.mutate().request {
+                    it.headers { headers: HttpHeaders ->
+                        headers.set(HttpHeaders.AUTHORIZATION, "Basic ${
```
akuleshov7, line 23: Encoding now, while we don't have https?

petertrr: Basic auth requires credentials to be encoded, so in order for spring-security in backend to understand it, I encode it here

api-gateway/src/main/kotlin/org/cqfn/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
petertrr: Do you mean `UserLoggingFilter`? It's in backend. However, here logging can  be improved still

akuleshov7 (approved):

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/User.kt
petertrr: Added logic to add user identity source to data passed downstream from gateway, as well as uniqueness constraint

PR #501, opened 2021-12-17
Stability fixes and authorization info for the FE

akuleshov7:
### What's done:
- Propagated the information about the logged-in user from Spring Security gateway to a FE
- Suppressed scheduled update of standard suites
- Changed links on the TopBar (to prevent 404)

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
@@ -42,6 +42,7 @@ class UpdateJob(
 @Profile("prod")
 class StandardSuitesUpdateScheduler(
     private val scheduler: Scheduler,
+    configProperties: ConfigProperties,
 ) {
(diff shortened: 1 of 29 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
```diff
@@ -39,7 +39,7 @@ class UpdateJob(
  * A component that is capable of scheduling [UpdateJob]
  */
 @Service
-@Profile("prod")
+@Profile("automatic-updates")
```
petertrr, line 42: I opened #507 to support this more properly, but it can be merged separately.

petertrr (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
akuleshov7: yes, in a separate PR

PR #502, opened 2021-12-17
Pie Chart. Support the logic for aggregation of test results by test suite

### LINKED ISSUE #475: Support the logic for aggregation of test results by test suite
1. We should be able to sort and filter test execution results by a test suite; (#481)
2. We should create a new view with a statistics, where all tests will be aggregated by suites;

Cheshiriks:
What's done:
* added pie chart
Closes #475


petertrr (changes requested):
Are you using the new component anywhere?

akuleshov7 (changes requested):
@Cheshiriks will not have time for this, @sanyavertolet will need to finish it…

akuleshov7:
@Cheshiriks as @sanyavertolet is busy - when you will finish it?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/externals/chart/PieChartBuilder.kt
```diff
@@ -0,0 +1,49 @@
...
+fun RBuilder.pieChart(
+    data: Array<DataPieChart>,
+    animate: Boolean = false,
+    segmentsShift: Int = 0,
+    viewBoxSize: IntArray = intArrayOf(100, 100),
+    radius: Int = 50,
+    onClick: (Event, Int) -> Unit,
+    onMouseOut: (Event, Int) -> Unit,
+    onMouseOver: (Event, Int) -> Unit,
+    handler: RHandler<PieChartProps> = {},
```
petertrr, line 38: I think such long list of parameters is redundant, as corresponding fields can be set inside `handler`.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/externals/chart/PieChartProps.kt
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Source data. Each entry represents a chart segment
+ *
+ * @property title segment name
+ * @property value value of segment
+ * @property color color of segment
+ * @property key custom value to be used as segments element keys
+ */
+data class DataPieChart(
```
petertrr, line 61: Shouldn't this be `external` too? Is this the class from the library?

Cheshiriks: Not really. It just expects an array of values for the pie.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/externals/chart/PieChartBuilder.kt
Cheshiriks: Yes, I agree. Removed all optional parameters.

petertrr (approved):

akuleshov7 (approved):

PR #503, opened 2021-12-17
Fixed matched and missing

Cheshiriks:
What's done:
* fixed matched and missing


save-backend/src/main/kotlin/org/cqfn/save/backend/service/TestExecutionService.kt
```diff
@@ -154,6 +154,8 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
                     it.startTime = testExecDto.startTimeSeconds?.secondsToLocalDateTime()
                     it.endTime = testExecDto.endTimeSeconds?.secondsToLocalDateTime()
                     it.status = testExecDto.status
+                    it.missingWarnings = testExecDto.missingWarnings
+                    it.matchedWarnings = testExecDto.matchedWarnings
```
petertrr, line 158: Please add assert on missed/matched count in `TestExecutionControllerTest`, so that we don't run in this issue again

petertrr (changes requested):

akuleshov7 (approved):

akuleshov7:
I tested it, works fine

petertrr (approved):

akuleshov7 (approved):

PR #507, opened 2021-12-20
Reschedule quartz task on trigger change

petertrr:
### What's done:
* Revert disable quartz scheduler
* Reschedule task on trigger update

This is continuation of https://github.com/diktat-static-analysis/save-cloud/pull/501/commits/b71b58d614d2733e0be77629fe2eafc160f1e127, which makes config more reliable.


save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
```diff
@@ -39,11 +40,12 @@ class UpdateJob(
  * A component that is capable of scheduling [UpdateJob]
  */
 @Service
-@Profile("automatic-updates")
+@Profile("prod")
 class StandardSuitesUpdateScheduler(
     private val scheduler: Scheduler,
     configProperties: ConfigProperties,
 ) {
+    private val logger = LoggerFactory.getLogger(javaClass)
```
kgevorkyan, line 48: `getLogger(StandardSuitesUpdateScheduler::class`)?

kgevorkyan (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/scheduling/StandardSuitesUpdateScheduler.kt
petertrr: That's essentially the same, I thought that using `this.javaClass` instead of explicit `this` class name can be more compact

PR #509, opened 2021-12-20
Mapping oauth provider -> username attribute

### LINKED ISSUE #497: Support mapping auth provider -> user attributes
E.g., default spring-security attribute for name for Github Oauth provider returns id instead of username, which is the field `login`

petertrr:
### What's done:
* WIP

### TODO:
* Just override uaername parameter in application.properties?

Closes #497


petertrr:
Actually, special config is not needed, because the proper attribute can be set in properties even for preset providers:
```properties
spring.security.oauth2.client.provider.github.user-name-attribute=login
```
I'll just add this to default properties in #498

PR #510, opened 2021-12-20
FE: Support logging in from multiple OAuth2 providers

petertrr:
### What's done:
* Added provider list request
* Added icons
* Rename `info` to `sec` in WebSecurityConfig.kt


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/WelcomeView.kt
```diff
@@ -67,6 +82,20 @@ class WelcomeView : AbstractView<WelcomeProps, IndexViewState>(true) {
         state.isValidPassword = true
     }
 
+    override fun componentDidMount() {
```
akuleshov7, line 85: We need more requests from front-end to make the performance of FE better :)

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/WelcomeView.kt
```diff
@@ -110,16 +139,12 @@ class WelcomeView : AbstractView<WelcomeProps, IndexViewState>(true) {
...
-                            classes = "btn btn-link px-3 text-white text-lg text-center"
-                        ) {
-                            attrs["style"] = kotlinext.js.jsObject<CSSProperties> {
-                                fontSize = "3.2rem".unsafeCast<FontSize>()
-                            }
-                            fontAwesomeIcon(icon = faGithub)
-                        }
+                    state.oauthProviders?.map {
+                        oauthLogin(it, when (it.registrationId) {
+                            "github" -> faGithub
```
akuleshov7, line 144: Will you break the design with it? fontSize/type/class

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/WelcomeView.kt
```diff
@@ -213,4 +238,18 @@ class WelcomeView : AbstractView<WelcomeProps, IndexViewState>(true) {
             display = Display.inline
         }
     }
+
+    private fun RBuilder.oauthLogin(provider: OauthProviderInfo, icon: dynamic) {
+        div("col text-center px-1") {
```
akuleshov7, line 243: ah, I see

PR #516, opened 2021-12-23
Add mapping between `User` and `Project`, `User` and `TestExecution`

### LINKED ISSUE #25: Backend: Support authenticated users: table in DB, associate tests execution with user
* [x] Create Users table in DB (generate unique ID) (done in #401)
* [ ] associate tests execution with user who requested it

petertrr:
### What's done:
* ~Split Project into Project.kt and ProjectDto.kt~ turned out to be too hard; should be done later on top of https://github.com/analysis-dev/save-cloud/pull/516/commits/db21126e90c8bce10ae0b599855375b37433aa9b under #527
* Add missing fields to Project
* Pass username of the user, that creates a new project, from `CreationView` and validate it before saving (only known users can save projects). In #417 we should check that only users with write access to a particular project can edit it (call `/update`, where username is not currently validated)
* Map user that requested execution to that execution
* Introduce `Project.stud(id)` for tests
* Enable `ForbiddenMethodCall` for prints
* Minor fixes in docs, minor refactorings
* More test data

Yet another preparation for #417, closes #25. Foreign keys in `project` and `execution` are nullable to avoid conflicts with existing data. Later non-nullability constraints can be added.


save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/Annotations.kt
```diff
@@ -22,7 +22,18 @@ expect annotation class GeneratedValue()
 
 @OptIn(ExperimentalMultiplatform::class)
 @OptionalExpectation
-expect annotation class JoinColumn()
+@Suppress("KDOC_NO_CONSTRUCTOR_PROPERTY", "LongParameterList")
+expect annotation class JoinColumn(
```
kgevorkyan, line 26: some ideas under this changes?

kgevorkyan (approved):

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/entities/Annotations.kt
petertrr: I wanted to use `JoinColumn` in common code, but it turned out to be impossible because of `foreignKey` parameter. IIRC, we had some troubles with it before too, so I decided to keep my changes here so that we don't re-implement them some more time later. Generally, we should move `Project` into `jvmMain` and remove these `OptionalExpectation`s completely.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/ProjectService.kt
```diff
@@ -1,33 +1,50 @@
...
         val exampleMatcher = ExampleMatcher.matchingAll()
             .withMatcher("name", ExampleMatcher.GenericPropertyMatchers.exact())
             .withMatcher("owner", ExampleMatcher.GenericPropertyMatchers.exact())
         val (projectId, projectSaveStatus) = projectRepository.findOne(Example.of(project, exampleMatcher)).map {
             Pair(it.id, ProjectSaveStatus.EXIST)
         }.orElseGet {
+            username?.let {
+                userRepository.findByName(username)
+                    .map {
+                        project.userId = it.id!!
```
petertrr, line 41: This is incorrect; userId from frontend is -1 in this case, because frontend knows only user name, and `Project` instance is not completed.

db/v-2/tables/test_execution.xml
```diff
@@ -23,4 +23,11 @@
         </update>
     </changeSet>
 
+    <changeSet id="test-execution-user" author="petertrr" context="dev or prod">
+        <addColumn tableName="test_execution">
```
petertrr, line 27: Shouldn't it be just `execution`?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -92,8 +94,8 @@ class ProjectController {
      * @return response
      */
     @PostMapping("/updateProject")
-    fun updateProject(@RequestBody project: Project): ResponseEntity<String>? {
-        val (_, projectStatus) = projectService.saveProject(project)
+    fun updateProject(@RequestBody project: Project): ResponseEntity<String> {
+        val (_, projectStatus) = projectService.saveProject(project, null)
```
petertrr, line 98: Should check that the user has write access, should be done in #417

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
```diff
@@ -84,6 +85,16 @@ external interface ProjectSaveViewState : State {
...
 }
 
+/**
+ * State of ProjectView component
+ */
+external interface ProjectSaveViewProps : PropsWithChildren {
+    /**
+     * Currently logged in user or null
+     */
+    var userInfo: UserInfo?
```
petertrr, line 95: Is it needed? Backend can retrieve user info from token

petertrr: Removed

db/v-2/tables/test_execution.xml
petertrr: Changed

kgevorkyan (approved):

PR #522, opened 2021-12-30
Security for methods in `ProjectController`

### LINKED ISSUE #417: Secure backend's endpoints with correct permission checks

petertrr:
### What's done:
* Add `userId` into `Authentication`
* Introduce role hierarchy
* Enable method security
* Add `ProjectPermissionEvaluator `
* `ProjectController`: 
  * split `/projects` into `/projects/all` and `/project` with the former returning every project and being available for admins only, and the latter returning all projects available for the current user
  * return only projects, that are either public or private and owned by the current user
  * in `/getProject` check user's access to the project
* Fix data in test DB
* More test data

Part of #417


save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
```diff
@@ -61,6 +74,26 @@ class WebSecurityConfig(
         .logout().disable()
         .formLogin().disable()
         .build()
+
+    @Bean
+    fun projectPermissionEvaluator() = ProjectPermissionEvaluator()
+
+    fun roleHierarchy(): RoleHierarchy = mapOf(
+        Role.ADMIN to listOf(Role.PROJECT_OWNER, Role.PROJECT_ADMIN, Role.VIEWER),
```
akuleshov7, line 82: let's remove OWNER and have ADMIN and VIEWER only (HOPE we won't need it)

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/domain/Role.kt
```diff
@@ -0,0 +1,32 @@
+package org.cqfn.save.domain
+
+/**
+ * User roles
+ */
+enum class Role {
```
akuleshov7, line 6: may be split this class into two roles? one for SAVE globally and second is for projects 

This is **NOT** required

save-backend/src/main/kotlin/org/cqfn/save/backend/security/ProjectPermissionEvaluator.kt
```diff
@@ -0,0 +1,36 @@
...
+     * @return whether user described by [authentication] can have [permission] on project [project]
+     */
+    fun hasPermission(authentication: Authentication, project: Project, permission: String): Boolean {
+        if (authentication.hasRole(Role.ADMIN)) {
+            return true
+        }
+
+        val userId = (authentication.details as AuthenticationDetails).id
+        return when (permission) {
+            "read" -> project.public || hasWriteAccess(userId, project)
```
akuleshov7, line 25: move to enum

save-backend/src/main/kotlin/org/cqfn/save/backend/security/ProjectPermissionEvaluator.kt
```diff
@@ -0,0 +1,36 @@
...
+            "write" -> hasWriteAccess(userId, project)
+            "delete" -> project.userId == userId
+            else -> false
+        }
+    }
+
+    private fun Authentication.hasRole(role: Role): Boolean = authorities.any { it.authority == role.asSpringSecurityRole() }
+
+    private fun hasWriteAccess(userId: Long?, project: Project) =
```
akuleshov7, line 34: private? may be it should be common and needed for everyone?

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
petertrr: Removed

save-backend/src/main/kotlin/org/cqfn/save/backend/security/ProjectPermissionEvaluator.kt
petertrr: Created enum

save-backend/src/main/kotlin/org/cqfn/save/backend/security/ProjectPermissionEvaluator.kt
petertrr: For now it's only used inside this class, let's update visibility later if required

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/domain/Role.kt
petertrr: Now we have `user_id` and `admin_ids` in `Project`, which is more or less a per-project role. Since there is no support for per-project roles in spring-security, I think we don't need it as a separate entity yet.

PR #523, opened 2022-01-01
Small updates to be able to build save-cloud on MAC

akuleshov7:


.run/save-frontend_run --continuous.run.xml
```diff
@@ -1,6 +1,11 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="save-frontend:run --continuous" type="GradleRunConfiguration" factoryName="Gradle">
     <ExternalSystemSettings>
+      <option name="env">
+        <map>
+          <entry key="NODE_OPTIONS" value="--openssl-legacy-provider" />
```
akuleshov7, line 6: export NODE_OPTIONS=--openssl-legacy-provider

akuleshov7: fuckers have broken community again: https://github.com/webpack/webpack/issues/14532

akuleshov7: Looks like need to enable it somehow in CI/CD build also

petertrr: > Looks like need to enable it somehow in CI/CD build also

It can be added using `env:` key in gradle step in gh workflow

akuleshov7: Not working with env due to a new issue: https://gitanswer.com/vscode-running-code-via-cli-triggers-error-openssl-legacy-provider-is-not-allowed-in-node-options-typescript-1046666817

PR #533, opened 2022-01-12
Hotfix for local deployment

kgevorkyan:


petertrr:
Some links: https://github.com/http-party/node-http-proxy#setup-a-stand-alone-proxy-server-with-proxy-request-header-re-writing

petertrr:
@kgevorkyan turns out, I forgot, that when we throw `BadCredentialsException` in backend, it falls back to usual basic authorization and displays password prompt in the browser. And we excplicitly throw this exception in case tere is no `X-Authoriztion-Source` in the incoming request. Please try to run frontend with my changes, I was able to connect to backend locally.

Also, we should enable `secure` profile in this PR too.

kgevorkyan:
@petertrr Yeah, now it's works, thanks

Think, it could be merged

petertrr (approved):

PR #538, opened 2022-01-13
Set JVM toolchains for liquibase

petertrr:
* Use JDK version from Versions.kt for liquibase tasks
* Use codecov action instead of bash uploader
* Update autoprefixer


petertrr:
@akuleshov7 we can configure it similarly for diktat. I initially dropped this change, because didn't want to apply java plugin in the root project.

akuleshov7 (approved):

PR #546, opened 2022-01-20
Support functionality of git-pull command for projects

kgevorkyan:
### What's done:
* Support functionality of git-pull command for projects
* Improve existing logic of `git checkout` functionality
* Extract `generateDirectory` method into few ones


save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
```diff
@@ -5,25 +5,161 @@
...
+ * @return jGit git entity
+ */
+fun pullProject(
+    gitDto: GitDto,
+    tmpDir: File,
+    userCredentials: CredentialsProvider?,
+    branchOrCommit: String?
+): Git? {
+    log.info("Starting pull project ${gitDto.url} into the $tmpDir")
+    val git = Git.open(tmpDir)
```
petertrr, line 70: Do operations on this `git` need to be wrapped in `use`?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
```diff
@@ -5,25 +5,161 @@
...
+ * @param userCredentials
+ * @param branchOrCommit
+ * @return jGit git entity
+ */
+fun pullProject(
+    gitDto: GitDto,
+    tmpDir: File,
+    userCredentials: CredentialsProvider?,
+    branchOrCommit: String?
+): Git? {
```
petertrr, line 68: Maybe return `Result<Git>` from here and log errors from a single place?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
```diff
@@ -5,25 +5,161 @@
...
+ * @param repoUrl
+ * @param branchOrCommit
+ * @return flag, whether the switching was successful
+ */
+@Suppress("FUNCTION_BOOLEAN_PREFIX")
+fun switchBranch(git: Git, repoUrl: String, branchOrCommit: String?): Boolean {
+    val branchName = if (branchOrCommit.isNullOrBlank()) getDefaultBranchName(repoUrl) else branchOrCommit
+    log.info("Start switch branch from ${git.repository.branch} to the $branchName for $repoUrl")
+    branchName ?: run {
+        log.error("Branch name wasn't provided and couldn't get default branch for repo $repoUrl")
```
petertrr, line 112: Here you know whether `branchOrCommit` was null, so we can make this message less ambiguous (at least include value of `branchOrCommit`)

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
```diff
@@ -5,25 +5,161 @@
...
+        return false
+    }
+
+    if (git.repository.branch == branchName.replace("${Constants.DEFAULT_REMOTE_NAME}/", "")) {
+        log.warn("Requested branch $branchName for git checkout command equals to the current branch, won't provide any actions")
+        return true
+    }
+
+    try {
+        try {
```
petertrr, line 122: Why nested `try`s?

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
kgevorkyan: Basically, if all is ok, this method return `git`, which is transfers to the `pullOrCloneProjectWithSpecificBranch`, which wrapped in `use` in places, where do we call this function

However, when we return `null`, it will be better to close `git`. despite the fact, that I didn't met any problems, when tested it

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
kgevorkyan: Changed

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
kgevorkyan: I separated this message in two ones

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
kgevorkyan: If we will get an exception `RefAlreadyExistsException` while calling `checkout(git, branchName, true)`, we need to call this method with flag `false`, which also can throw an exception `RefNotFoundException`

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/utils/GitUtil.kt
```diff
@@ -88,14 +93,14 @@ fun pullProject(
             .setFastForward(MergeCommand.FastForwardMode.FF)
             .call()
     } catch (ex: RefNotAdvertisedException) {
-        log.error("Provided branch $branchName seems to be an detached commit, pull command won't be performed!")
-        return null
+        git.close()
+        return Result.failure(RefNotAdvertisedException("Provided branch $branchName seems to be an detached commit, pull command won't be performed!"))
```
petertrr, line 97: What is the original error message in `ex`? Would it be better to wrap it into something like `IllegalArgumentException` with custom message and original exception as `cause`?

kgevorkyan: They talking about, that `This Ref may not exist in the remote or may be hidden by permission settings`

I changed to `IllegalArgumentException `, for some reasons, when I used `cause` it prints nothing, so I put `ex` into message too

petertrr: Please put it as `cause`, just for consistency, too.

petertrr (approved):

PR #547, opened 2022-01-20
infra: Use env variables in docker-compose.yaml

petertrr:
* Rename docker-compose.yaml.template to docker-compose.yaml
* Update DockerStackConfiguration.kt
* Add explicit versions for 3rd party docker images
* Add renovate config for docker images

Initially it seemed to be a good idea to have complete control over compose file while generating it from a template. However, with standard docker way of doing it (via environment variables/environment files) we will be able to use standard tooling (syntax highlighting / format validation / auth-updates etc.). Fixing versions of 3rd party images should also improve stability of deployments.

### DIFF
--- a/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
+++ b/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
@@ -25,8 +25,9 @@ const val MYSQL_STARTUP_DELAY_MILLIS = 10_000L
 fun Project.createStackDeployTask(profile: String) {
     tasks.register("generateComposeFile") {
         description = "Set project version in docker-compose file"
-        val templateFile = "$rootDir/docker-compose.yaml.template"
+        val templateFile = "$rootDir/docker-compose.yaml"
         val composeFile = "$buildDir/docker-compose.yaml"
+        val envFile = "$buildDir/.env"
         inputs.file(templateFile)
         inputs.property("project version", version.toString())
         inputs.property("profile", profile)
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -90,15 +100,16 @@ fun Project.createStackDeployTask(profile: String) {
...
+        // this command puts env variables into compose file
+        val composeCmd = "docker-compose -f ${rootProject.buildDir}/docker-compose.yaml --env-file ${rootProject.buildDir}/.env config"
+        val stackCmd = "docker stack deploy --compose-file -" +
             if (useOverride && composeOverride.exists()) {
-                add("--compose-file")
-                add(composeOverride.canonicalPath)
-            }
-        }.toTypedArray()
-        commandLine("docker", "stack", "deploy", *args, "save")
+                " --compose-file ${composeOverride.canonicalPath}"
```
kgevorkyan, line 107: in this branch we will have `docker stack deploy --compose-file - --compose-file ${composeOverride.canonicalPath}`?
is this an valid command?

petertrr: No, the complete command will be
```bash
docker-compose --file buiild/docker-compose.yaml --env-file build/.env config | docker stack deploy --compose-file - --compose-file ${composeOverride.canonicalPath}
```
where `--compose-file -` means that config will be read from stdin. `docker-compose` will substitute variables with values from `.env` (apparently, `docker stack` doesn't support this functionality).

kgevorkyan (approved):

PR #557, opened 2022-01-31
infra: Cache save-cli version for 10 minutes in a file in build directory

### LINKED ISSUE #536: Improve `getSaveCliVersion()` in build logic
Now we check the version of save-cli on every gradle invocation. This is bad practice, also sometimes we hit github API rate limit. We should instead make this method a proper task with a cached output, which will be re-checked from time to time by gradle itself.

petertrr:
* Cache save-cli version for 10 minutes in a file in build directory
* Do not apply default diktat configuration for root project (it was leading to diktat being applied on sarif4k files)
* Fix deprecated options in detekt.yml
* Enable SARIF reports for detekt and upload them to Github

Closes #536

### DIFF
--- a/.github/workflows/detekt.yml
+++ b/.github/workflows/detekt.yml
@@ -24,11 +24,13 @@ jobs:
           arguments: |
             detektAll
       - name: Copy SARIF reports into a single directory
+        if: ${{ always() }}
         run: |
           mkdir -p build/detekt-sarif-reports
           i=0
           find . -name "detekt.sarif" | while read -r f; do echo "$f -> detekt-$i.sarif" && cp $f build/detekt-sarif-reports/detekt-$i.sarif || echo Cannot copy && echo Copied && i=$((i+1)); done
       - name: Upload SARIF to Github
+        if: ${{ always() }}
         uses: github/codeql-action/upload-sarif@v1
         with:
           sarif_file: build/detekt-sarif-reports
--- a/.github/workflows/detekt.yml
+++ b/.github/workflows/detekt.yml
@@ -8,8 +8,6 @@ on:
 jobs:
   detekt_check:
     runs-on: ubuntu-20.04
-    env:
-      GRADLE_OPTS: -Dorg.gradle.daemon=false
 
     steps:
       - uses: actions/checkout@v2.4.0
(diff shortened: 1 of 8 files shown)


.github/workflows/detekt.yml
```diff
@@ -20,12 +18,19 @@ jobs:
...
-          # additional files to calculate key for dependency cache
-          gradle-dependencies-cache-key: |
-            buildSrc/**/Versions.kt
-      # Workaround for bug in gradle-cache-action + detekt
-      - name: Run gradle command
-        run: ./gradlew detektAll
+          arguments: |
+            detektAll
+      - name: Copy SARIF reports into a single directory
+        if: ${{ always() }}
```
kgevorkyan, line 27: Why do we need this `if`?

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DetektConfiguration.kt
```diff
@@ -19,8 +19,12 @@ fun Project.configureDetekt() {
     apply<DetektPlugin>()
     configure<DetektExtension> {
         config = rootProject.files("detekt.yml")
+        basePath = rootDir.canonicalPath
```
kgevorkyan, line 22: Where do we use it?

.github/workflows/detekt.yml
```diff
@@ -20,12 +18,19 @@ jobs:
...
+        run: |
+          mkdir -p build/detekt-sarif-reports
+          i=0
+          find . -name "detekt.sarif" | while read -r f; do echo "$f -> detekt-$i.sarif" && cp $f build/detekt-sarif-reports/detekt-$i.sarif || echo Cannot copy && echo Copied && i=$((i+1)); done
```
kgevorkyan, line 31: May be also print which iteration couldn't be copied, i.e. `i` in `echo Cannot copy`?

save-orchestrator/build.gradle.kts
```diff
@@ -20,15 +22,19 @@ tasks.withType<KotlinCompile> {
 }
 
 // if required, file can be provided manually
-val saveCliVersion = getSaveCliVersion()
-@Suppress("RUN_IN_SCRIPT")
-if (!file("$buildDir/resources/main/save-$saveCliVersion-linuxX64.kexe").exists()) {
```
kgevorkyan, line 25: What about this check now?
We will download anyway, but don't overwrite, if file already exists? Do we really need download file in this case?

petertrr: We used to query save-cli version from github on every build. Now it's saved by `getSaveCliVersion` task into a properties file, which is updated once per 10 minutes. Then, download task won't download it if it already exists, because we can assume that the file on github is the same.

.github/workflows/detekt.yml
petertrr: Because if detekt gradle task fails, we still want to upload reports to display info about those failures. By default, github actions will skip all steps after a failed one.

buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DetektConfiguration.kt
petertrr: It's required for correct display of detekt reports on github: with this parameter detekt will set paths relative to repository root and github will correctly display them, Here's an example: https://github.com/analysis-dev/save-cloud/security/code-scanning/2?query=ref%3Arefs%2Fpull%2F557%2Fmerge+ref%3Arefs%2Fpull%2F557%2Fhead+ref%3Arefs%2Fheads%2Finfra%2Fcache-save-cli-version%23536

kgevorkyan (approved):

.github/workflows/detekt.yml
petertrr: It's logged right after message with file names, so I think it should be enough as is.

PR #559, opened 2022-02-02
Fix: debug info is not available from frontend

kgevorkyan:
### What's done:
* Fix paths matching

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
@@ -170,7 +170,8 @@ class App : RComponent<PropsWithChildren, AppState>() {
 
                         Route {
                             attrs {
-                                path = "/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath/*"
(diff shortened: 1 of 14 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
```diff
@@ -170,7 +170,8 @@ class App : RComponent<PropsWithChildren, AppState>() {
 
                         Route {
                             attrs {
-                                path = "/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/:testFilePath/*"
+                                // Since testFilePath can represent the nested path, we catch it as *
+                                path = "/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/*"
```
kgevorkyan, line 174: Since from `v6` they suppport only `:id-style params and *` ([route-path-patterns](https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-route-path-patterns))

And `path = "/:owner/:name/history/execution/:executionId/details/:testSuiteName/:pluginName/testFilePath"` without `*` didn't worked, I place the whole `testFilePath` under the `*`

petertrr (approved):

Issue #493, opened 2021-12-16
FE: rerun button: handle exceptions in `invokeOnCompletion`


petertrr:
E.g., if an unathorized user requests rerun, then alert is displayed, and only then the page is redirected to login. We should handle exceptions or non-2xx codes inside `invokeOnCompletion` before showing the alert.

Also, it's time to change alert to something prettier.

Issue #389, opened 2021-11-02
Redirection from the creationView page


akuleshov7:
We observe flaky behavior during the creation of the project. In some cases the project is created, but still stays on the creation page. The log in the front-end console has the following:

```
CompletionHandlerException {message_8yp7un$_0: 'Exception in completion handler InvokeOnCompletion@1[job@2] for StandaloneCoroutine{Cancelled}@2', cause_th0jdv$_0: UninitializedPropertyAccessException, name: 'CompletionHandlerException', stack: 'CompletionHandlerException: Exception in completio…otlin-dce-dev/kotlinx-coroutines-core.js:3971:27)'}


handleCoroutineExceptionImpl | @ | kotlinx-coroutines-core.js:31912
-- | -- | --
  | handleCoroutineException | @ | kotlinx-coroutines-core.js:2062
  | AbstractCoroutine.handleOnCompletionException_tcv7n7$ | @ | kotlinx-coroutines-core.js:443
  | JobSupport.notifyCompletion_mgxta4$_0 | @ | kotlinx-coroutines-core.js:3473
  | JobSupport.completeStateFinalization_a4ilmi$_0 | @ | kotlinx-coroutines-core.js:3409
  | JobSupport.finalizeFinishingState_10mr1z$_0 | @ | kotlinx-coroutines-core.js:3317
  | JobSupport.tryMakeCompletingSlowPath_uh1ctj$_0 | @ | kotlinx-coroutines-core.js:4024
  | JobSupport.tryMakeCompleting_w5s53t$_0 | @ | kotlinx-coroutines-core.js:3987
  | JobSupport.makeCompletingOnce_8ea4ql$ | @ | kotlinx-coroutines-core.js:3971
  | AbstractCoroutine.resumeWith_tl1gpc$ | @ | kotlinx-coroutines-core.js:434
  | CoroutineImpl.resumeWith_tl1gpc$ | @ | kotlin.js:38779
  | DispatchedTask.run | @ | kotlinx-coroutines-core.js:28875
  | MessageQueue.process | @ | kotlinx-coroutines-core.js:32252
  | eval | @ | kotlinx-coroutines-core.js:32207
  | Promise.then (async) |   |  
  | WindowMessageQueue.schedule | @ | kotlinx-coroutines-core.js:32212
  | MessageQueue.enqueue_771g0p$ | @ | kotlinx-coroutines-core.js:32239
[...]

PR #561, opened 2022-02-02
Frontend: Replace usages of `GlobalScope` with explicit coroutine scope

### LINKED ISSUE #320: Frontend: Use explicit `CoroutineScope` per Component
Now we are using `GlobalScope` everywhere. We should instead have a scope per component class, which is canceled in `componentWillUnmount` method

petertrr:
* Replace usages of GlobalScope with explicit coroutine scope in `RComponent`s using ComponentWithScope.kt
* Replace usages of GlobalScope with explicit coroutine scope in functional components
* Fix some compiler warnings
* Restructure nested `p` tags to resolve a warning

We also have some issues with `invokeOnCompletion` in coroutines (#389, #493), which should not be affected by this PR and still need to be investigated.

Closes #320

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/Utils.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/Utils.kt
@@ -88,12 +88,3 @@ internal fun RBuilder.multilineTextWithIndices(text: String) {
         }
     }
 }
-
-/**
- * @param cleanupCallback
- */
-internal fun EffectBuilder.cleanup(cleanupCallback: Cleanup) {
-    this.unsafeCast<Array<Cleanup>>().run {
-        set(lastIndex + 1, cleanupCallback)
-    }
-}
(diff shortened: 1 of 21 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/Utils.kt
```diff
@@ -86,3 +88,12 @@ internal fun RBuilder.multilineTextWithIndices(text: String) {
         }
     }
 }
+
+/**
+ * @param cleanupCallback
+ */
+internal fun EffectBuilder.cleanup(cleanupCallback: Cleanup) {
```
petertrr, line 95: `w: /home/runner/work/save-cloud/save-cloud/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/Utils.kt: (95, 28): Extension is shadowed by a member: public final inline fun cleanup(noinline block: Cleanup /* = () -> Unit */): Unit`

kgevorkyan (approved):

akuleshov7:
oh shit, you made it finally. Nice! Thank you so much!

Issue #194, opened 2021-07-19
Improve usage of MockServer in tests


petertrr:
* [x] Reduce code duplication and extract it into a reusable component
* [x] Add additional logging to this component
* [x] Check that all enqueued mock responses are used
* [x] Add a flavor with mocked responses per request path, because requests can have not determined order

petertrr:
```kotlin
fun createMockWebServer(logger: Logger) = MockWebServer().apply {
    dispatcher = LoggingQueueDispatcher(logger)
    (dispatcher as QueueDispatcher).setFailFast(true)
}

internal class LoggingQueueDispatcher(private val logger: Logger) : QueueDispatcher() {
    override fun dispatch(request: RecordedRequest): MockResponse {
        return super.dispatch(request).also {
            logger.info("For request ${request.requestUrl} returning mocked response $it")
        }
    }
}
```

akuleshov7:
we will never do it

PR #564, opened 2022-02-04
Improve usage of MockServer in tests

sanyavertolet:
Improve usage of MockServer in tests

This PR closese #194 

### What's done:
* [x] Created submodule for test utils
* [x] Implemented function that creates `MockWebServer` instance with additional logging
* [x] Implemented `LoggingQueueDispatcher`, which provides better logging and stores `MockResponse`s in different queues - one queue for method path
* [x] Improved logging
* [x] Improved all the tests that use `MockWebServer`
* [x] Implemented uploading SARIF files as artifacts both in `detekt.yml` and `diktat.yml`


save-orchestrator/src/test/kotlin/org/cqfn/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
```diff
@@ -338,13 +341,16 @@ class HeartbeatControllerTest {
             // todo: should be initialized in @BeforeAll, but it gets called after @DynamicPropertySource
             mockServer = MockWebServer()
             // todo: extract request-based dispatcher to save-cloud-common
```
petertrr, line 343: Are you planning to implement this too? There should be a type of dispatcher, that can enqueue responses by path. Now we sometimes face concurrency issues, because order of requests is undetermined, and responses are taken from MockServer sequentially. [This](https://github.com/square/okhttp/tree/master/mockwebserver#dispatcher) is an example of what I'm talkinb about

save-backend/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ tasks.withType<Test> {
 
 dependencies {
     implementation(projects.saveCloudCommon)
+    implementation(projects.testUtils)
```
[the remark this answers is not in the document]
sanyavertolet: Agree, done

save-orchestrator/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ tasks.withType<Test> {
 
 dependencies {
     api(projects.saveCloudCommon)
+    api(projects.testUtils)
```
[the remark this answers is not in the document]
sanyavertolet: As far as I understood we need to use api here because it is a replace of compile keyword (https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration).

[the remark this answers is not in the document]
petertrr: `api` is needed, when there some project consumes this project as a dependency and you want a transitive dependency to be added to that project too. If you only need to use dependency in this project, without marking it for other consumers, `implementation` should be used. Basically, there is no observable difference here, just `implementation` is more common in such cases. Not sure, why we added save-cloud-common as `api`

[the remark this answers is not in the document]
sanyavertolet: Ok, got it. Thanks for proper answer.

akuleshov7:
21 days left after opening of this PR...

save-orchestrator/src/test/kotlin/org/cqfn/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
sanyavertolet: Ok, done it. Now we have a better dispatcher that has some thread-safe queues: on each path you have one queue. Moreover, `checkQueues()` was added in order to check if all queues are empty (have no enqueued responses with non-empty body)

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -0,0 +1,137 @@
...
+ * Queue dispatcher with additional logging
+ *
+ * @param logger
+ */
+class LoggingQueueDispatcher(private val logger: Logger) : Dispatcher() {
+    private val responses: ResponsesMap = ConcurrentHashMap()
+    private val defaultResponses: ConcurrentMap<String, MockResponse> = ConcurrentHashMap()
+    private var failFastResponse: MockResponse = MockResponse().setResponseCode(HttpURLConnection.HTTP_NOT_FOUND)
+
+    private fun getMethodPath(fullPath: String?) = fullPath?.let { Regex("/[\\w]*").find(it)?.value } ?: ""
```
petertrr, line 24: Why only letters and only first part or URL? Is it intentional (e.g. `/api/method/1` -> `api`)?

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    @Suppress("UnsafeCallOnNullableType", "AVOID_NULL_CHECKS")
+    override fun dispatch(request: RecordedRequest): MockResponse {
+        val path = getMethodPath(request.path)
+        val result = if (defaultResponses[path] != null) {
+            logger.info("Default response [${defaultResponses[path]}] exists for path [$path].")
```
petertrr, line 30:
```suggestion
            logger.debug("Default response [${defaultResponses[path]}] exists for path [$path].")
```
Also, I think it would make sense to create a separate logger for this class instead of providing one and then tweak its level via `resources/test/application.properties`

save-preprocessor/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ configureSpringBoot()
 
 dependencies {
     implementation(projects.saveCloudCommon)
+    implementation(projects.testUtils)
```
petertrr, line 13:
```suggestion
    testImplementation(projects.testUtils)
```

sanyavertolet: Thanks

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
sanyavertolet: As far as I saw, we do not have anything like `/api/method/1 -> api` yet. But it would be better to use `Regex("^/[^?]*[^/]")`. Thanks

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
sanyavertolet: Ok, done

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -2,27 +2,107 @@
...
+/**
+ * @param path path to store enqueued response
+ * @param response response to enqueue
+ * @throws ClassCastException
+ */
+fun MockWebServer.enqueue(path: String, response: MockResponse) {
+    if (dispatcher is LoggingQueueDispatcher) {
+        (dispatcher as LoggingQueueDispatcher).enqueueResponse(path, response.clone())
+    } else {
+        throw ClassCastException("dispatcher type should be LoggingQueueDispatcher")
```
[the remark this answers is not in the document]
sanyavertolet: Changed it

petertrr (approved):

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -19,34 +19,45 @@ class LoggingQueueDispatcher : Dispatcher() {
...
 
     @Suppress("UnsafeCallOnNullableType", "AVOID_NULL_CHECKS")
     override fun dispatch(request: RecordedRequest): MockResponse {
-        val path = getMethodPath(request.path)
-        val result = if (defaultResponses[path] != null) {
-            logger.debug("Default response [${defaultResponses[path]}] exists for path [$path].")
-            defaultResponses[path]!!
-        } else if (responses[path]?.peek() == null) {
-            logger.info("No response is present in queue with path [$path].")
+        val regexKeyForDefaultResponses = getProperRegexKey(request.path, defaultResponses.keys.toSet())
```
petertrr, line 36:
```suggestion
        val regexKeyForDefaultResponses = getProperRegexKey(request.path, defaultResponses.keySet())
```

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -19,34 +19,45 @@ class LoggingQueueDispatcher : Dispatcher() {
     private val defaultResponses: ConcurrentMap<String, MockResponse> = ConcurrentHashMap()
     private var failFastResponse: MockResponse = MockResponse().setResponseCode(HttpURLConnection.HTTP_NOT_FOUND)
 
-    private fun getMethodPath(fullPath: String?) = fullPath?.let { Regex("^/[^?]*[^/]").find(it)?.value } ?: ""
+    private fun getProperRegexKey(path: String?, setOfRegexes: Set<String>) = path?.let {
+        setOfRegexes
+            .filter { regex -> Regex(regex).containsMatchIn(it) }
```
petertrr, line 24: You can store `Regex`s as keys instead of creating them on each invocation

petertrr (approved):

Issue #535, opened 2022-01-13
Agent: Investigate and support parallelism


petertrr:
Now, despite using mutithreaded context for coroutines, we don't see actual parallelism, which is required for proper #208 . Note: Kotlin/Native is moving to a new memory model, which is already available/supported in coroutines 1.6.0: https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md#update-the-libraries, so we can add parallelism already keeping in mind the new MM

PR #572, opened 2022-02-14
(WIP) Agent: Support new memory model

petertrr:
This PR contains some experiments for #535 based on https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md#update-the-libraries. It adds required compiler flags and adapts some API changes from coroutines, including an ability to create thread pool with multiple threads in Native.

This PR exists only so that this information isn't lost of forgotten, it's not really required right now, since #535 is already fixed with existing MM.


akuleshov7:
@kgevorkyan @petertrr what is with this experiment? It looks good, any problems with that?

petertrr:
> @kgevorkyan @petertrr what is with this experiment? It looks good, any problems with that?

The only problem is that ktor hasn't published any new previews of 1.6.x branch with new MM, only 2.x, which is still in beta. I believe we should finish this PR after ktor 2 is released, and probably Kotlin 1.6.20 will be available by that time too.

PR #573, opened 2022-02-14
backend: Add hibernate-micrometer

petertrr:
* Add autoconfigured hibernate-metrics
* Allow access to `/actuator` in backend, because it comes only from internal network


kgevorkyan (approved):
Let's may be suppress extra debug info from it if any

PR #574, opened 2022-02-15
Backend: Secure API endpoints with permission checks

### LINKED ISSUE #417: Secure backend's endpoints with correct permission checks

petertrr:
### What's done:
* Group all public endpoints in a single list
* Apply permission checks to all `/api/**` endpoints:
  * `WRITE` permission on a project to submit execution requests for this project and rerun executions
  * `READ` permission to access executions and test executions for a project
  * `DELETE` permission on a project to delete executions for this project
* Move `/internal` controllers into a separate package

`DownloadFilesController`: now it's hard to apply any permission checks, because files are not separated by projects and/or organizations.

Closes #417

On a side note - how can we ensure that all methods are secure? They either use `@PreAuthorize` etc, or use `projectPermissionEvaluator` or something similar, or are listed in `pathMatchers` in security config.


save-backend/src/main/kotlin/org/cqfn/save/backend/security/ProjectPermissionEvaluator.kt
```diff
@@ -27,24 +35,55 @@ class ProjectPermissionEvaluator {
...
+        if (authentication.hasRole(Role.SUPER_ADMIN)) {
             return true
         }
 
         val userId = (authentication.details as AuthenticationDetails).id
         return when (permission) {
             Permission.READ -> project.public || hasWriteAccess(userId, project)
             Permission.WRITE -> hasWriteAccess(userId, project)
+            // fixme: check role to ensure its `OWNER`
             Permission.DELETE -> project.userId == userId
```
petertrr, line 47: I mean, we should somehow check data for consistency: role `OWNER` is specifically for project owner, but we also have `project.user_id` for project owner. We should resolve this redundancy by removing one of the entities or by adding checks.

save-backend/src/test/kotlin/org/cqfn/save/backend/DownloadFilesTest.kt
```diff
@@ -46,6 +48,7 @@ import kotlin.io.path.createTempFile
 import kotlin.io.path.name
 import kotlin.io.path.writeLines
 
+@ActiveProfiles("test")
```
petertrr, line 51: No security checks here, because there is no security logic in files API now

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -93,98 +112,126 @@ class ExecutionController(private val executionService: ExecutionService,
...
-            executionService.deleteExecutionByIds(executionIds)
-        } catch (e: IllegalArgumentException) {
-            log.warn("Could not find the following executions: $executionIds or related objects", e)
-            throw ResponseStatusException(HttpStatus.NOT_FOUND, "Failed to delete executions for the following reason: ${e.message}")
+    fun deleteExecutionsByExecutionIds(@RequestParam executionIds: List<Long>, authentication: Authentication): Mono<ResponseEntity<*>> = Flux.fromIterable(executionIds)
+        .map { it to executionService.findExecution(it) }
+        .filterAndInvoke({ (id, _) -> log.warn("Cannot delete execution id=$id because it's missing in the DB") }) { (_, execution) ->
+            execution.isPresent
+        }
+        .map { (_, execution) -> execution.get() }
```
kgevorkyan, line 204: Quite a long chain to get projects, and may be it will be useful in the future to reuse this logic, maybe we can extract this logic into separate method?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -63,14 +64,21 @@ class CloneRepositoryController(
...
         authentication: Authentication,
-    ): Mono<StringResponse> = sendToPreprocessor(
-        executionRequest,
-        ExecutionType.GIT,
-        authentication.username(),
-        files
-    ) { newExecutionId ->
-        part("executionRequest", executionRequest.copy(executionId = newExecutionId))
+    ): Mono<StringResponse> = with(executionRequest.project) {
+        // Project cannot be taken from executionRequest directly for permission evaluation:
```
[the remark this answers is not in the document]
petertrr: > or maybe better to use `messageIfNotFound` argument

No, this string is sent in a response body, I've added it just for consistency with original implementation. Frontend (or other user agent like curl) shouldn't know these details, I think. And we'll see exception logged in service logs.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
petertrr: Yes, it's long, but for now it's the only method, where we accept a list of execution IDs. So at first we should filter out non-existent, then map the rest to projects. Now we use this method only to delete a single execution, and even for the future it would make sense to simplify logic by failing if some IDs don't exist or belong to different projects.

save-backend/src/test/resources/application.properties
```diff
@@ -12,9 +12,12 @@ spring.sql.init.mode=always
...
+spring.jpa.properties.hibernate.use_sql_comments=true
 spring.jpa.properties.hibernate.validator.apply_to_ddl=false
 spring.jpa.hibernate.use-new-id-generator-mappings=false
 # logging
 logging.level.org.cqfn.save.backend.service.TestService=DEBUG
 logging.level.org.cqfn.save.backend.service.TestExecutionService=DEBUG
-logging.level.org.cqfn.save.backend.controllers.TestController=DEBUG
\ No newline at end of file
+logging.level.org.cqfn.save.backend.controllers.internal.TestController=DEBUG
+logging.level.org.hibernate.SQL=DEBUG
```
kgevorkyan, line 22: What kind of information we will get by these `debug` and `trace`, I mean will it clutter up our output a lot?

kgevorkyan (approved):
generally LGTM, but also please test it manually, since of a lot of changes

save-backend/src/test/resources/application.properties
petertrr: Yes, a lot, but I it's only for tests and I thought that sometimes currently logged info (SQL requests with `?` isntead of parameter values) is not enough for debugging

kgevorkyan: ok, didn't pay attention, that it is for tests

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -89,7 +91,9 @@ class ProjectController(private val projectService: ProjectService,
             val organization = organizationService.getOrganizationById(organizationId)
             projectService.findByNameAndOrganization(name, organization)
         }
-        return getPermissionProject(project, authentication)
+        return with(projectPermissionEvaluator) {
+            project.filterByPermission(authentication, Permission.WRITE, HttpStatus.FORBIDDEN)
```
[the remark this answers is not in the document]
petertrr: Also, `ProjectPermissionEvaluator` should use default role (`VIEWER`) if user has no mapping in `lnk_user_project`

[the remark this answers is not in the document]
petertrr: >  I'll recheck logic here

So, there is a fixme comment and we'll show a different view for users, that have read-only access to a project.

save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/Execution.kt
```diff
@@ -36,7 +37,7 @@ import javax.persistence.ManyToOne
 @Entity
 class Execution(
 
-    @ManyToOne
+    @ManyToOne(fetch = FetchType.EAGER)
```
petertrr, line 40: Just explicitly specifying default value, maybe we'll change it to LAZY later

PR #577, opened 2022-02-17
Remove `diktatCheckAll` and call `diktatCheck` directly

petertrr:


gradle.properties
```diff
@@ -5,7 +5,8 @@ org.gradle.jvmargs=-Xmx1g -XX:MaxMetaspaceSize=512m
 org.gradle.parallel=true
 org.gradle.vfs.watch=true
 ## https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand
-org.gradle.configureondemand=true
+## there will be a replacement `org.gradle.unsafe.isolated-project`: https://github.com/gradle/gradle/pull/16930
+#org.gradle.configureondemand=true
```
petertrr, line 9: I've read that this flag can speed up e.g. running individual tests in a repository with many subprojects, but apparently because of it I wasn't able to use IDEA debugger (`JavaToolchainService` wasn't available for diktat task configuration)

akuleshov7 (approved):

PR #579, opened 2022-02-18
Frontend: Fix recreation of functional components

### LINKED ISSUE #426: Frontend: Focus on the test-resources information fields is lost after editing (Git mode page)
If we will copy-paste the some info to that fields all will be okay,

but if we will try to edit fields and enter the values **manually** the focus of cursor will be lost, and it will be necessary to click to the fields one more time after each new symbol, which is quite inconvenient 

bug

petertrr:
All functional components shouldn't be called directly from `Component.render` method. Instead, they should be stored in class properties. Explanation of this behavior can be found [here](https://dev.to/igor_bykov/react-calling-functional-components-as-functions-1d3l)
* The following functional components are covered: `executionStatistics`, `testResourcesSelection`, `cardComponent`, `topBar`, `scrollToTopButton`, `executionTestsNotFound`, `testExecutionFiltersRow`, `fileUploader`, `testExecutionDetailsView`, `withRouter`
* function components inside other function components are OK, we need to make sure that top-level function components aren't recreated; `testStatusComponent` shouldn't be extracted, because it's called for a new table row each time
* Make Boolean properties in external interfaces nullable
* Changed `executionTestsNotFound` to receive tests count via props instead of function parameter
* Replace String with dynamic for fa icons in HistoryView.kt

Closes #426

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
@@ -32,6 +32,12 @@ import kotlinx.html.id
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
 
+private val scrollToTopButton = scrollToTopButton()
+
+private val topBar = topBar()
+
+private val testExecutionDetailsView = testExecutionDetailsView()
+
 /**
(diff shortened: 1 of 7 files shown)


kgevorkyan (approved):
lgtm

PR #587, opened 2022-02-21
Fix of serialization corresponding to the date time in standard mode

kgevorkyan:
### What's done:
* Use kotlinx.datetime.LocalDateTime.serializer() inside LocalDateTimeSerializer
* Write dates as strings in Jackson
* Fix JSON serialization in multipart request from preprocessor

### DIFF
--- a/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/config/LocalDateTimeConfig.kt
+++ b/save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/config/LocalDateTimeConfig.kt
@@ -1,3 +1,8 @@
+/**
+ * Additional configuration to establish serialization of LocalDateTime consistent with other services.
+ * There is a copy of this configuration in save-orchestrator as well, and they should be consistent.
+ */
+
 @file:Suppress(
     "MISSING_KDOC_TOP_LEVEL",
     "MISSING_KDOC_CLASS_ELEMENTS",


petertrr:
We should try to use `kotlinx.datetime.LocalDateTime.serializer()` instead of our custom one and make backend encode dates in strings, as it used to do. I'll try to do it in this branch.

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -81,16 +84,22 @@ typealias Status = Mono<ResponseEntity<HttpStatus>>
...
         it.defaultCodecs().kotlinSerializationJsonEncoder(kotlinSerializationJsonEncoder)
         it.defaultCodecs().kotlinSerializationJsonDecoder(kotlinSerializationJsonDecoder)
-    }
-        .build()
-    private val webClientOrchestrator = WebClient.create(configProperties.orchestrator)
+    }.build()
+    private val webClientOrchestrator = WebClient.builder().baseUrl(configProperties.orchestrator).codecs {
+        it.defaultCodecs().kotlinSerializationJsonEncoder(kotlinSerializationJsonEncoder)
+        it.defaultCodecs().kotlinSerializationJsonDecoder(kotlinSerializationJsonDecoder)
+        it.defaultCodecs().multipartCodecs().encoder(Jackson2JsonEncoder(objectMapper))
```
petertrr, line 99: We have to use Jackson to serialize `Execution`, because it's not `@Serializable` and spring-boot's `KotlinSerializationJsonEncoder` claims that it can't serialize it (`canEncode` returns `false`)

save-preprocessor/src/main/kotlin/org/cqfn/save/preprocessor/config/LocalDateTimeConfig.kt
```diff
@@ -0,0 +1,66 @@
+@file:Suppress(
```
kgevorkyan, line 1: May be add some comments, that this file should be in consistent with the same from preprocessor?

petertrr: Added

PR #590, opened 2022-02-24
Fixes related to security and permissions

### LINKED ISSUE #589: Backend: `/api/projects/update` is not working
Return type of function should be a reactive Publisher in order to user `@PreAuthorize` annotation

petertrr:
* Fix `/projects/update`
* Workaround: disable editing of project name, because we rely on it to update the project
* Fix roles hierarchy: change `ADMIN` to `SUPER_ADMIN` in ProjectPermissionEvaluator, update `RolesHierarchy` in `WebSecurityConfig`
* Enable `secure` profile of backend by default
* Update permissions in test data
* Include all detekt reports in sarif-upload (they produce separate files for different tasks in multiplatform mode)

Some things cherry-picked from #574, as it became too big. This PR covers only bugs, fixed in #574, without new features. Closes #589


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
```diff
@@ -666,12 +666,14 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
 
     private fun turnEditMode(isOff: Boolean) {
         projectInformation.keys.forEach {
-            val informationKey = (document.getElementById(it) as HTMLInputElement).apply {
-                disabled = isOff
+            val informationKey = document.getElementById(it) as HTMLInputElement
+            if (it != "Tested tool name: ") {
+                // temporary workaround for https://github.com/analysis-dev/save-cloud/issues/589#issuecomment-1049674021
```
kgevorkyan, line 671: By current logic, it seems, that we really just create new record in DB, instead of changing
Strange, It seemed to me that I was checking manually, when created this logic, that we update project info, not create an new one 😐

kgevorkyan (approved):
lgtm, changes already discussed and reviewed in #574

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
petertrr: I've opened #591 for this

PR #595, opened 2022-02-28
Minor fix for organization

Cheshiriks:
What's done:
* rename owner to organization
* added load organization in organizationView
* deleted owner in project table

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
@@ -49,24 +49,30 @@ external interface OrganizationViewState : State {
      * Image to owner avatar
      */
     var image: ImageInfo?
+
+    /**
+     * Organization
+     */
+    var organization: Organization?
 }
 
 /**
  * A Component for owner view
  */
 class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(false) {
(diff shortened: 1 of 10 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -54,6 +55,7 @@ external interface OrganizationViewState : State {
  * A Component for owner view
  */
 class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(false) {
+    private var organization = Organization("", null, null, null)
```
petertrr, line 58: I believe it should be in state, otherwise data will not update after it's fetched in `componentDidMount`

Cheshiriks: The organization is used only for the title name, so it is not necessary to add it to the state

petertrr: Yes, but request is sent asynchronously, so there is no guarantee that organization.name will be set with the correct value when title is rendered

Cheshiriks: Ok. Changed

petertrr (approved):

PR #596, opened 2022-03-01
Create `projectInfo` component, fix project info data updates

### LINKED ISSUE #588: Frontend: Project info is absent after page reloading
At the initial state all fields are `N/A`, however DB contain all necessary information

After clicking to the `edit` button and then to `cancel`, information appears:

petertrr:
* Create `projectInfo` component
* Store project in state of ProjectView
* Store draft project in state of `projectInfo`
* Only send request for update if project has actually changed

Closes #588

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ProjectInfo.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ProjectInfo.kt
@@ -1,22 +1,27 @@
+/**
+ * Function component for project info and edit support
+ */
+
+@file:Suppress("FILE_NAME_MATCH_CLASS")
+
 package org.cqfn.save.frontend.components.basic
 
 import org.cqfn.save.entities.Project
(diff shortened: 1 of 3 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ProjectInfo.kt
```diff
@@ -0,0 +1,135 @@
...
+
+import org.cqfn.save.entities.Project
+import org.cqfn.save.frontend.externals.fontawesome.faCheck
+import org.cqfn.save.frontend.externals.fontawesome.faTimesCircle
+import org.cqfn.save.frontend.externals.fontawesome.fontAwesomeIcon
+
+import react.FC
+import react.Props
+import react.dom.html.InputType
+import react.dom.html.ReactHTML.button
```
petertrr, line 17: `react.FC` rather than `react.fc` and DOM elements from `react.dom.html.ReactHTML` is a new kotlin-react API, which uses builders without `attrs`

akuleshov7 (approved):

PR #597, opened 2022-03-01
Mark agents as crashed after a number of skipped heartbeats

kgevorkyan:
### What's done:
* Create background listener, which will mark agents as crashed after the configurable period, during which there were no heartbeats
* Add tests


save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestExecutionRepository.kt
```diff
@@ -101,6 +101,14 @@ interface TestExecutionRepository : BaseEntityRepository<TestExecution>, JpaSpec
      */
     fun findByExecutionIdAndTestId(executionId: Long, testId: Long): List<TestExecution>
 
+    /**
+     * @param executionId
+     * @param agentId
+     * @return list of TestExecution's
+     */
+    @Suppress("TYPE_ALIAS")
+    fun findByExecutionIdAndAgentId(executionId: Long, agentId: Long): Optional<List<TestExecution>>
```
petertrr, line 110:
```suggestion
    fun findByExecutionIdAndAgentId(executionId: Long, agentId: Long): List<TestExecution>
```
I believe, spring-data methods that return collections will never return null or Optional.empty, but rather an empty collection

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -82,18 +101,77 @@ class HeartbeatController(private val agentService: AgentService,
...
         }
 
+    /**
+     * Collect information about latest heartbeats from agents, in aim to determine crashed one later
+     *
+     * @param agentId
+     * @param state
+     */
+    fun updateAgentHeartbeatTimeStamps(agentId: String, state: AgentState) {
+        agentsLatestHeartBeatsMap[agentId] = state.name to LocalDateTime.now()
```
petertrr, line 115: I thought, we store timestamp on heartbeat sending in the agent... We should consider adding it in the future

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -71,7 +90,7 @@ class HeartbeatController(private val agentService: AgentService,
                         }
                     }
                     AgentState.BUSY -> Mono.just(ContinueResponse)
-                    AgentState.BACKEND_FAILURE, AgentState.BACKEND_UNREACHABLE, AgentState.CLI_FAILED, AgentState.STOPPED_BY_ORCH -> Mono.just(WaitResponse)
+                    AgentState.BACKEND_FAILURE, AgentState.BACKEND_UNREACHABLE, AgentState.CLI_FAILED, AgentState.STOPPED_BY_ORCH, AgentState.CRASHED -> Mono.just(WaitResponse)
```
petertrr, line 93: Can agent really send a `CRASHED` state? It should be a state that means that agent is offline. We shouldn't treat it silently, at least log a warning or an error.

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -126,3 +214,18 @@ class HeartbeatController(private val agentService: AgentService,
...
+ * Background inspector, which detect crashed agents
+ *
+ * @property heartbeatController
+ */
+class HeartBeatInspector(private val heartbeatController: HeartbeatController, private val heartBeatInspectorInterval: Long) : Thread() {
+    override fun run() {
+        while (!interrupted()) {
+            heartbeatController.determineCrashedAgents()
+            heartbeatController.processCrashedAgents()
+            heartbeatController.saveFromInterruptionSleep(heartBeatInspectorInterval)
```
petertrr, line 228: You can use spring mechanisms for scheduled task execution with `@Scheduled` annotation instead of manual time management

save-orchestrator/src/test/kotlin/org/cqfn/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
```diff
@@ -216,6 +218,71 @@ class HeartbeatControllerTest {
...
+                ),
+                mapOf(1L to "")
+            ),
+            testSuite = TestSuite(TestSuiteType.PROJECT, "", null, null, LocalDateTime.now(), ".", ".").apply {
+                id = 0
+            },
+            mockAgentStatuses = false,
+        ) {
+            // FixMe: we actually need to check the size of crashed agents list somehow
+            verify(dockerService, atLeast(2)).stopAgents(any())
```
petertrr, line 282: There should be a matcher for exactly 2 in mockito instead of `atLeast`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -131,6 +131,14 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
         testExecutionService.assignAgentByTest(agentContainerId, testDtos)
     }
 
+    /**
+     * @param crashedAgents the list of agents, which weren't sent heartbeats for a some time and are considered as crashed
+     */
+    @PostMapping(value = ["/internal/testExecution/markTestExecutionsOfCrashedAgentsAsFailed"])
```
petertrr, line 137: Maybe make this method not so specific, like `/testExecution/setStatusByAgentIds?status=FAILED`?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
kgevorkyan: Agree, changed

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TestExecutionRepository.kt
kgevorkyan: Agree, removed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
kgevorkyan: But in this case, value should be compiler time known, so couldn't be configured from `.properties`, as I see

We use different values in production code and in tests, also it might be not hard-coded in future too

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
kgevorkyan: Changed

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
petertrr: You can use SpEL to access properties in annotation parameters, like https://www.baeldung.com/spring-scheduled-tasks#parameterizing-the-schedule

save-orchestrator/src/test/kotlin/org/cqfn/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
kgevorkyan: Actually, the situation is quite more complicated
Since we are not actually start containers, they couldn't be stopped by `stopAgent`, and thus the `crashedAgentsList` won't be cleared, so `stopAgent` method will try to stop them time after time, so it's not really good check at all

It would be good to have access to the `crashedAgentsList` from tests and just check the size of it, but for now I don't know, that it could be possible

I will think more about more appropriate assertsions in these tests

kgevorkyan: Changed to assertion, which check `crashedAgentsList` size, previously, probably because of cache, it didn't worked for me

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
kgevorkyan: Yeah, thanks, added

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/controller/HeartbeatController.kt
kgevorkyan: Open https://github.com/analysis-dev/save-cloud/issues/609

petertrr (approved):

Issue #562, opened 2022-02-03
CreateOrganization View 


akuleshov7:
We need to create a logic for the creation of new organizations. What is needed from you:
![image](https://user-images.githubusercontent.com/58667063/152341924-b315a7c7-563d-40cb-9c6b-a6df07fd160a.png)

1) Create new Organization button; 
2) Create new View for the creation of the Organization;
3) Backend endpoint that will create organization ();  
4) Database update with a new organization;

Later, after Organization view will be ready - we will also we will also need a create new project from the organization view

Cheshiriks:
You also need to create those organizations that we have are the owners of projects on the prod. And update the prod database for mapping the project to the organization.

sanyavertolet:
When we merge #610, I will need to add uploading organization avatar when creating organization.

akuleshov7:
> When we merge #610, I will need to add uploading organization avatar when creating organization.

not critical at all, backlog. Thank you for the initial contribution

PR #610, opened 2022-03-11
CreateOrganization View

sanyavertolet:
This issue covers almost all the #562 

### What's done:
* Added button to create new Organization
* Added new View for the creation of the organization
* Created backend endpoint that will create organization

(#562)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -28,4 +31,32 @@ internal class OrganizationController(private val organizationService: Organizat
...
+        val ownerId = (authentication.details as AuthenticationDetails).id
+        val (organizationId, organizationStatus) = organizationService.saveOrganization(
+            newOrganization.apply {
+                this.ownerId = ownerId
+                this.dateCreated = LocalDateTime.now()
+            }
+        )
+
+        if (organizationStatus == OrganizationSaveStatus.EXIST) {
+            logger.warn("Organization with id = $organizationId already exists")
```
petertrr, line 51: Will any data get overwritten in this case?

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/domain/OrganizationSaveStatus.kt
```diff
@@ -0,0 +1,22 @@
+package org.cqfn.save.domain
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Enum of organization save status
+ * @property message
+ */
+@Serializable
+@Suppress("CUSTOM_GETTERS_SETTERS")
```
petertrr, line 10: Why is this needed?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/OrganizationService.kt
```diff
@@ -10,6 +14,27 @@ import org.springframework.stereotype.Service
...
+     * Store [organization] in the database
+     *
+     * @param organization an [Organization] to store
+     * @return organization's id, should never return null
+     */
+    @Suppress("UnsafeCallOnNullableType")
+    fun saveOrganization(organization: Organization): Pair<Long, OrganizationSaveStatus> {
+        val exampleMatcher = ExampleMatcher.matchingAll()
+            .withMatcher("name", ExampleMatcher.GenericPropertyMatchers.exact())
+            .withMatcher("ownerId", ExampleMatcher.GenericPropertyMatchers.exact())
```
petertrr, line 27: I recognize copy-paste from `ProjectService`, but doesn't `matchingAll` already mean all fields match `exact`? I don't remember, why we use it like this there, but I think it's redundant.

sanyavertolet: Oh yes, there was a copy-paste from `ProjectService`. Now it is something like that:
```
val exampleMatcher = ExampleMatcher.matching()
            .withMatcher("name", ExampleMatcher.GenericPropertyMatchers.exact())
```
But I cannot see what is the difference between `matching` and `matchingAll` cause according to [`docs.spring.io`](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/ExampleMatcher.html) they do exactly the same.

sanyavertolet: Actually this is what source code looks like:
```
static ExampleMatcher matching() {
	return matchingAll();
}
```

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/domain/OrganizationSaveStatus.kt
sanyavertolet: Copy-paste problem, deleted it.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: No, it will not because if there is an existing organization that conflicts the new one, an existing one will be returned from `saveOrganization`.

petertrr: Right, we are checking if it exists inside the method. Maybe, then rename it to something like `getOrSaveOrganization`?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/OrganizationService.kt
petertrr: I mean, can we simply drop two lines with specific matchers and only leave `matchingAll`? I believe, result will be the same

sanyavertolet: The idea is to match only name because either we get one organization per user (`matchingAny`) or many organizations with same name (`matching`/`matchingAll`)

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: Maybe but we need to rename same method for `ProjectController`. I guess it would be nice to keep same format for same things in different classes.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/OrganizationService.kt
petertrr: Then we can simply use `findByName`, because example matcher is better suited for different case - when you would like to match against multiple fields

sanyavertolet: Oh ok, will fix it soon

sanyavertolet: Done

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -23,14 +23,14 @@ internal class OrganizationController(private val organizationService: Organizat
...
-            organizationService.findByName(name)
+            organizationService.findByName(name) ?: throw NoSuchElementException("Organization with name [$name] was not found.")
 
     /**
      * @param owner owner name
      * @return a image
      */
     @GetMapping("/avatar")
-    fun avatar(@RequestParam owner: String): ImageInfo? = organizationService.findByName(owner).avatar.let { ImageInfo(it) }
+    fun avatar(@RequestParam owner: String): ImageInfo? = organizationService.findByName(owner)?.avatar.let { ImageInfo(it) }
```
petertrr, line 33: Should it be `?.let`? Otherwise function never returns null, only null wrapped into `ImageInfo`. If that's intended, we can change returned type

save-backend/src/main/kotlin/org/cqfn/save/backend/service/OrganizationService.kt
```diff
@@ -51,9 +44,9 @@ class OrganizationService(private val organizationRepository: OrganizationReposi
      * @param relativePath
      */
     fun saveAvatar(name: String, relativePath: String) {
-        val organization = organizationRepository.findByName(name).apply {
+        val organization = organizationRepository.findByName(name)?.apply {
```
petertrr, line 47: Please handle missing organization here (maybe, with an exception). Now you silently ignore missing organization and don't indicate that no avatar has actually been saved.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -20,12 +23,40 @@ internal class OrganizationController(private val organizationService: Organizat
...
+        val ownerId = (authentication.details as AuthenticationDetails).id
+        val (organizationId, organizationStatus) = organizationService.getOrSaveOrganization(
+            newOrganization.apply {
+                this.ownerId = ownerId
+                this.dateCreated = LocalDateTime.now()
+            }
+        )
+
+        if (organizationStatus == OrganizationSaveStatus.EXIST) {
+            logger.warn("Organization with id = $organizationId already exists")
```
petertrr, line 51:
```suggestion
            logger.info("Attempt to save an organization with id = $organizationI, but it already exists")
```
I think it shouldn't be a warning, because it is a valid and properly handled case. Otherwise it'll just pollute logs.

sanyavertolet: Ok, agree

save-backend/src/main/kotlin/org/cqfn/save/backend/service/OrganizationService.kt
sanyavertolet: Yes, true. Now throwing exception.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: Ok, changed return type.

petertrr (approved):

PR #615, opened 2022-03-14
Set default role for a new user

### LINKED ISSUE #583: Backend: Set default role for new users
Now it's copied from OAuth, e.g. for github new users will get `ROLE_USER,SCOPE_read:user`. These roles are unknown to save-cloud, and we can't handle that user's actions. The default role should be `ROLE_VIEWER`; if we ever create tighter integrations with different OAuth providers, we can start using existing roles from there, but we don't plan to now.

petertrr:
Closes #583


api-gateway/src/main/kotlin/org/cqfn/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
```diff
@@ -34,7 +35,13 @@ class StoringServerAuthenticationSuccessHandler(
     ): Mono<Void> {
         logger.info("Authenticated user ${authentication.userName()} with authentication type ${authentication::class}, will send data to backend")
 
-        val user = authentication.toUser()
+        val user = authentication.toUser().apply {
+            // https://github.com/analysis-dev/save-cloud/issues/583
+            // fixme: this sets a default role for a new user with minimal scope, however this way we discard existing role
+            // from authentication provider. In the future we may want to use this information and have a mapping of existing
+            // roles to save-cloud roles.
+            role = Role.VIEWER.asSpringSecurityRole()
```
kgevorkyan, line 43: How it worked previously? If there wasn't any record in DB about user, he wasn't be able to do anything?
I thought, that he was able at least to view several public pages

kgevorkyan (approved):

api-gateway/src/main/kotlin/org/cqfn/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
petertrr: It actually didn't, we had to change the role manually for a new user (but with just a few users the problem wasn't as urgent)

Issue #208, opened 2021-07-28
Orchestrator: Mark agents as crashed after a number of skipped heartbeats


petertrr:
After, e.g., 4 skipped heartbeats agent should be marked as crashed and orchestrator should attempt to stop it, in case it's hanging. This task depends on #315 , because now heartbeats can be irregular, since they are waiting on process execution

akuleshov7:
also we need to make checks for the following:
1) If AGENT returned only part of results -> MARK ALL other tests as FAILED (with internal error);
2) If Agent stooped heartbeating for a configurable time ->  MARK ALL tests from this agent as FAILED (with internal error) and stop an agent;
3) Check why SAVE processBuilder does not work with save timeouts;

PR #616, opened 2022-03-14
Mark remaining test executions as failed if agent returned only part of results

kgevorkyan:
### What's done:
* If AGENT returned only part of results -> MARK ALL other tests as FAILED (with internal error); (continuation of [#208](https://github.com/analysis-dev/save-cloud/issues/208))
* Agent: store timestamps in heartbeats
* Add test

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/Heartbeat.kt
+++ b/save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/Heartbeat.kt
@@ -4,11 +4,11 @@
 
 package org.cqfn.save.agent
 
-import kotlinx.serialization.Contextual
 import org.cqfn.save.test.TestDto
-
-import kotlinx.serialization.Serializable
 import org.cqfn.save.utils.LocalDateTime
+
+import kotlinx.serialization.Contextual
+import kotlinx.serialization.Serializable
 
 /**
  * Progress of tests execution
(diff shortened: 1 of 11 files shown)


save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/Heartbeat.kt
```diff
@@ -27,11 +29,16 @@ data class ExecutionProgress(val percentCompletion: Int) {
...
 @Serializable
-data class Heartbeat(val agentId: String,
-                     val state: AgentState,
-                     val executionProgress: ExecutionProgress)
+data class Heartbeat(
+    val agentId: String,
+    val state: AgentState,
+    val executionProgress: ExecutionProgress,
+    @Contextual
+    val timestamp: LocalDateTime,
```
petertrr, line 40: Our `org.cqfn.save.utils.LocalDateTime` was intended as a workaround for Hibernate, which can store Java's LocalDateTime, but can't with kotlinx one. Here we don't depend on Hibernate, so we can use `kotlinx.datetime` normally, without workarounds with type aliases and custom serialization strategies.

petertrr (approved):

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/Heartbeat.kt
kgevorkyan: Changed to `kotlinx.datetime`

save-cloud-common/src/commonMain/kotlin/org/cqfn/save/agent/Heartbeat.kt
```diff
@@ -37,7 +37,7 @@ data class Heartbeat(
     val state: AgentState,
     val executionProgress: ExecutionProgress,
     @Contextual
```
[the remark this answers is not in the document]
kgevorkyan: Yeah, I also removed another trash

Issue #344, opened 2021-10-22
Orchestrator: Cache base images for test execution


petertrr:
Now we have a Dockerfile-like setup: image is based on an image from repo, then `apt` is used to install additional packages etc. This intermediate image should be cached and reused later.

petertrr:
Having `COPY resources $resourcesPath` instruction before all `apt-get` invocations is likely to invalidate docker build cache, because our resources change more frequently than system dependencies.

PR #618, opened 2022-03-15
Measure duration of docker commands called from orchestrator

petertrr:
* Measure time of `docker build` command. Will be useful when we need to tune performance in the future.

Related to #344

```
# HELP save_orchestrator_docker_build_seconds_max  
# TYPE save_orchestrator_docker_build_seconds_max gauge
save_orchestrator_docker_build_seconds_max{baseImage="openjdk:11",cmd="BuildImageCmd",} 13.381837036
# HELP save_orchestrator_docker_build_seconds  
# TYPE save_orchestrator_docker_build_seconds summary
save_orchestrator_docker_build_seconds_count{baseImage="openjdk:11",cmd="BuildImageCmd",} 1.0
save_orchestrator_docker_build_seconds_sum{baseImage="openjdk:11",cmd="BuildImageCmd",} 13.381837036
```

### DIFF
--- a/save-orchestrator/build.gradle.kts
+++ b/save-orchestrator/build.gradle.kts
@@ -92,7 +92,3 @@ kotlin.sourceSets.getByName("main") {
 tasks.withType<KotlinCompile>().forEach {
     it.dependsOn(generateVersionFileTaskProvider)
 }
-
-tasks.withType<Test> {
-    testLogging.showStandardStreams = true
-}
--- a/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/docker/ContainerManager.kt
+++ b/save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/docker/ContainerManager.kt
@@ -1,11 +1,13 @@
 package org.cqfn.save.orchestrator.docker
 
+import org.cqfn.save.domain.Sdk
 import org.cqfn.save.orchestrator.config.DockerSettings
 import org.cqfn.save.orchestrator.copyRecursivelyWithAttributes
 import org.cqfn.save.orchestrator.getHostIp
 
 import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.command.BuildImageResultCallback
+import com.github.dockerjava.api.model.BuildResponseItem
 import com.github.dockerjava.api.model.HostConfig
 import com.github.dockerjava.api.model.LogConfig
 import com.github.dockerjava.core.DefaultDockerClientConfig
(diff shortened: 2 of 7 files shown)


save-orchestrator/build.gradle.kts
```diff
@@ -92,7 +92,3 @@ kotlin.sourceSets.getByName("main") {
 tasks.withType<KotlinCompile>().forEach {
     it.dependsOn(generateVersionFileTaskProvider)
 }
-
-tasks.withType<Test> {
```
kgevorkyan, line 96: Why?

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/docker/ContainerManager.kt
```diff
@@ -148,13 +156,20 @@ class ContainerManager(private val settings: DockerSettings) {
...
         val buildImageResultCallback: BuildImageResultCallback = try {
-            dockerClient.buildImageCmd(dockerFile)
+            val buildCmd = dockerClient.buildImageCmd(dockerFile)
                 .withBaseDirectory(tmpDir)
                 .withTags(setOf(imageName))
                 .withExtraHosts(hostIp?.let {
                     setOf("host.docker.internal:$hostIp")
                 } ?: emptySet())
-                .start()
+            buildCmd.execTimed(meterRegistry, "save.orchestrator.docker.build", "baseImage", baseImage) { record ->
```
kgevorkyan, line 165: What if `buildCmd` will be an `emptySet`, which behavior we will see? I don't see any `isEmpty` like checks

save-orchestrator/build.gradle.kts
petertrr: I thought that it pollutes logs too much, on CI it's hard to find which test has failed among all those log lines. Do you find it useful? I usually navigate to HTML reports anyway, but that may be just me.

save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/docker/ContainerManager.kt
petertrr: No, `emptySet` is a possible value for `extraHosts` parameter. `buildCmd` is always a docker command instance

kgevorkyan: Ok, didn't notice

save-orchestrator/build.gradle.kts
kgevorkyan: Usually I used it in course of development, however, yes, we've a got a lot of extra logs, so I can enable it manually if I need

kgevorkyan (approved):

PR #620, opened 2022-03-16
Bugfix for MockServer utils

### LINKED ISSUE #617: Bug in MockServer utils
We use `regexKey` for determination which request were enqueued, however it behaves not as expected

In case, if we have several names, which starts from similar substr, like `/testExecution` and `/testExecutions/agent/...` it is impossible to enqueue `/testExecutions/agent/...`, because `/testExecution` always recognized instead of `/testExecutions/agent/...` despite the fact, that the latest one was passed to the mock server by full name.

There are also **no any warnings** `logger.warn("For path $path found more than one key from ResponsesMap: [$it]. Taking ${it.first()}")` for some reasons

Steps to reproduce: In https://github.com/analysis-dev/save-cloud/pull/616 there was provided new test `"should mark test executions as failed if agent returned only part of results"`, which use mock request `/testExecutions/agent/...`, however it didn't mocked until `mockServer.setDefaultResponseForPath("/testExecution/.*")` were changed
Without this change, `/testExecution/` absorb all requests, which starts from `testExecution`

sanyavertolet:
This PR closes #617 

### What's done:
* Fixed a problem when default response had higher priority than enqueued one.
* Improved logging


save-backend/src/test/kotlin/org/cqfn/save/backend/controller/ExecutionControllerTest.kt
```diff
@@ -220,7 +220,7 @@ class ExecutionControllerTest {
         }
 
         mockServerPreprocessor.enqueue(
-            "/rerunExecution",
+            "/rerunExecution.*$",
```
kgevorkyan, line 223: Do we really need `$` after `.*`? the `.*` should already match the remaining part of request string

test-utils/src/main/kotlin/org/cqfn/save/testutils/MockWebServerUtils.kt
```diff
@@ -20,35 +20,35 @@ class LoggingQueueDispatcher : Dispatcher() {
     private var failFastResponse: MockResponse = MockResponse().setResponseCode(HttpURLConnection.HTTP_NOT_FOUND)
 
     private fun getProperRegexKey(path: String?, setOfRegexes: Iterable<String>) = path?.let {
```
kgevorkyan, line 22: Let's may be write some tests for logic, corresponding with regex extraction
Thus, we will be sure, that similar requests will process properly and there will not some similar situations like in https://github.com/analysis-dev/save-cloud/issues/617

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/ExecutionControllerTest.kt
sanyavertolet: Well, `$` is there because we used to use `.containsMatchIn()` and now we moved to `.matches()` so now `$` is redundant

sanyavertolet: Done

save-preprocessor/src/test/kotlin/org/cqfn/save/preprocessor/controllers/DownloadProjectTest.kt
```diff
@@ -422,13 +422,13 @@ class DownloadProjectTest(
 
         // /updateExecution
         mockServerBackend.enqueue(
-            "/updateExecution$",
+            "/updateExecution",
```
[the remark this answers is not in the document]
sanyavertolet: No, we do not

test-utils/src/test/kotlin/org/cqfn/save/testutils/LoggingQueueDispatcherTest.kt
```diff
@@ -0,0 +1,91 @@
...
+    @Suppress("UnsafeCallOnNullableType", "UseEmptyCounterpart")
+    fun checkDefaultQueue() {
+        dispatcher.setDefaultResponseForPath("/users([/])?", MockResponse().setResponseCode(403))
+        dispatcher.enqueueResponse("/users/sanyavertolet", MockResponse().setResponseCode(200))
+        val permittedRequest = RecordedRequest(
+            "GET /users/sanyavertolet ",
+            MockResponse().headers,
+            emptyList(),
+            0,
+            Buffer().write("GET /example321 ".toByteArray()),
```
kgevorkyan, line 70: Shouldn't it be `GET /users/sanyavertolet`?

sanyavertolet: Yep, will fix it

sanyavertolet: Done

kgevorkyan (approved):
lgtm

PR #625, opened 2022-03-18
Support permission management on backend

### LINKED ISSUE #622: Add permission management endpoints in backend API

petertrr:
* Move Permission.kt to save-cloud-common
* `PermissionController`:
  * Any user with read access can see roles of other users in an organization
  * Only organization owner can set users' roles
* Tests

Closes #622

I tried to add swagger docs to `PermissionController`, even though it adds a lot of code with annotations.
image


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -0,0 +1,115 @@
...
+    )
+    @ApiResponse(responseCode = "200", description = "Permission added")
+    @ApiResponse(responseCode = "404", description = "Requested user or project doesn't exist or the project is hidden from the current user")
+    fun getRole(@PathVariable organizationName: String,
+                @PathVariable projectName: String,
+                @RequestParam userName: String,
+                authentication: Authentication,
+    ): Mono<Role> = permissionService.findUserAndProject(userName, organizationName, projectName)
+        .filter { (_, project) ->
+            // roles are available for all VIEWERs of public project and for any member of private project
```
kgevorkyan, line 65: Don't you mean, that user should be **at least** VIEWER in public projects?

save-backend/src/main/kotlin/org/cqfn/save/backend/service/LnkUserProjectService.kt
```diff
@@ -29,4 +31,16 @@ class LnkUserProjectService(private val lnkUserProjectRepository: LnkUserProject
         .ifEmpty { listOf(Role.VIEWER) }
         .singleOrNull()
         ?: throw IllegalStateException("Multiple roles are set for userId=$userId and project=$project")
+
+    /**
+     * Set role of [user] on a project [project] to [role]
+     */
+    @Suppress("KDOC_WITHOUT_PARAM_TAG")
+    fun addRole(user: User, project: Project, role: Role) {
```
kgevorkyan, line 39: May be not `addRole` but something like `save`? We don't check of existence of `user/project` and so on here, and just saving

save-backend/src/main/kotlin/org/cqfn/save/backend/service/PermissionService.kt
```diff
@@ -0,0 +1,70 @@
...
+        getRole(user, project)
+    }
+
+    /**
+     * @param user
+     * @param project
+     * @return role of [user] in [project], may throw exception if data is not consistent
+     */
+    @Suppress("UnsafeCallOnNullableType")
+    internal fun getRole(user: User, project: Project): Role = lnkUserProjectService.findRoleByUserIdAndProject(user.id!!, project)
```
kgevorkyan, line 39: any intention to make this version only `internal`? another is `public`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
petertrr: No, because `VIEWER` is our default role, and so every user is a viewer in a public project.

save-backend/src/main/kotlin/org/cqfn/save/backend/service/PermissionService.kt
petertrr: Yes, because it's used only in this service and in tests.

petertrr: Moved one of them into tests; it wasn't used anywhere else.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -0,0 +1,115 @@
...
+) {
+    @GetMapping("/{organizationName}/{projectName}")
+    @Operation(
+        description = "Get role for a user on a particular project",
+        parameters = [
+            Parameter(`in` = ParameterIn.HEADER, name = "X-Authorization-Source", required = true),
+        ]
+    )
+    @ApiResponse(responseCode = "200", description = "Permission added")
+    @ApiResponse(responseCode = "404", description = "Requested user or project doesn't exist or the project is hidden from the current user")
```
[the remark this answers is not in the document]
petertrr: Changed a bit, though this is a common approach for the whole API, I don't think we should describe it verbosely everywhere

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
petertrr: Updated description

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -0,0 +1,115 @@
...
+    private val projectPermissionEvaluator: ProjectPermissionEvaluator,
+) {
+    @GetMapping("/{organizationName}/{projectName}")
+    @Operation(
+        description = "Get role for a user on a particular project",
+        parameters = [
+            Parameter(`in` = ParameterIn.HEADER, name = "X-Authorization-Source", required = true),
+        ]
+    )
+    @ApiResponse(responseCode = "200", description = "Permission added")
```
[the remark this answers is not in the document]
petertrr: Thx, a typo. Changed

PR #629, opened 2022-03-21
Extract some spring boot annotations into `ApplicationConfiguration` class

petertrr:
This removes necessity to mock all repositories in simple `@WebFluxTest` tests.

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/SaveApplication.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/SaveApplication.kt
@@ -2,14 +2,9 @@ package org.cqfn.save.backend
 
 import org.cqfn.save.backend.configs.ConfigProperties
 import org.springframework.boot.SpringApplication
-import org.springframework.boot.actuate.autoconfigure.metrics.orm.jpa.HibernateMetricsAutoConfiguration
-import org.springframework.boot.autoconfigure.ImportAutoConfiguration
 import org.springframework.boot.autoconfigure.SpringBootApplication
-import org.springframework.boot.autoconfigure.domain.EntityScan
 import org.springframework.boot.context.properties.EnableConfigurationProperties
-import org.springframework.data.jpa.repository.config.EnableJpaRepositories
 import org.springframework.http.ResponseEntity
-import org.springframework.web.reactive.config.EnableWebFlux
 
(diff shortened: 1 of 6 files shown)


save-backend/src/test/kotlin/org/cqfn/save/backend/controller/CloningRepositoryControllerTest.kt
```diff
@@ -66,23 +63,9 @@ import kotlin.io.path.createFile
...
-    MockBean(GitRepository::class),
-    MockBean(StandardSuitesUpdateScheduler::class),
-    MockBean(UserRepository::class),
-    MockBean(AwesomeBenchmarksRepository::class),
-    MockBean(OrganizationRepository::class),
-    MockBean(OrganizationController::class),
-    MockBean(OrganizationService::class),
-    MockBean(LnkUserProjectRepository::class),
     MockBean(ProjectPermissionEvaluator::class),
+    MockBean(UserRepository::class),
```
kgevorkyan, line 68: Why this repo should be mocked, while other not? What's the difference?

save-backend/src/main/kotlin/org/cqfn/save/backend/SaveApplication.kt
```diff
@@ -19,11 +14,7 @@ typealias EmptyResponse = ResponseEntity<Void>
  * An entrypoint for spring for save-backend
  */
 @SpringBootApplication
-@EnableWebFlux
```
kgevorkyan, line 22: And this will be configured even without `@Import(ApplicationConfiguration)`? Or why we don't need configuration for application itself?

save-backend/src/test/kotlin/org/cqfn/save/backend/controller/CloningRepositoryControllerTest.kt
petertrr: Because this one is among dependencies of the controller under test. Other were mocked, because we accidentally enabled auto-configuration for all repositories.

save-backend/src/main/kotlin/org/cqfn/save/backend/SaveApplication.kt
petertrr: Yes, all configuration classes are auto-discovered

PR #631, opened 2022-03-22
Warning if list organization for user is empty

### LINKED ISSUE #623: Organizations should be selected from the LIST of EXISTING organizations created by THIS user
image

Cheshiriks:
What's done:
* added warning on field
Closes #623
Сни

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/SelectForm.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/SelectForm.kt
@@ -106,7 +106,7 @@ fun selectFormRequired(
 
             if (elements.isEmpty()) {
                 div("invalid-feedback d-block") {
-                    +"You don't have organizations"
+                    +"You don't have access to any organizations"
                 }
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/SelectForm.kt
```diff
@@ -104,7 +104,11 @@ fun selectFormRequired(
                 }
             }
 
-            if (props.validInput == false) {
+            if (elements.isEmpty()) {
+                div("invalid-feedback d-block") {
+                    +"You don't have organizations"
```
petertrr, line 109: Maybe `You don't have access to any organizations`?

petertrr (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/SelectForm.kt
Cheshiriks: Fixed

PR #632, opened 2022-03-23
Better support for development on Windows

petertrr:
* Add profile `docker-tcp` in orchestrator. In this profile orchestrator can communicate with docker daemon via HTTP API instead of Unix socket. This allows to run orchestrator in docker in Windows environment, if Docker daemon is exposed via HTTP port from Docker for Windows.
* Add `win` profile for preprocessor and orchestrator with more consistent shared folder paths


akuleshov7 (approved):
I suppose we need to finish It and merge asap, before new members joined us

Issue #626, opened 2022-03-19
Authorization: improvements


akuleshov7:
1) when something goes wrong with authorization providers (Github/Codehub) - user should not see a blank page with an error (it should be handled in a proper way)

2) if user does not has a permission (for example - not authorised to see some project or even was not logged in) we should not silently redirect him to the login page. For user it looks like a bug.

petertrr:
> 2. if user does not has a permission (for example - not authorised to see some project or even was not logged in) we should not silently redirect him to the login page. For user it looks like a bug.

Actually, after #602 it should be relatively easily fixable. I've kept the former logic for 401 codes (with redirect), but we can replace it with displaying custom error popup with login prompt (which will then redirect to main page, or just perform login right there)

If the user doesn't have permissions, we should already see the 403 error popup, I've only kept logic with redirects for 401, when the user is not logged it at all.

petertrr:
> 1. when something goes wrong with authorization providers (Github/Codehub) - user should not see a blank page with an error (it should be handled in a proper way)

Related: #409. We could make a dedicated page for unsuccessful logins, but a generic one will be useful as well.

PR #633, opened 2022-03-23
Add custom error page in backend

### LINKED ISSUE #409: Add default error page in backend
Instead of `Whitelabel Error Page`

petertrr:
* Backend: Add `error.html` and custom error endpoint for spring boot
* Frontend: Change `style-loader` (embeds CSS into JS) to [`mini-css-extract-plugin`](https://webpack.js.org/plugins/mini-css-extract-plugin/)(saves CSS into a separate file), use separate css file instead of bundled into JS
* Gateway: Add custom error endpoint, which delegates to backend's `/error`

Closes #409. After this we can make a dedicated login error page for #626 in a similar way.

image

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
@@ -46,7 +46,7 @@ class WebSecurityConfig(
             .pathMatchers("/", "/internal/**", "/actuator/**", *publicEndpoints.toTypedArray())
             .permitAll()
             // resources for frontend
-            .pathMatchers("/*.html", "/*.js*", "/img/**")
+            .pathMatchers("/*.html", "/*.js*", "/*.css", "/img/**", "favicon.ico")
             .permitAll()
     }
         .and().run {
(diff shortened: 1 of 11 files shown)


save-backend/src/test/kotlin/org/cqfn/save/backend/service/PermissionServiceTest.kt
```diff
@@ -79,7 +79,7 @@ class PermissionServiceTest {
     }
 
     private fun PermissionService.getRole(userName: String, projectName: String, organizationName: String): Mono<Role> =
-             findUserAndProject(userName, organizationName, projectName).map { (user, project) ->
+            findUserAndProject(userName, organizationName, projectName).map { (user, project) ->
```
petertrr, line 82: Very unexpected fact: in https://github.com/analysis-dev/save-cloud/runs/5646708887?check_suite_focus=true diktat_check actually failed, but gradle-cache-action failed silently and the check has been green. That's why we now have code style violation in master.

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebSecurityConfig.kt
```diff
@@ -46,7 +46,7 @@ class WebSecurityConfig(
             .pathMatchers("/", "/internal/**", "/actuator/**", *publicEndpoints.toTypedArray())
             .permitAll()
             // resources for frontend
-            .pathMatchers("/*.html", "/*.js*", "/img/**")
+            .pathMatchers("/*.html", "/*.js*", "/*.css", "/img/**", "favicon.ico")
```
kgevorkyan, line 49: But we don't have `favicon` what's for it?

kgevorkyan (approved):

PR #634, opened 2022-03-24
Create table for link users to organization

Cheshiriks:
What's done:
* created table lnk_user_organization
* created service and entity LnkUserOrganization

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/LnkUserOrganization.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/LnkUserOrganization.kt
@@ -1,0 +1,21 @@
+package org.cqfn.save.entities
+
+import javax.persistence.Entity
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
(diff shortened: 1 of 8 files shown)


save-cloud-common/src/jvmMain/kotlin/org/cqfn/save/entities/LnkUserOrganization.kt
```diff
@@ -0,0 +1,21 @@
...
+import javax.persistence.Entity
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+
+/**
+ * @property organization
+ * @property user in organization
+ */
+@Entity
+class LnkUserOrganization(
```
petertrr, line 12: Do we need something like `isAdmin` or a more general `organizationRole` here?

Cheshiriks: I thought about it. But I decided that at this stage we don't need it. If needed in the future, we will add.

petertrr (approved):

PR #636, opened 2022-03-24
Display proper message for unauthenticated access

### LINKED ISSUE #626: Authorization: improvements

petertrr:
* Frontend: Add separate handling of 401 responses in error modal
* Gateway: display error page if OAuth2 login attempt throws exception

image

Closes #626

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
@@ -119,9 +119,7 @@ internal fun Component<*, *>.classComponentResponseHandler(
     // dirty hack to determine whether this component contains `setResponse` in its context.
     // If we add another context with a function, this logic will break.
     val hasResponseContext = this.asDynamic().context is Function<*>
-    if (!hasResponseContext) {
-        noopResponseHandler(response)
-    } else {
+    if (hasResponseContext) {
         this.unsafeCast<Component<*, *>>().withModalResponseHandler(response)
     }
 }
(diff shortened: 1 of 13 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
```diff
@@ -116,12 +116,11 @@ suspend fun WithRequestStatusContext.post(
 internal fun Component<*, *>.classComponentResponseHandler(
     response: Response
 ) {
-    val isUnauthorized = response.status == 401.toShort()
     // dirty hack to determine whether this component contains `setResponse` in its context.
     // If we add another context with a function, this logic will break.
     val hasResponseContext = this.asDynamic().context is Function<*>
-    if (isUnauthorized || !hasResponseContext) {
-        redirectResponseHandler(response)
+    if (!hasResponseContext) {
```
kgevorkyan, line 122: Why do we need `noop` here, while it just returns `Unit`?

Could it be just like this, without else branch?

```suggestion
    if (hasResponseContext) {
        this.unsafeCast<Component<*, *>>().withModalResponseHandler(response)
    }
```

kgevorkyan (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/utils/RequestUtils.kt
petertrr: Yes, that makes sense, even though we don't have a valid fallback solution now. But anyway, we should always check that all class components have correct context type.

PR #645, opened 2022-03-28
Make organizations info available for anonymous users

### LINKED ISSUE #639: Backend: Organizations list should be available in anonymous mode
- Add endpoints to list of public endpoints
- Support nullable Authentiocation in controller

petertrr:
* Make `/api/organization/**` public
* Refine permission checks in OrganizationController.kt
* Change some API methods for organizations, change return type of annotated methods to Publisher
* Make list of projects by organization public

Closes #639

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
@@ -7,6 +7,7 @@ import org.cqfn.save.domain.OrganizationSaveStatus
 import org.cqfn.save.entities.Organization
 import org.slf4j.LoggerFactory
 import org.springframework.http.ResponseEntity
+import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
 import org.springframework.web.bind.annotation.*
 import java.time.LocalDateTime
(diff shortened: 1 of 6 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -123,10 +123,10 @@ class ProjectController(private val projectService: ProjectService,
      * @return project by name and organization name
      */
     @GetMapping("/get/projects-by-organization")
-    @PreAuthorize("hasRole('VIEWER')")
+    @PreAuthorize("permitAll()")
```
kgevorkyan, line 126: But what with private projects after this?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -18,37 +23,47 @@ import java.time.LocalDateTime
...
+    }
 
     /**
      * @param authentication an [Authentication] representing an authenticated request
      * @return list of organization by owner id
      */
     @GetMapping("/get/list")
-    fun getOrganizationsByOwnerId(authentication: Authentication): List<Organization> {
+    @PreAuthorize("permitAll()")
+    fun getOrganizationsByOwnerId(authentication: Authentication?): Flux<Organization> {
```
kgevorkyan, line 43: Why do we need here and in some other places to wrap with `Flux/Mono`, while it worked before without it?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
petertrr: They are filtered out by `hasPermission` check

kgevorkyan: Yes, but formally, we didn't need this `hasRole('VIEWER')` annotation before, since anyway we will correctly filter out only necessary once by `hasPermission`?

petertrr: Yes, but I think having those annotations on all methods makes code more consistent, especially since this controller is now public. It just shows that we didn't forget about permission check, it's just sort of no-op.

kgevorkyan (approved):

PR #647, opened 2022-03-28
Add choice between public and private project in CreateProject view

### LINKED ISSUE #638: Add choice between public and private project in CreateProject view

Cheshiriks:
What's done:
* added choice between public and private project

Closes #638

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
@@ -14,6 +14,8 @@ import org.cqfn.save.frontend.components.basic.inputTextFormOptional
 import org.cqfn.save.frontend.components.basic.inputTextFormRequired
 import org.cqfn.save.frontend.components.basic.selectFormRequired
 import org.cqfn.save.frontend.components.errorStatusContext
+import org.cqfn.save.frontend.externals.fontawesome.faQuestionCircle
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
```diff
@@ -352,15 +359,42 @@ class CreationView : AbstractView<Props, ProjectSaveViewState>(true) {
...
+                                                    input(classes = "form-check-input") {
+                                                        attrs.defaultChecked = !state.isPublic!!
+                                                        attrs["name"] = "projectVisibility"
+                                                        attrs["type"] = "radio"
+                                                        attrs["id"] = "isProjectPrivateSwitch"
+                                                        attrs["value"] = "false"
+                                                    }
+                                                    label("form-check-label") {
+                                                        attrs["htmlFor"] = "isProjectPrivateSwitch"
+                                                        +"Private"
```
petertrr, line 390: It would be useful to also have a popup here, which explains publicity: private projects are visible for user, organization admins and selected users, while public ones are visible for everyone.

petertrr (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
Cheshiriks: Yes, I agree. Added

PR #649, opened 2022-03-29
Add caching of the user role

### LINKED ISSUE #580: Add caching of the user role
It is necessary to implement caching of the user role so as not to constantly access the database.

Cheshiriks:
What's done:
* added caching of the user role
Closes #580


petertrr (commented):
I thought we were planning to use Hibernate's L2 cache? Also, do we need any additional dependencies? Which cache implementation will be used with these changes?

akuleshov7:
@Cheshiriks are you sure that this implementation really works and does L2 caching? The L1 cache is there by the default already, so you do not change anything. But L1 cache is tightly linked to the current Thread. 

I suggest you to read about L2 cache in Hibernate on Baeldung.

akuleshov7:
@nulls @icemachined we will need your expert suggestions and help on this as @Cheshiriks is stuck a little bit with that

sanyavertolet:
Will it ever be merged?

kgevorkyan:
👀👀👀

PR #685, opened 2022-04-11
Create settings view for users

### LINKED ISSUE #646: Create settings view for users
image

We should have settings page for the user with organizations/projects and related configurations

Cheshiriks:
What's done:
* Created settings view for users
userr
pag2
pag3

Closes #646


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -0,0 +1,49 @@
+package org.cqfn.save.backend.controllers
```
kgevorkyan, line 1: Since `username` can be similar for different sources (github, gittee), this logic should be probably reworked

Cheshiriks: Yes. I think we will fix this later, when the Source field will be filled in for users

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -96,16 +98,26 @@ class DownloadFilesController(
...
      * @return [Mono] with response
      */
     @Suppress("UnsafeCallOnNullableType")
     @PostMapping(value = ["/api/image/upload"], consumes = [MediaType.MULTIPART_FORM_DATA_VALUE])
-    fun uploadImage(@RequestPart("file") file: Mono<FilePart>, @RequestParam owner: String) =
-            additionalToolsFileSystemRepository.saveImage(file, owner).map { imageInfo ->
+    fun uploadImage(
+        @RequestPart("file") file: Mono<FilePart>,
+        @RequestParam owner: String,
+        @RequestParam(required = false) isOrganization: Boolean = true
```
petertrr, line 109: Maybe, we should introduce a enum for image kinds?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -96,16 +98,26 @@ class DownloadFilesController(
...
+    ) =
+            additionalToolsFileSystemRepository.saveImage(file, owner, isOrganization).map { imageInfo ->
                 ResponseEntity.status(
                     imageInfo.path?.let {
-                        organizationService.saveAvatar(owner, it)
-                        HttpStatus.OK
+                        if (isOrganization) {
```
petertrr, line 114: I think this is too much code for calling inside `ResponseEntity` creation. Please extract saving before creating `ResponseEntity`

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val userRepository: UserRepository,
+) {
+    /**
+     * @param userName username
+     * @return [ImageInfo] about user's avatar
+     */
+    @GetMapping("/{userName}/avatar")
+    @PreAuthorize("permitAll()")
+    fun avatar(@PathVariable userName: String): Mono<ImageInfo> = Mono.fromCallable {
+        userRepository.findByName(userName).get().avatar
```
petertrr, line 25: You can wrap repository query into our extension method `justOrNotFound` to correctly return 404 instead of throwing in `get()`, if user is absent.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -0,0 +1,54 @@
...
+    @PreAuthorize("permitAll()")
+    fun findByName(@PathVariable userName: String): Mono<UserInfo> = Mono.fromCallable {
+        userRepository.findByName(userName).get().toUserInfo()
+    }
+
+    /**
+     * @param newUserInfo
+     */
+    @PostMapping("/save")
+    fun saveUser(@RequestBody newUserInfo: UserInfo) {
```
petertrr, line 43: Any permission checks here?

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TimestampBasedFileSystemRepository.kt
```diff
@@ -125,7 +126,7 @@ class TimestampBasedFileSystemRepository(configProperties: ConfigProperties) {
                     .collect(Collectors.summingLong { it })
                     .map {
                         logger.info("Saved $it bytes into $this")
-                        val relativePath = ("/$owner/$name")
+                        val relativePath = if (isOrganization) ("/$imageName/$name") else ("/users/$imageName/$name")
```
petertrr, line 129: You are duplicating logic of storage path construction in several places in this class. It should be extracted into a method

save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
```diff
@@ -142,6 +145,39 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
                                 }
                             }
 
+                            Route {
+                                attrs {
+                                    path = "/:user/Settings/Profile"
```
petertrr, line 150:
```suggestion
                                    path = "/:user/settings/profile"
```
Why capital letters? Usually, everything is lowercase in URLs

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
```diff
@@ -206,3 +212,7 @@ fun topBar() = fc<TopBarProps> { props ->
         attrs.isOpen = isLogoutModalOpen
     }
 }
+
+private fun userSettings(userName: String) {
```
petertrr, line 216: Extremely confusing method naming, it should at least mention that it will cause redirect to a different page

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
```diff
@@ -189,9 +192,12 @@ fun topBar() = fc<TopBarProps> { props ->
                 // Dropdown - User Information
                 div("dropdown-menu dropdown-menu-right shadow animated--grow-in") {
                     attrs["aria-labelledby"] = "userDropdown"
-                    // FixMe: temporary disable Profile DropDown, will need to link it with the user in the future
-                    // dropdownEntry("cogs", "Profile")
-                    dropdownEntry("sign-out-alt", "Log out") {
+                    dropdownEntry(faCog, "Settings") {
+                        attrs.onClickFunction = {
+                            userSettings(props.userInfo?.name ?: "Unknown")
```
petertrr, line 197: I think we shouldn't display this dropdown for anonymous users at all and have a `log in` button instead. And now it should have not `Unknown`, but `Not logged in` or `Unauthenticated`

petertrr (changes requested):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
Cheshiriks: I agree. But it's easier to test locally

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: This method is called only by the user for himself. Are there any other permission checks needed?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/DownloadFilesController.kt
Cheshiriks: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TimestampBasedFileSystemRepository.kt
Cheshiriks: In this place, a relative path to the file itself is collected. And in another case, the full path to the directory itself

save-frontend/src/main/kotlin/org/cqfn/save/frontend/App.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
Cheshiriks: Remove method

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/DownloadFilesController.kt
Cheshiriks: Yes. Added enum and changed logic

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
petertrr: Does anything stop me from doing `curl -d '{"name": "Cheshiriks"}' saveourtool.com/api/users/save`?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
petertrr: Yeah, but maybe at least we can change text now? Not `Unknown`, but `Unauthenticated`

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TimestampBasedFileSystemRepository.kt
petertrr: Yes, but the directory name `users` is duplicated. You can extract relative path creation logic into a method, which accepts AvatarType, and for absolute paths call `rootDir.resolve(methodForRelativePath(type))`

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
Cheshiriks: Ok. Added

save-backend/src/main/kotlin/org/cqfn/save/backend/repository/TimestampBasedFileSystemRepository.kt
Cheshiriks: Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: Added annotation

petertrr: Again, I can be authenticated as myself and still change your user data. You should check if the user from `Authentication` object is the same that the user in `UserInfo`.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
```diff
@@ -194,7 +194,7 @@ fun topBar() = fc<TopBarProps> { props ->
                     attrs["aria-labelledby"] = "userDropdown"
                     dropdownEntry(faCog, "Settings") {
                         attrs.onClickFunction = {
-                            window.location.href = "#/${props.userInfo?.name}/settings/email"
+                            window.location.href = "#/${props.userInfo?.name ?: "Unauthenticated"}/settings/email"
```
petertrr, line 197: Maybe, we can already disable it if `props.userInfo?.name == null`? What's the point of showing settings view to anonymous users?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: Ok. Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -36,18 +38,22 @@ class UsersDetailsController(
...
+        if (user.id == userId) {
+            userRepository.save(user.apply {
+                email = newUserInfo.email
+                company = newUserInfo.company
+                company = newUserInfo.location
+                company = newUserInfo.gitHub
+                company = newUserInfo.linkedin
+                company = newUserInfo.twitter
+            })
+        }
```
petertrr, line 57: `else throw ResponseStatusException(HttpStatus.FORBIDDEN)`
Now you are not doing anything, but still returning `200 OK`

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/TopBar.kt
Cheshiriks: Agree. Fixed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: fixed

petertrr (approved):

PR #692, opened 2022-04-15
check-git-connectivity request shouldn't be sent if form fields are empty

### LINKED ISSUE #630: check-git-connectivity request shouldn't be sent if form fields are empty
Now we send a network request which will definitely fail.

Cheshiriks:
What's done:
* Added throw exception
Closes #630


petertrr (commented):
Sorry, in the issue I actually meant request from the frontend. Connectivity between frontend and backend is a more important part of communication in terms of optimization.

akuleshov7 (approved):

akuleshov7:
> Sorry, in the issue I actually meant request from the frontend. Connectivity between frontend and backend is a more important part of communication in terms of optimization.

But this one should also be done, I guess. After that Vlad can support warning on the FE.

PR #696, opened 2022-04-18
Configs for more flexible deployment

petertrr:
* Gateway/actuator: Don't check password if subnet matches (checking request origin is enough of a security measure, and also authenticated health checks are not supported well enough in Kubernetes)
* Backend/postprocessor: Read secrets base path from env. A bit of un-hardcoding paths from env postprocessor to allow more flexible configuration in other environemtns

Fixme (that we kind of forgot about): logs from postprocessor are not visible, see https://stackoverflow.com/a/42840529

### DIFF
--- a/api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
+++ b/api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
@@ -171,6 +171,7 @@ class WebSecurityConfig(
     fun actuatorSecurityWebFilterChain(
         http: ServerHttpSecurity
     ): SecurityWebFilterChain = http.run {
+        // Allow access to actuator only from a set of addresses or subnets, without any additional checks.
         authorizeExchange()
             .matchers(
                 AndServerWebExchangeMatcher(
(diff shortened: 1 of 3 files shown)


api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
```diff
@@ -185,22 +183,8 @@ class WebSecurityConfig(
                     }
                 )
             )
-            .authenticated()
+            .permitAll()
     }
-        .and().httpBasic().run {
-            val userFromProps = configurationProperties.basicCredentials?.split(' ')?.run {
```
kgevorkyan, line 191: do we still need corresponding field `basicCredentials` in `ConfigurationProperties`?

api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
```diff
@@ -185,22 +183,8 @@ class WebSecurityConfig(
                     }
                 )
             )
-            .authenticated()
+            .permitAll()
```
kgevorkyan, line 186: Maybe add some comments here too, why do we `permitAll` from now

api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
petertrr: No, removed

api-gateway/src/main/kotlin/org/cqfn/save/gateway/security/WebSecurityConfig.kt
petertrr: Yes, added

PR #697, opened 2022-04-20
Backend: set `Cache-Control` for static resources

petertrr:
* Support setting `Cache-Control` header for some resources served from backend
  * For `/api/avatars` it should work right away, because images with same name shouldn't be changed usually
  * For JS, CSS and HTML files we should use more fine-grained controls for caches, currently build timestamp is used as `Last-Modified` value

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
@@ -31,14 +31,18 @@ class WebConfiguration(
     fun staticResourceRouter() = router {
         GET("/{*resourcePath}") {
             val resourcePath = it.pathVariable("resourcePath")
-            val resource = ClassPathResource("static/$resourcePath")
+            val resource = if (resourcePath.isNotBlank() && resourcePath != "/") {
+                ClassPathResource("static/$resourcePath")
+            } else {
+                ClassPathResource("static/index.html")
+            }
             val cacheControl: (ServerResponse.BodyBuilder) -> ServerResponse.BodyBuilder = when (resource.file.extension) {
                 "js" -> { b -> b.cacheControl(10.minutes) { cachePublic() } }
                 "css" -> { b -> b.cacheControl(10.minutes) { cachePublic() } }
                 else -> { b -> b.cacheControl(CacheControl.noCache()) }
             }
             ok().run(cacheControl)
-                .bodyValue(resource)
+                .body(BodyInserters.fromResource(resource))
         }
     }
 
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
@@ -1,23 +1,18 @@
 package org.cqfn.save.backend.configs
 
-import org.cqfn.save.backend.utils.secondsToLocalDateTime
 import org.cqfn.save.backend.utils.toInstant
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
```diff
@@ -43,11 +71,38 @@ class WebConfiguration(
...
+            .bodyValue(resource)
+    }
+
+    private fun ServerResponse.BodyBuilder.cacheControl(duration: kotlin.time.Duration,
+                                                        cacheControlCustomizer: CacheControl.() -> CacheControl = { this },
+    ): ServerResponse.BodyBuilder =
+            cacheControl(CacheControl.maxAge(duration.toJavaDuration()).run(cacheControlCustomizer))
+
+    companion object {
+        private val shortExpirationTime = 10.minutes
```
kgevorkyan, line 105: Could you please describe your thoughts and observations about how it actually worked on practice?

 * After building the jar, if the time which is passed, less, than 10 minutes, than we won't make requests for checking the content and for uploading resources at all?
* If the elapsed time is more than 10 min, then we will make request, to check if the content (or some checksum) is changed, and only in this case will upload resources one more time?
* If content is the same, even, if it pass a lot of time, we keep our cache as is?

kgevorkyan: Also it will be good, if we won't download them always, if content is the same, but the elapsed time is large than 10 min
But this is not so critical, since before this moment we had no cache at all

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
```diff
@@ -1,35 +1,63 @@
 package org.cqfn.save.backend.configs
```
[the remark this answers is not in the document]
petertrr: After I added an explicit `Last-Modified` header, Firefox started to cache the resources too.

save-backend/src/main/kotlin/org/cqfn/save/backend/configs/WebConfiguration.kt
petertrr: There is an RFC about caching: https://datatracker.ietf.org/doc/html/rfc7234#section-5.2, also docs on MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#freshness.
Correct combination of `Last-Modified` in response and `If-Modified-Since` in request should make browsers cache content, while periodically querying the server and getting `304 Not Changed`. After we deploy a new version of save-cloud, `Last-Modified` changes and browser will (after some time) revalidate its cache. As I understand it, with `max-age=600` the browser must validate resource freshness each 10 minutes, after getting `304 Not Modified` it will reset resource age and serve it from cache for another 10 minutes.
That being said, I've noticed that Chrome caches content with only `Cache-Control`, while Firefox needs `Last-Modified` or `ETag` too.

There is [another technique](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#revved_resources) to ensure freshness of important resources (HTML, JS, CSS) by adding content hash or version directly to the filename (webpack supports this functionality out of the box). This way new files will have a different URL and will be downloaded unconditionally. We may add this in this PR or in the future too.

kgevorkyan: Great, thanks for detailed answer

About `revving` - good technique, I think, that we can play with it in future

kgevorkyan (approved):

PR #700, opened 2022-04-22
Start versioning of backend API

kgevorkyan:
### What's done:
* Start versioning of backend API - add v1 version into endpoints
* Create configuration beans for swagger ui, which are responsible for creation of tabs,
    which are grouping the endpoints by versions
* Add generated OpenAPI spec into repo

speens


save-backend/build.gradle.kts
```diff
@@ -6,6 +6,20 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
 plugins {
     kotlin("jvm")
+    // this plugin will generate generateOpenApiDocs task
+    // running this task, it will writes the OpenAPI spec into a backend-api-docs.json file in save-backend dir.
+    id("org.springdoc.openapi-gradle-plugin") version "1.3.4"
+}
+
+openApi {
+    apiDocsUrl.set("http://localhost:5000/internal/v3/api-docs/latest")
+    outputFileName.set("$rootDir/save-backend/backend-api-docs.json")
```
petertrr, line 16: We could also add a validation task, that checks that current API isn't breaking compatibility with previous with the same version. But it can be done in the future

petertrr (commented):
Do you plan to update JSON in the repo on every change in backend controllers? Maybe, it's better to make JSON and Yaml specs of API publically available to download from the running server, e.g. by creating additional paths in Nginx config?

save-backend/build.gradle.kts
```diff
@@ -6,6 +6,20 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
+}
+
+kotlin {
+    jvmToolchain {
+        (this as JavaToolchainSpec).languageVersion.set(JavaLanguageVersion.of(Versions.jdk))
```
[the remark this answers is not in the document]
kgevorkyan: At some moment, I've got the errors corresponding the different versions and added this
However, now I can't reproduce them, removed

save-backend/build.gradle.kts
akuleshov7: Yep, later. I even do not think that versioning is was not so needed now

akuleshov7 (approved):

kgevorkyan:
> Do you plan to update JSON in the repo on every change in backend controllers? Maybe, it's better to make JSON and Yaml specs of API publically available to download from the running server, e.g. by creating additional paths in Nginx config?

Added additional paths for nginx configuration
About JSON in repo - it seems, that there is no way to create spec without working backend instance, the convenient one way is to use gradle plugin, so it's necessary to update it manually

I also remember, that we discussed, that it will be good, if user can be able just to open repo without downloading all project to check the current API

petertrr (approved):

PR #702, opened 2022-04-22
Bugfix: error 500 on backend when adding unknown user

### LINKED ISSUE #701: Exception on wrong username while adding user to project
When adding a person with wrong username, backend throws an exception and user gets `500` error on frontend. Should be `404 Not Found`

sanyavertolet:
This PR closes #701 

### What's done:
- [x] Fixed incorrect exception throwing (now throwing ResponseStatusException)

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
@@ -112,7 +112,7 @@ class PermissionController(
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
         }
-        .zipWhen { Mono.justOrEmpty(projectService.findUserByName(setRoleRequest.userName)) }
+        .zipWith(Mono.justOrEmpty(projectService.findUserByName(setRoleRequest.userName)))
         .switchIfEmpty {
             Mono.error((ResponseStatusException(HttpStatus.NOT_FOUND)))
         }
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
@@ -112,13 +112,15 @@ class PermissionController(
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
         }
-        .filter {
+        .zipWhen { Mono.justOrEmpty(projectService.findUserByName(setRoleRequest.userName)) }
+        .switchIfEmpty {
+            Mono.error((ResponseStatusException(HttpStatus.NOT_FOUND)))
+        }
+        .filter { (project, user) ->
             // fixme: could be `@PreAuthorize`, but organizationService cannot be found smh
             val userId = (authentication.details as AuthenticationDetails).id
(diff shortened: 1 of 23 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -116,7 +116,9 @@ class PermissionController(
             // fixme: could be `@PreAuthorize`, but organizationService cannot be found smh
             val userId = (authentication.details as AuthenticationDetails).id
             val hasOrganizationPermissions = organizationService.canChangeRoles(organizationName, userId)
-            val hasProjectPermissions = projectService.canChangeRoles(it, userId, setRoleRequest.userName, setRoleRequest.role)
+            val hasProjectPermissions =
+                    projectService.canChangeRoles(it, userId, setRoleRequest.userName, setRoleRequest.role)
+                        ?: throw ResponseStatusException(HttpStatus.NOT_FOUND)
```
petertrr, line 121: I think checking whether the user exists can be moved before the actual permissions check logic, e.g.:
```kotlin
...
.filter { project: Project ->
        // if project is hidden from the user, who attempts permission update,
        // then we should return 404
        projectPermissionEvaluator.hasPermission(authentication, project, Permission.READ)
    }
.zip(Mono.just( /* find user */ ))
.switchIfEmpty { /* if any of zipped Monos completes empty, the resulting Mono will copmlete empty */  }  
.filter { (project, user) ->
}
```
Then in ProjectService you should expect that `otherUser` exists.

sanyavertolet: done

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -112,11 +112,15 @@ class PermissionController(
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
         }
-        .filter {
+        .zipWhen { Mono.justOrEmpty(projectService.findUserByName(setRoleRequest.userName)) }
```
petertrr, line 115: Why not simply
```suggestion
        .zipWith(Mono.justOrEmpty(projectService.findUserByName(setRoleRequest.userName)))
```

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
```diff
@@ -150,10 +154,14 @@ class PermissionController(
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
         }
-        .filter { project ->
+        .zipWhen { Mono.justOrEmpty(projectService.findUserByName(userName)) }
```
petertrr, line 157: And this one?
```suggestion
        .zipWith(Mono.justOrEmpty(projectService.findUserByName(userName)))
```

petertrr (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/PermissionController.kt
sanyavertolet: of course, sorry, thank you

PR #703, opened 2022-04-23
Add description in organization

Cheshiriks:
What's done:
added description in organization

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
@@ -1,10 +1,12 @@
 package org.cqfn.save.backend.controllers
 
 import org.cqfn.save.backend.StringResponse
+import org.cqfn.save.backend.service.LnkUserOrganizationService
 import org.cqfn.save.backend.service.OrganizationService
 import org.cqfn.save.backend.utils.AuthenticationDetails
 import org.cqfn.save.domain.ImageInfo
 import org.cqfn.save.domain.OrganizationSaveStatus
+import org.cqfn.save.domain.Role
 import org.cqfn.save.entities.Organization
 import org.slf4j.LoggerFactory
 import org.springframework.http.ResponseEntity
@@ -21,7 +23,10 @@ import java.time.LocalDateTime
  */
 @RestController
 @RequestMapping("/api/organization")
-internal class OrganizationController(private val organizationService: OrganizationService) {
+internal class OrganizationController(
+    private val organizationService: OrganizationService,
+    private val lnkUserOrganizationService: LnkUserOrganizationService,
+) {
     /**
      * @param organizationName
      * @return Organization
@@ -80,6 +85,37 @@ internal class OrganizationController(private val organizationService: Organizat
             ResponseEntity.ok(organizationStatus.message)
         }
         return Mono.just(response)
+    }
+
+    /**
+     * @param organization updateOrganization
+     * @param authentication an [Authentication] representing an authenticated request
+     * @return response
+     */
+    @PostMapping("/update")
+    @PreAuthorize("isAuthenticated()")
+    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> {
+        val userId = (authentication.details as AuthenticationDetails).id
+        val role = lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organization.name)
+        val response = if (role.priority >= 2) {
+            organizationService.updateOrganization(organization)
+            ResponseEntity.ok("Organization updated")
+        } else {
+            ResponseEntity.badRequest().body("Access denied")
+        }
+        return Mono.just(response)
+    }
+
+    /**
+     * @param organizationName
+     * @param authentication an [Authentication] representing an authenticated request
+     * @return role
+     */
+    @PostMapping("/role")
+    @PreAuthorize("isAuthenticated()")
+    fun getRoleOrganization(@RequestBody organizationName: String, authentication: Authentication): Mono<Role> {
+        val userId = (authentication.details as AuthenticationDetails).id
+        return Mono.fromCallable { lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organizationName) }
     }
 
     companion object {
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
@@ -9,6 +9,7 @@ import org.cqfn.save.domain.OrganizationSaveStatus
 import org.cqfn.save.domain.Role
 import org.cqfn.save.entities.Organization
 import org.slf4j.LoggerFactory
+import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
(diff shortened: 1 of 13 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -82,6 +87,37 @@ internal class OrganizationController(private val organizationService: Organizat
...
+    @PostMapping("/{organizationName}/update")
+    @PreAuthorize("isAuthenticated()")
+    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> {
+        val userId = (authentication.details as AuthenticationDetails).id
+        val role = lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organization.name)
+        val response = if (role.priority >= 2) {
+            organizationService.updateOrganization(organization)
+            ResponseEntity.ok("Organization updated")
+        } else {
+            ResponseEntity.badRequest().body("Access denied")
```
petertrr, line 104: Returning `400` here is too general. Since all organizations are public, you can return `403` here.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -448,10 +493,35 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
         }
     }
 
+    private fun setNewDescription(value: String) {
+        descriptionTmp = value
+    }
+
+    private fun onOrganizationSave() {
+        setState {
+            organization?.apply {
```
petertrr, line 502: Will it actually cause state change? I thought, state property needs to be assigned with a new value

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -87,10 +102,12 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
             val avatar = getAvatar()
             val organizationLoaded = getOrganization(props.organizationName)
             val projectsLoaded = getProjectsForOrganization()
+            val role = getRoleInOrganization()
```
akuleshov7, line 105: any default value for `isEditDisabled`?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -239,18 +256,46 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                                 }
                                 +"Description"
                             }
-                            button(classes = "btn btn-link text-xs text-muted text-left ml-auto") {
-                                +"Edit  "
-                                fontAwesomeIcon(icon = faEdit)
-                                attrs.onClickFunction = {
-
+                            if ((state.roleInOrganization?.priority ?: 0) >= 2) {
```
akuleshov7, line 259: very strange logic. You should rely on `isEditDisabled`. You have created a special state for it.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -239,18 +256,46 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
+                                button(classes = "btn btn-link text-xs text-muted text-left ml-auto") {
+                                    +"Edit  "
+                                    fontAwesomeIcon(icon = faEdit)
+                                    attrs.onClickFunction = {
+                                        turnEditMode(isOff = false)
```
akuleshov7, line 264: on click you set `isEditDisabled = false`? Why?

akuleshov7 (changes requested):
unify the logic for  isEditEnabled

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Don't hide. Allow field to be edited after clicking

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Yes, this update will change the state

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Added

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -82,6 +87,37 @@ internal class OrganizationController(private val organizationService: Organizat
...
+     * @param organization updateOrganization
+     * @param authentication an [Authentication] representing an authenticated request
+     * @return response
+     */
+    @PostMapping("/{organizationName}/update")
+    @PreAuthorize("isAuthenticated()")
+    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> {
+        val userId = (authentication.details as AuthenticationDetails).id
+        val role = lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organization.name)
+        val response = if (role.priority >= 2) {
```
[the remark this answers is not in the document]
Cheshiriks: Changed

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -448,10 +493,35 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
+            organization?.apply {
+                description = descriptionTmp
+            }
+        }
+        val headers = Headers().also {
+            it.set("Accept", "application/json")
+            it.set("Content-Type", "application/json")
+        }
+        scope.launch {
+            post("$apiUrl/organization/${props.organizationName}/update", headers, Json.encodeToString(state.organization))
```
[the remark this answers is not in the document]
Cheshiriks: Ok. Added check

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
petertrr: In pure React `setState` accepts an instance of state object; but I forgot that in Kotlin there is an extension function, that accepts a lambda.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -449,10 +489,43 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
         }
     }
 
+    private fun setNewDescription(value: String) {
+        descriptionTmp = value
+    }
+
+    private fun onOrganizationSave() {
```
petertrr, line 496: This function should do nothing if `state.organization == null`, but to achieve that you are currently doing two checks and launching a coroutine even if the value is null. Why not accept `Organization` as a non-nullable parameter in this function and perform nullability check before calling this function?

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -449,10 +489,43 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
+        }
+        val headers = Headers().also {
+            it.set("Accept", "application/json")
+            it.set("Content-Type", "application/json")
+        }
+        scope.launch {
+            val response = newOrganization?.let {
+                post("$apiUrl/organization/${props.organizationName}/update", headers, Json.encodeToString(newOrganization))
+            }
+            response?.let { resp ->
```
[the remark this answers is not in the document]
Cheshiriks: Yes I agree. But then the DiKtat begins to swear due to the use of the internal lambda

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Fixed

petertrr (approved):

akuleshov7 (approved):

PR #725, opened 2022-05-04
Kubernetes: Support database migrations using init-containers

petertrr:
* Helm: Add init-containers for save-backend Deployment to run database migrations with gradle and liquibase
* Gradle: Reading database credentials using path from environment variable
* Helm: Add configs to deploy MySQL into Kubernetes cluster (enabled by `mysql.external: false` in Helm values), update Secrets for dev profile, update README.md
* Backend: Correctly disable DockerSecretsDatabaseProcessor depending on profiles
* Backend: Use deferred logger in DockerSecretsDatabaseProcessor

Caveats:
* Gradle takes a lot of time to take off inside a container, because it needs to download a lot of dependencies. However, we have a lot of control on reading secrets and still can have depoyment procedures unified for different cases. To mitigate this, we can build our own base image for builds and publish to ghcr (update on every gradle and/or Kotlin version changes). It can also be used inside Github Actions.

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/postprocessor/DockerSecretsDatabaseProcessor.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/postprocessor/DockerSecretsDatabaseProcessor.kt
@@ -1,9 +1,8 @@
 package org.cqfn.save.backend.postprocessor
 
-import org.slf4j.LoggerFactory
 import org.springframework.boot.SpringApplication
 import org.springframework.boot.env.EnvironmentPostProcessor
-import org.springframework.context.annotation.Profile
+import org.springframework.boot.logging.DeferredLogFactory
 import org.springframework.core.env.ConfigurableEnvironment
 import org.springframework.core.env.PropertiesPropertySource
 import org.springframework.core.io.FileSystemResource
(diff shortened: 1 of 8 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/postprocessor/DockerSecretsDatabaseProcessor.kt
```diff
@@ -14,11 +13,18 @@ import java.util.Properties
...
-class DockerSecretsDatabaseProcessor : EnvironmentPostProcessor {
-    private val log = LoggerFactory.getLogger(DockerSecretsDatabaseProcessor::class.java)
+class DockerSecretsDatabaseProcessor(
+    logFactory: DeferredLogFactory
+) : EnvironmentPostProcessor {
+    private val log = logFactory.getLog(DockerSecretsDatabaseProcessor::class.java)
 
     override fun postProcessEnvironment(environment: ConfigurableEnvironment, application: SpringApplication) {
+        // Since `EnvironmentPostProcessor` is created before application context is fully initialized, usual means of registration control
+        // like `@Profile` cannot be used. `environment` here should contain at least profiles set from env variable or system properties.
```
kgevorkyan, line 23: so, previously, the annotation didn't actually worked?

save-cloud-charts/save-cloud/templates/backend-deployment.yaml
```diff
@@ -29,6 +29,33 @@ spec:
             - name: database-secret
               mountPath: {{ .Values.backend.dbPasswordFile }}
           {{- include "spring-boot.management" .Values.backend | nindent 10 }}
+      initContainers:
+        - name: git-cloner
+          image: alpine/git
+          args:
+            - clone
+            - --single-branch
+            - --
```
kgevorkyan, line 38: empty arg?

save-backend/src/main/kotlin/org/cqfn/save/backend/postprocessor/DockerSecretsDatabaseProcessor.kt
petertrr: Yes, neither annotation nor logging

save-cloud-charts/save-cloud/values-minikube.yaml
```diff
@@ -10,6 +10,7 @@ gateway:
     management.endpoints.web.exposure.include=*
     gateway.knownActuatorConsumers=172.0.0.0/8
 mysql:
-  ip: 192.168.65.2
+  external: false
```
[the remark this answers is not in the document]
petertrr: This file is supposed to be an additional source of data to override default `values.yaml`

save-cloud-charts/save-cloud/templates/backend-deployment.yaml
petertrr: Honestly I've just copied this one from an example, but `git clone` supports a double dash to separate all named arguments from repo url

kgevorkyan (approved):

Issue #723, opened 2022-05-04
Deleting organizations and projects at frontend


sanyavertolet:
For now we cannot delete projects and organizations. We need to 
- [x] Improve `updateProject` and `updateOrganization` methods in order to be able to mark project or organization as `DELETED`.
- [x] Implement deleting at frontend

sanyavertolet:
Implemented for organizations in #736

Issue #706, opened 2022-04-26
Global role icon


sanyavertolet:
We need to:
- [x] Provide some information to users who have global role with a higher priority than `VIEWER`;
- [x] Add `favicon`;
- [x] Add icons to replace source on frontend;
- [x] Create a method `OrganizationService#getGlobalRoleOrOrganizationRole`;
- [x] Create a method `ProjectService#getGlobalRoleOrProjectRole`;
- [x] Add confirmation modal when opening settings menu and having role `Role.SUPER_ADMIN`

sanyavertolet:
Decided in #736: 
> we can create a methods like `OrganizationService#getGlobalRoleOrOrganizationRole`

PR #736, opened 2022-05-06
Minor issues with permissions and organizations

### LINKED ISSUE #734: 403 for super admins
When `SUPERADMIN` tries to manage roles in groups or save organization description, he gets 403

sanyavertolet:
This PR closes #734 as well as a part of #723 

### What's done:
- [x] Implemented redirecting to OrganizationView on successful organization creation
- [x] Fixed 403 for global super admins
- [x] Added globalRole field to UserInfo
- [x] Added endpoint to get self globalRole
- [x] Moved `.canChangeRoles()` to `...PermissionEvaluator`
- [x] Fixed a bug in deleting organization

### What's left out of the score:
 * Better tests for `organizationPermissionEvaluator` and `projectPermissionEvaluator`


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -97,9 +106,12 @@ internal class OrganizationController(
      */
     @PostMapping("/{organizationName}/update")
     @PreAuthorize("isAuthenticated()")
+    @Suppress("UnsafeCallOnNullableType")
     fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> {
         val userId = (authentication.details as AuthenticationDetails).id
-        val role = lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organization.name)
+        val organizationRole = lnkUserOrganizationService.findRoleByUserIdAndOrganizationName(userId, organization.name)
+        val globalRole = userDetailsService.getGlobalRole(authentication)
+        val role = listOf(organizationRole, globalRole).maxByOrNull { it.priority }!!
```
petertrr, line 114: There is a lot of repetitions of this construction: get role, get glolbal role, keep highest. It should be extracted somewhere.

Also, it starts to concern me now: if we allow any action to superadmin, performing all the checks under the hood, it can become really easy do damage data from frontend by mistake, e.g. change change another organization's description thinking it's yours. I think we should add additional indication on frontend that such actions are performed using superadmin permissions rather than usual permissions.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ManageUserRoleCard.kt
```diff
@@ -177,7 +177,7 @@ fun manageUserRoleCardComponent(
                 it.decodeFromJsonString<String>()
             }
             .toRole()
-        setSelfRole(role)
+        setSelfRole(listOf(role, props.selfUserInfo.globalRole ?: Role.NONE).maxByOrNull { it.priority }!!)
```
petertrr, line 180: And this, imo, is super unobvious, see the previous comment

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -89,7 +93,9 @@ class CreateOrganizationView : AbstractView<Props, OrganizationSaveViewState>(tr
                     post("$apiUrl/organization/save", headers, Json.encodeToString(newOrganizationRequest))
 
             if (responseFromCreationOrganization.ok) {
-                window.location.href = "${window.location.origin}#/${organizationName.replace(" ", "%20")}/"
```
petertrr, line 92: What was wrong with this redirect? Incorrect value of organization name?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: ok, will add some logic to get rid of repetitions.

Talking about superadmins doing something by mistake, I should say that we have some kind of issue for that (#706). I will implement it there because this PR is sort of bugfix. Your idea is good, but do we need to implement it taking into consideration that there will be only 6 of us who is able to damage something having superadmin roles

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/ManageUserRoleCard.kt
sanyavertolet: Yes, ok, makes sense

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreateOrganizationView.kt
sanyavertolet: The name was fine. First problem is that we used `ButtonType.submit`. As far as I understood, redirects are not possible with it. Second issue, I assume, is that I changed `window.location.origin` inside of coroutine. When I moved it out of coroutine it started to work fine.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
petertrr: > a function somewhere that gets global role and either organizationRole or projectRole

Yes, I think we can do that in backend, e.g. create a methods like `OrganizationService#getGlobalRoleOrOrganizationRole`

Regarding the second topic, ok, let's properly implement indication uner #706.

sanyavertolet: Ok, mentioned it there.

petertrr (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreateOrganizationView.kt
petertrr: Sounds a bit strange, but I can believe that some stuff can't be done from a coroutine.

Issue #748, opened 2022-05-12
`favicon.ico` should be supported in api-gateway


petertrr:
@sanyavertolet proxying requests for favicon is not implemented in api-gateway, so it's still inaccessible in the real environment. Please fix it, as well as permission checks for images on gateway (that logic is currently duplicated between gateway and backend)

_Originally posted by @petertrr in https://github.com/analysis-dev/save-cloud/issues/745#issuecomment-1124939295_

PR #749, opened 2022-05-12
Redirects of frontend

### LINKED ISSUE #747: Redirect on project creation
We need to be redirected to project view after the project was created

sanyavertolet:
This PR closes #747 and #748 

### What's done:
- [x] Added redirect when creating project;
- [x] Added page reload when need to log in;
- [x] Made creator of the project be the first owner of it;
- [x] Updated resource_route in api-gateway in order to support favicon.

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
@@ -1,12 +1,15 @@
 package org.cqfn.save.backend.controllers
 
 import org.cqfn.save.backend.StringResponse
+import org.cqfn.save.backend.repository.UserRepository
 import org.cqfn.save.backend.security.ProjectPermissionEvaluator
 import org.cqfn.save.backend.service.GitService
+import org.cqfn.save.backend.service.LnkUserProjectService
 import org.cqfn.save.backend.service.OrganizationService
 import org.cqfn.save.backend.service.ProjectService
 import org.cqfn.save.backend.utils.AuthenticationDetails
 import org.cqfn.save.domain.ProjectSaveStatus
+import org.cqfn.save.domain.Role
 import org.cqfn.save.entities.GitDto
 import org.cqfn.save.entities.NewProjectDto
 import org.cqfn.save.entities.Project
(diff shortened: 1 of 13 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -149,6 +155,9 @@ class ProjectController(private val projectService: ProjectService,
             val saveGit = gitService.saveGit(it, projectId)
             log.info("Save new git id = ${saveGit.id}")
         }
+        val project = projectService.findById(projectId).get()
+        val owner = userRepository.findById(userId).get()
+        lnkUserProjectService.setRole(owner, project, Role.OWNER)
```
petertrr, line 160: That's a shame that we need two DB queries only to comply with JPA model, while in reality mapping uses only IDs that we already have. However I'm not sure how it can be changed and still be readable. Maybe using `@Query` in repository class and creating a new method, that accepts only IDs

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -93,9 +87,8 @@ class CreateOrganizationView : AbstractView<Props, OrganizationSaveViewState>(tr
                     post("$apiUrl/organization/save", headers, Json.encodeToString(newOrganizationRequest))
 
             if (responseFromCreationOrganization.ok) {
-                setState {
-                    isToRedirect = true
-                }
+                window.location.href = "${window.location.origin}#/${organizationName.replace(" ", "%20")}/"
+                window.location.reload()
```
petertrr, line 91: Why do we need reloading here?

sanyavertolet: It seems that we need to rerender the page so either we need to reload it, or change the state. @akuleshov7 finds it better to reload it.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
petertrr: Yes, something like this.

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreateOrganizationView.kt
petertrr: There is also a way to redirect [using react-router api](https://reactrouter.com/docs/en/v6/api#usenavigate) and we should start using it eventually, maybe not in this PR

sanyavertolet: Ok, will take into consideration.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
sanyavertolet: Done

PR #751, opened 2022-05-13
Bugfix for OrganizationSettingsMenu

### LINKED ISSUE #730: 404 at organizationSettingsMenu
Wrong url is used in requests:
`https://saveourtool.com/api/v1/organizations//users`

sanyavertolet:
This PR closes #730.

### What's done:
- [x] Fixed a bug with OrganizationSettingsMenu: now passing organizationName to OrganizationSettingsMenu

### DIFF
--- a/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
+++ b/save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
@@ -435,7 +435,6 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
     private fun RBuilder.renderSettings() {
         child(organizationSettingsMenu) {
             attrs.organizationName = props.organizationName
-            attrs.selfRole = Role.VIEWER
             attrs.currentUserInfo = props.currentUserInfo ?: UserInfo("Undefined")
         }
     }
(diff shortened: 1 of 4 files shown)


akuleshov7 (approved):
quite easy

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
```diff
@@ -434,8 +434,8 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
 
     private fun RBuilder.renderSettings() {
         child(organizationSettingsMenu) {
-            attrs.organization = state.organization!!
-            attrs.selfRole = Role.SUPER_ADMIN
+            attrs.organizationName = props.organizationName
+            attrs.selfRole = Role.VIEWER
```
petertrr, line 438: I can see that this prop is not used anywhere, is it needed?

akuleshov7:
BUT you need to test the following SCENARIO: 
1) request the organization with a FAKE name: BLABLA
2) see the result

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/OrganizationView.kt
akuleshov7: what is the `attrs.selfRole`

sanyavertolet: It used to be useful but now we do not need It. Removed it from both project and organization.

petertrr (approved):

PR #752, opened 2022-05-13
Extract common functionality of save-api into the separate module

kgevorkyan:
### What's done:
* Extract common functionality of save-api into the separate module in aim to use it as library from third-party projects
* Configure publishing of save-api lib


save-api/build.gradle.kts
```diff
@@ -26,13 +23,24 @@ kotlin {
...
     implementation(libs.ktor.client.logging)
-    implementation(libs.kotlinx.serialization.properties)
 }
+
+publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            groupId = "org.cqfn"
+            artifactId = "save-cloud-api"
+            version = "0.1.2"
```
petertrr, line 40: Is `project.version` fine here?

.github/workflows/save-api-examples-release.yml
```diff
@@ -0,0 +1,53 @@
...
+  OSSRH_USERNAME: ${{ secrets.SONATYPE_USER }}
+  OSSRH_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+
+jobs:
+  release:
+    name: Build release
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
```
petertrr, line 22: It's a jvm-only artifact, I think we don't need a matrix build

.github/workflows/save-api-examples-release.yml
```diff
@@ -0,0 +1,53 @@
+name: Create release
+
+on:
+  push:
+    branches:
+      - main
```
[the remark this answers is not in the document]
kgevorkyan: 🤦‍♂️ Sure, thanks

save-api/build.gradle.kts
kgevorkyan: Actually I'm lost in our versioning in save-cloud

.github/workflows/save-api-examples-release.yml
kgevorkyan: Fixed

save-api/build.gradle.kts
kgevorkyan: Changed to `v1`

petertrr: Let's still have `project.version` and have a common versioning for the whole repository. Once we add breaking changes into API, that require a new release of the client library, we will create a git tag on the whole repository and release the library under the corresponding version.

petertrr (approved):

PR #759, opened 2022-05-16
Move statistics for pie chart to backend

### LINKED ISSUE #673: ProjectStatisticsMenu improvements
* Statistics shouldn't be calculated on the frontend; data for the chart should be aggregated on backend instead.
* Since we have retention policy for executions, details data can become unavailable. This should be handled with a proper message.
* Table should use server side pagination (#746)

Cheshiriks:
* move statistics for pie chart to backend
Closes #673

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
@@ -9,6 +9,7 @@ import org.cqfn.save.backend.service.TestExecutionService
 import org.cqfn.save.backend.utils.justOrNotFound
 import org.cqfn.save.domain.TestResultLocation
 import org.cqfn.save.domain.TestResultStatus
+import org.cqfn.save.info.DataPieChart
 import org.cqfn.save.permission.Permission
 import org.cqfn.save.test.TestDto
 import org.cqfn.save.v1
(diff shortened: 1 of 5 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -100,6 +101,25 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
                 }
             }
 
+    /**
+     * @param executionId an ID of Execution to group TestExecutions
+     * @param authentication
+     * @return a list of [TestExecutionDto]s
+     */
+    @GetMapping(path = ["/api/$v1/testLatestExecutionsPieChart"])
```
petertrr, line 109: I think it's too specific to have it in common API. You could have a method `/testLatestExecutionsStatistics`, that returns a map `test suite name` -> `number of tests` or something similar. Then it can be reused not only on our frontend, but in any integration. Also color is needed only on frontend, so there is no need to calculate it on backend.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -100,6 +101,25 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
...
+        @RequestParam executionId: Long,
+        authentication: Authentication,
+    ): Mono<List<DataPieChart>> =
+            justOrNotFound(executionService.findExecution(executionId)).filterWhen {
+                projectPermissionEvaluator.checkPermissions(authentication, it, Permission.READ)
+            }.mapNotNull {
+                testExecutionService.getTestExecutions(executionId).groupBy { it.test.testSuite.name }.map { (testSuiteName, testExecutions) ->
+                    DataPieChart(testSuiteName, testExecutions.count())
+                }
+            }
```
petertrr, line 121:
```suggestion
            }.flatMapIterable {
                testExecutionService.getTestExecutions(executionId).groupBy { it.test.testSuite.name }
            }.map { (testSuiteName, testExecutions) ->
                DataPieChart(testSuiteName, testExecutions.count())
            }
```
This way you won't need `Mono<List<>>` and will have less nested lambdas

petertrr:
Closed because I thought tests are counted on frontend, but counting on backend has been already implemented

Issue #515, opened 2021-12-23
Frontend: Check if debug info exists before displaying the active link


petertrr:
Also, if debug info is not present, display something instead of an empty page

akuleshov7:
This is about the debug info that is available for all tests on the executionView. 
Also while the assignee of this issue will investigate it - probably he will help with making this view more user-friendly.

akuleshov7:
<img width="1727" alt="image" src="https://user-images.githubusercontent.com/58667063/168488610-dbf0e123-c604-4050-a48e-eba7602e24e1.png">

@petertrr could you please investigate why we are getting 404 now?

akuleshov7:
and 400 for Pending executions OH, @petertrr welcome to hell

<img width="1727" alt="image" src="https://user-images.githubusercontent.com/58667063/168489084-358bd47c-0480-4b1d-9bbf-cedbcfbd4874.png">

petertrr:
> <img alt="image" width="1727" src="https://user-images.githubusercontent.com/58667063/168488610-dbf0e123-c604-4050-a48e-eba7602e24e1.png">
> 
> @petertrr could you please investigate why we are getting 404?

I believe these particular executions have been run through another instance of save-backend, and because of this debug info files are not available for our usual backend. Should've cleaned those executions up in time..

petertrr:
> and 400 for Pending executions OH, @petertrr welcome to hell

This one is because `agentContainerId` is null in requests. And it's null there, because it's null in the fetched data, because execution must've hanged at some point.

akuleshov7:
> > and 400 for Pending executions OH, @petertrr welcome to hell
> 
> This one is because `agentContainerId` is null in requests. And it's null there, because it's null in the fetched data, because execution must've hanged at some point.

We should handle properly all errors in this case… Not friendly

petertrr:
> We should handle properly all errors in this case… Not friendly

I believe for this we should also prioritize #331 to properly display errors for executions that fail early.

petertrr:
With #760 every test execution in the table has `hasDebugInfo` property. This property should be used on the frontend to set correct appearance of rows and link for debug info. Also, while `agentId` is null, link for debug info should also be disabled.

cc @akuleshov7

petertrr:
Resolved in #893 by @akuleshov7

PR #760, opened 2022-05-16
Check if test executions have debug info

petertrr:
* Add flag into TestExecutionDto
* Set flag in some endpoints

Part of #515 only on backend

### DIFF
--- a/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
+++ b/save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
@@ -1,14 +1,22 @@
 package org.cqfn.save.backend.controllers
 
+import io.swagger.v3.oas.annotations.Parameter
+import io.swagger.v3.oas.annotations.enums.ParameterIn
+import io.swagger.v3.oas.annotations.tags.Tag
+import io.swagger.v3.oas.annotations.tags.Tags
 import org.cqfn.save.agent.AgentState
 import org.cqfn.save.agent.TestExecutionDto
 import org.cqfn.save.agent.TestSuiteExecutionStatisticDto
+import org.cqfn.save.backend.configs.ApiSwaggerSupport
+import org.cqfn.save.backend.repository.TestDataFilesystemRepository
 import org.cqfn.save.backend.security.ProjectPermissionEvaluator
 import org.cqfn.save.backend.service.ExecutionService
 import org.cqfn.save.backend.service.TestExecutionService
 import org.cqfn.save.backend.utils.justOrNotFound
+import org.cqfn.save.core.utils.runIf
 import org.cqfn.save.domain.TestResultLocation
 import org.cqfn.save.domain.TestResultStatus
+import org.cqfn.save.from
 import org.cqfn.save.permission.Permission
 import org.cqfn.save.test.TestDto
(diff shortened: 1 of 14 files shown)


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -49,24 +63,43 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
...
+        example = "basic"
+    )
     @Suppress("LongParameterList", "TOO_MANY_PARAMETERS", "TYPE_ALIAS")
     fun getTestExecutions(
         @RequestParam executionId: Long,
         @RequestParam page: Int,
         @RequestParam size: Int,
         @RequestParam(required = false) status: TestResultStatus?,
         @RequestParam(required = false) testSuite: String?,
+        @RequestParam(required = false, defaultValue = "false") checkDebugInfo: Boolean,
```
kgevorkyan, line 83: Why defaultValue = "false"?

kgevorkyan (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
petertrr: We should perform additional operations (here - some filesystem queries) only if it's explicitly requested by the consumer

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
```diff
@@ -49,24 +63,43 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
      * @param status
      * @param testSuite
      * @param authentication
+     * @param checkDebugInfo if true, response will contain information about whether debug info data is available for this test execution
      * @return a list of [TestExecutionDto]s
      */
     @GetMapping(path = ["/api/$v1/testExecutions"])
+    @Parameter(
```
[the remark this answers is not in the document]
petertrr: Good idea, added

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/TestExecutionController.kt
kgevorkyan: But if it's not requested, but debug info doesn't exist. we will get previous problem again?

petertrr: It should be handled on frontend properly: if there is a possibility that debug info is requested, then its presence should be checked. If component doesn't show any related info, then it shouldn't care whether it's available or not.
Also - I've updated the PR and made this field nullable. This way we can distinguish cases when debug info is unavailable and when it hassn't been checked

kgevorkyan: Didn't look, that this method used not only by frontend
Sure, of course, should be `false`

petertrr: It's used currently only by frontend, but in several places, only some of them are related to debug info. If this method is used by API integrations, then they also may need this information only someties

PR #763, opened 2022-05-16
Limit names of the project and organisations by the length

### LINKED ISSUE #672: Limit names of the project and organisations by the length
We should limit names by length, so it won't break the frontend.
Also names should be a one single name without spaces

Cheshiriks:
* Added limit names of the project and organisations by the length
Closes #672

naame


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
```diff
@@ -241,7 +241,8 @@ class CreationView : AbstractView<Props, ProjectSaveViewState>(true) {
             setState { isValidOrganization = true }
         }
 
-        if (fieldsMap[InputTypes.PROJECT_NAME].isNullOrBlank()) {
+        val projectName = fieldsMap[InputTypes.PROJECT_NAME]
+        if (projectName.isNullOrBlank() || projectName.contains(" ") || projectName.length > 64) {
```
petertrr, line 245: We have a common method for error message, but validation is duplicated in two places. Please extract validation into a method too

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/basic/InputForms.kt
```diff
@@ -87,8 +93,14 @@ internal fun RBuilder.inputTextFormRequired(
                 }
 
                 if (!validInput) {
-                    div("invalid-feedback d-block") {
-                        +"Please input a valid ${form.str}"
+                    if (isProjectOrOrganizationName) {
+                        div("invalid-feedback d-block") {
+                            +"Please input a valid ${form.str}. The name must not be longer than 64 characters and contain spaces."
```
petertrr, line 98:
```suggestion
                            +"Please input a valid ${form.str}. The name can be no longer than 64 characters and can't contain any spaces."
```
I think it sounds a little bit better

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/CreationView.kt
Cheshiriks: Fixed

petertrr (approved):

Issue #755, opened 2022-05-14
The logic for the deleting of projects is not working 


akuleshov7:
Before:
<img width="1154" alt="image" src="https://user-images.githubusercontent.com/58667063/168431760-1f29a78c-b193-4749-9825-f1c83445df6b.png">

After (nothing changed):
<img width="1661" alt="image" src="https://user-images.githubusercontent.com/58667063/168431823-0c8a6782-135a-40c9-b9cf-ed69bc463f38.png">

petertrr:
The root cause is a redundant `window.reload()` right after the API request. Coroutine was forced to cancel and we didn't see any indication of an error

PR #765, opened 2022-05-16
Removed useless errors

akuleshov7:
### What's done:
- Removed useless errors that are triggered when the request failed;
- We already have global errors for this case;


save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/HistoryView.kt
```diff
@@ -254,7 +246,6 @@ class HistoryView : AbstractView<HistoryProps, HistoryViewState>(false) {
             setState {
                 isDeleteExecutionWindowOpen = false
             }
-            window.location.reload()
```
petertrr, line 257: Removing this line should also fix #755

petertrr (approved):

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
```diff
@@ -502,6 +489,9 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
                     }
                     val newGitDto = gitDto?.copy(url = urlWithTests, branch = newBranch, hash = newCommit)
                         ?: GitDto(url = urlWithTests, branch = newBranch, hash = newCommit)
+
+                    console.log(newGitDto)
```
kgevorkyan, line 493: Maybe add this one too into ForbiddenMethodCall in detekt?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ExecutionController.kt
```diff
@@ -253,6 +254,21 @@ class ExecutionController(private val executionService: ExecutionService,
             }
     }
 
+    @GetMapping(path = ["/api/$v1/getTestRootPathByExecutionId"])
+    @Transactional
+    fun getTestRootPathByExecutionId(@RequestParam id: Long, authentication: Authentication): Mono<String> =
+        Mono.justOrEmpty(executionService.findExecution(id))
+            .switchIfEmpty() {
+                Mono.error(ResponseStatusException(HttpStatus.NO_CONTENT))
```
petertrr, line 262: Why `204`? Maybe, `NOT_FOUND`?

akuleshov7: Yeah, good idea. My intention was to explicitly mark that we don't have content in the response

save-frontend/src/main/kotlin/org/cqfn/save/frontend/components/views/ProjectView.kt
akuleshov7: Mentioned that in https://github.com/analysis-dev/diktat/issues/954 assigned to @nulls

akuleshov7: Both methods: `['kotlin.js.console.log']` and `['kotlin.js.console']` do not work

PR #769, opened 2022-05-16
Better deletions on backend

### LINKED ISSUE #766: Create new endpoints for deleting projects/organizations
We need to have `delete` endpoints in project and organization controllers that would be used in SAVE.
These endpoints should simply update the status of the Project/Organization

sanyavertolet:
This PR closes #766 

### What's done:
- [x] Implemented new endpoint for organization deletion
- [x] Implemented new endpoint for project deletion
- [x] Added tests on organization deletion
- [x] Added tests on project deletion


save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -197,6 +192,27 @@ class ProjectController(
...
+     * @param project
+     * @param authentication
+     * @return response
+     */
+    @PostMapping("/delete")
+    fun deleteProject(@RequestBody project: Project, authentication: Authentication): Mono<StringResponse> =
+            projectService.findWithPermissionByNameAndOrganization(
+                authentication, project.name, project.organization, Permission.DELETE
+            )
+                .filter { projectPermissionEvaluator.hasPermission(authentication, project, Permission.DELETE) }
```
petertrr, line 205: Isn't it duplication of `findWithPermission...`?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -114,6 +114,23 @@ internal class OrganizationController(
...
 
+    /**
+     * @param organizationName
+     * @param authentication
+     */
+    @DeleteMapping("/{organizationName}/delete")
+    @PreAuthorize("isAuthenticated()")
+    fun deleteOrganization(@PathVariable organizationName: String, authentication: Authentication): Mono<StringResponse> {
+        val role = lnkUserOrganizationService.getGlobalRoleOrOrganizationRole(authentication, organizationName)
+        val response = if (role.priority >= Role.OWNER.priority) {
```
petertrr, line 125: Can we use `organizationPermissionEvaluator` here, like you in ProjectController?

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
```diff
@@ -197,6 +192,27 @@ class ProjectController(
             ResponseEntity.ok(projectStatus.message)
         }
 
+    /**
+     * @param project
+     * @param authentication
+     * @return response
+     */
+    @PostMapping("/delete")
```
akuleshov7, line 200: why not `DeleteMapping` here?

akuleshov7 (approved):

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -114,6 +114,23 @@ internal class OrganizationController(
...
+            organizationService.deleteOrganization(organizationName)
+            ResponseEntity.ok("Organization deleted")
+        } else {
+            ResponseEntity.status(HttpStatus.FORBIDDEN).build()
+        }
+        return Mono.just(response)
```
akuleshov7, line 131: btw - please unify  the body of the function into a single Mono (functionally with `map`)

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
sanyavertolet: The point is that `WebTestServer` with `.delete()` method doesn't support `.body()` and the idea was to pass `Project` with a request, but now we decided to pass only `organizationName` and `projectName` instead.

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
```diff
@@ -114,6 +114,23 @@ internal class OrganizationController(
         return Mono.just(response)
     }
 
+    /**
+     * @param organizationName
+     * @param authentication
+     */
+    @DeleteMapping("/{organizationName}/delete")
+    @PreAuthorize("isAuthenticated()")
```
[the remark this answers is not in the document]
sanyavertolet: No need, will delete it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
sanyavertolet: yep, seems like it is

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: ok, will do it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/OrganizationController.kt
sanyavertolet: ok, will do it

save-backend/src/main/kotlin/org/cqfn/save/backend/controllers/ProjectController.kt
akuleshov7: So? you decided to change parameters and will change Mapping?

sanyavertolet: Yes, will use DELETE method instead of POST

petertrr (approved):

Issue #328, opened 2021-10-19
Frontend: Display loading animation, when TableComponent is loading the next page


petertrr:
If connection is slow, or server is slow, then it's not obvious, that the page has been changed.

akuleshov7:
Can be done with `index.html` page

akuleshov7:
This issue is related to all views where we need to wait until loading has finished

Issue #172, opened 2021-07-09
Frontend: Use spinner as loading indicator by default


petertrr:
Now we only use it in ProjectView. Can be implemented with [`Suspense`](https://reactjs.org/docs/concurrent-mode-suspense.html) in `App.kt`:
```kotlin
hashRouter {
    child(Suspense, props = jsObject{
        children =
    })
```

akuleshov7:
Also need to create some preloading page while save is loading

petertrr:
Implemented in #823, #830, #852 under #328

PR #772, opened 2022-05-17
FE table filters: If no resuts are found, display it explicitly

### LINKED ISSUE #484: FE table filters: If no resuts are found, display it explicitly
Now it just stays there with empty table and you can't be sure if there are no results or if there has been an error with fetching.

Cheshiriks:
* Added message if data in table is empty
Closes #484
screeeenк


petertrr (approved):
LGTM, but as I understand with this change users may see `No results found` for several moments while data is still being fetched. #172 and #328 should be proper ways to fix it.

Issue #783, opened 2022-05-20
Orchestrator: Support executing agents in Kubernetes


petertrr:
* [x] #787 Support reading some configuration properties in save-agent from environment
* [x] Refactor save-orchestrator to remove adjustments to container before its start (done in #787)
* Support logic for running a job with configurable number of replicas in `kubernetes` profile:
  * [x] #792 
  * [x] #795 
* [x] #839
* [x] #829
* [x] #1048
* [ ] Set resource limits for agents (kill processes that are starting to consume too much) - done in #1114

PR #787, opened 2022-05-20
Read agent id from env

petertrr:
* Update save-agent
* Refactoring in save-orchestrator

First part of #783. This change allows a more unified setup for Docker and Kubernetes, because orchestrator won't need to modify the container before starting it. All static data is copied into the image itself, and additional configuration of individual agents can be done via env.


save-orchestrator/src/main/kotlin/org/cqfn/save/orchestrator/service/DockerService.kt
```diff
@@ -237,6 +243,26 @@ class DockerService(private val configProperties: ConfigProperties,
             changeOwnerRecursively(resourcesPath, "cnb")
         }
 
+        val agentPropertiesFile = File(resourcesPath, "agent.properties")
```
kgevorkyan, line 246: May be extract this into something `fillAgentPropertiesFile()`?

petertrr: Yes, thanks, extracted

kgevorkyan:
If it is not difficult for you, could you please also fix my typo in `save-api` in this PR?

https://github.com/analysis-dev/save-cloud/blob/9e1a2d86028b21432384b8973fb3fd4279ec443a/save-api/src/main/kotlin/org/cqfn/save/api/utils/RequestUtils.kt#L150

It should be `shortFileInfo` variable, and not class name 🤦‍♂️

My current PR's seems far for merging and I don't want to open an new one, because of one line

kgevorkyan (approved):

PR #792, opened 2022-05-23
Orchestrator: Refactoring, preparation for deployment of save-agents into Kubernetes

### LINKED ISSUE #783: Orchestrator: Support executing agents in Kubernetes

petertrr:
I've split the original PR into two smaller ones: this one contains necessary refactorings, and the next one will introduce support for Kubernetes API in save-orchestrator.
* Refactor `ContainerManager`: introduce `AgentRunner` interface for all operations that interact with the underlying engine; create `DockerClient` as a bean, use it in both DockerService (for image preparation) and DockerAgentRunner (for everything else)
  * Orchestrator is now able to manage agents not only by the list of IDs, but by the ID of execution itself (i.e. all operations should be run on a batch of agents)
* Fix incorrect error handling in `AgentsController`: a publisher was being discarded with `doOnError`, unlike `onErrorResume`
* Change the way of reading env file in the agent start cmd

Part of #783


nulls (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/AgentRunner.kt
```diff
@@ -0,0 +1,56 @@
...
+     */
+    fun cleanup(executionId: Long)
+
+    /**
+     * Base on id of an execution load data about existing running agents for it.
+     * TODO: implement under https://github.com/saveourtool/save-cloud/issues/11
+     *
+     * @param executionId
+     */
+    fun discover(executionId: Long) {
```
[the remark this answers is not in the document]
petertrr: We still will need to load corresponding objects, because eventually we manage agents using our orchestrator and keep their states in our DB.

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -0,0 +1,196 @@
...
+    configProperties: ConfigProperties,
+    private val dockerClient: DockerClient,
+    private val meterRegistry: MeterRegistry,
+) : AgentRunner {
+    private val settings: DockerSettings = requireNotNull(configProperties.docker) {
+        "orchestrator.docker properties are missing but are required with current active profiles"
+    }
+
+    @Suppress("TYPE_ALIAS")
+    private val agentIdsByExecution: ConcurrentMap<Long, MutableList<String>> = ConcurrentHashMap()
```
akuleshov7, line 40: Can there really be a race?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -0,0 +1,196 @@
...
+     */
+    @Suppress("UnsafeCallOnNullableType", "TOO_LONG_FUNCTION")
+    private fun createContainerFromImage(baseImageId: String,
+                                         workingDir: String,
+                                         runCmd: String,
+                                         containerName: String,
+    ): String {
+        val baseImage = dockerClient.listImagesCmd().execTimed(meterRegistry, "$DOCKER_METRIC_PREFIX.image.list")!!.find {
+            // fixme: sometimes createImageCmd returns short id without prefix, sometimes full and with prefix.
+            it.id.replaceFirst("sha256:", "").startsWith(baseImageId.replaceFirst("sha256:", ""))
```
akuleshov7, line 125: why replace, if it is the same in both IDs: `sha256:`

save-orchestrator/src/test/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManagerTest.kt
```diff
@@ -47,47 +53,49 @@ class ContainerManagerTest {
...
+            agentRunCmd = "./script.sh",
+        ).single()
+        val inspectContainerResponse = dockerClient
             .inspectContainerCmd(testContainerId)
             .exec()
 
         Assertions.assertEquals("bash", inspectContainerResponse.path)
         Assertions.assertArrayEquals(
-            arrayOf("-c", "source .env && ./script.sh"),
+            arrayOf("-c", "env \$(cat .env | xargs) ./script.sh"),
```
akuleshov7, line 69: why do we need to pass arguments with `xargs` here

akuleshov7 (approved):
lgtm, but have small questions

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
petertrr: I don't really remember, it's another piece of old code. I think, we can get one id with this prefix and the other one without and I couldn't fully determine why. I think, it might behave differently on different versions of Docker. So if any of IDs has this prefix, we compare only stuff after it

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/AgentRunner.kt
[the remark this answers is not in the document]
akuleshov7: hm, I got it, ok

PR #793, opened 2022-05-24
save-api: Fixes for publication

petertrr:
* Update artifact group to `com.saveourtool.save`
* Configure publishing for save-cloud-common


kgevorkyan:
Thanks for fixing

I tried several experiments with adding of
```kotlin
publishing {
    publications {
...
```
into the root config, but it didn't give any visible results
btw, why without `configurePublishing()` in `common` publications are empty?

kgevorkyan:
Thanks for fixing

I tried several experiments with adding of
```kotlin
publishing {
    publications {
...
```
into the root config, but it didn't give any visible results
btw, why without `configurePublishing()` in `common` publications are empty?

kgevorkyan (approved):

petertrr:
> btw, why without `configurePublishing()` in `common` publications are empty?

Gradle's best practices suggest that a project shouldn't be configured from other projects, from root project too. So every project that needs to be published should have `maven-publish` applied and configured.

PR #794, opened 2022-05-24
Moving to saveourtool domain (omg)

akuleshov7:
### What's done:
- updated packages
- updated deploy scripts
- updated links on the FE


petertrr:
Maybe, we can do this in two iterations: updates related to GH organization (links and docker images) and renaming of packages?

Issue #11, opened 2021-02-20
Orchestrator: On startup discover already running agents


petertrr:
On startup, check in DB for running SAVE agents and discover corresponding docker containers.

petertrr:
Orchestrator can look for a specific pattern of container names, for example. But OTOH, running agents will start heartbeating into orchestrator, so after 15 seconds after it boots it should already be aware of all agents.
Another question that arises here is that we should ensure, that orchestartor host name doesn't change and it is still acessible by agents. Orchestrator should connect them to `save` docker network on creation.

petertrr:
Orchestrator should firstly retrieve agent ids from DB, then ensure they are actually running. FFor ids, that are not running, it should update their status. For agents that are running but not present in the DB - if we can't figure what they are doing (I believe joins in DB are the only way now), we should kill them.

PR #795, opened 2022-05-26
Support running save-agents as Kubernetes Job

### LINKED ISSUE #783: Orchestrator: Support executing agents in Kubernetes

petertrr:
Continuation of #792, parent issue is #783

* Orchestrator: Support interaction with Kubernetes API with Fabric8 client
* Add a dedicated `ServiceAccount` for orchestrator
* Deploy backend and orchestrator with fixed ClusterIP
* Refactoring:
  * Split buildAndCreateContaienrs into build and create
  * Refactor HeartbeatController: store info per execution
* Always upload SARIF reports for diktat, because otherwise there are a lot of stale conversations by github-code-scanning bot

### TODO (for separate PRs):
* rerun. In rerun mode we are currently performing a cleanup before starting execution once again. We remove containers and images used for the previous attempt, because we reuse same IDs and names. However, in Kubernetes there shouldn't be such problem: if we delete a Job, then we will be able to create a new one with the same name.
* ~Use ConfigMap instead of configuration via agent.properties~ - since this file should exist per execution (and even per agent, because we have `id` there), there will be too many ConfigMaps this way
* orchestrator should be able to load information about existing (running) agents (there is an issue for this, #11)
* `stop` operation should accepts `executionId` and not individual agent IDs. For k8s it naturally should delete a Job, and then all pods get stopped by the job controller. For now, deletion of a Job can be done with manual `/cleanup` operation.

NB: I've noticed several times that in Minukube the first request from backend to preprocessor (`Sending request to preprocessor (executionType GIT) to start save file for project id=1`) is silently lost: no error on backend, no handling on preprocessor.


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/KubernetesManager.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    @Suppress("TOO_LONG_FUNCTION")
+    override fun create(executionId: Long,
+                        baseImageId: String,
+                        replicas: Int,
+                        workingDir: String,
+                        agentRunCmd: String,
+    ): List<String> {
+        // fixme: pass image name instead of ID from the outside
+        val baseImage = dockerClient.listImagesCmd().execTimed(meterRegistry, "$DOCKER_METRIC_PREFIX.image.list")!!.find {
```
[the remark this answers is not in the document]
petertrr: Fixed, but can't be resolved automatically because of problems with SARIF uploading when there are no errors 😕

akuleshov7:
> NB: I've noticed several times that in Minukube the first request from backend to preprocessor (Sending request to preprocessor (executionType GIT) to start save file for project id=1) is silently lost: no error on backend, no handling on preprocessor.

Their internal bug?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/AgentsController.kt
```diff
@@ -107,6 +110,10 @@
     fun findAgentIdsForExecution(@RequestParam executionId: Long) = agentRepository.findByExecutionId(executionId)
         .map(Agent::containerId)
 
+    @GetMapping("/agents/{agentId}/execution/id")
+    fun findExecutionIdByAgentId(@PathVariable agentId: String) = agentRepository.findByContainerId(agentId)
```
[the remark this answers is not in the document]
akuleshov7: Who's the fuck are you?

petertrr:
> Their internal bug?

I haven't really looked much into it. In case it'll be reproducible on the real env, and not only in Minikube, we'll have to investigate. Or we can just configure retries for all WebClients that we are using - it won't hurt even with good network

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/ConfigProperties.kt
```diff
@@ -56,6 +58,17 @@ data class DockerSettings(
     val runtime: String = "runc",
 )
 
+/**
+ * @property apiServerUrl see https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/
```
akuleshov7, line 62: could you please add extra info here - very difficult to understand what are these properties for

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -88,10 +96,11 @@ class AgentsController(
     }
 
     /**
+     * @param executionId ID of execution for which agents should be stopped
      * @param agentIds list of IDs of agents to stop
      */
     @PostMapping("/stopAgents")
-    fun stopAgents(@RequestBody agentIds: List<String>) {
+    fun stopAgents(@RequestParam executionId: Long, @RequestBody agentIds: List<String>) {
```
akuleshov7, line 103: why do we need executionId here?

save-cloud-charts/save-cloud/values-minikube.yaml
```diff
@@ -3,12 +3,16 @@
 profile: dev
 backend:
   profile: dev,secure,kubernetes
+  clusterIP: 10.106.182.9
```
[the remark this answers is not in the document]
petertrr: The original idea was to assign fixed IPs for backend and orchestrator, put them into `agent.properties` and then the agent would use them to connect to these services. However, it shouldn't actually be needed anymore, because agents are run as Kubernetes pods and should be able to connect to other services using DNS. I'll change this part

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -67,22 +69,28 @@ class DockerAgentRunner(
...
         }
     }
 
-    override fun stopByAgentId(agentId: String) {
+    override fun stopByAgentId(agentId: String): Boolean {
         logger.info("Stopping agent with id=$agentId")
         val state = dockerClient.inspectContainerCmd(agentId).exec().state
-        if (state.running == true) {
-            dockerClient.stopContainerCmd(agentId).exec()
+        return if (state.running == true) {
```
akuleshov7, line 83: if (state.running) ?
Or it can be null?

If null - then we should log it properly:
`"Agent with id=$agentId was requested to be stopped, but it actual state=$state"`

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
petertrr: I wanted to move from stopping individual agents by ID to terminating the whole group of agents using executionId. Because, unlike previous setup with Docker, in Kubernetes we have an entity that acts as a grouping for agents - a Job. However, I've decided to postpone this change till later PRs. I'll remove this parameter for now

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
petertrr: Yep, it's nullable, because it's marked as `@CheckForNull` in Java code. However, [Docker API docs](https://docs.docker.com/engine/api/v1.41/#operation/ContainerInspect) don't specify when it can be absent. They suggest, however, to use `Status == running` instead of `running` boolean field.

akuleshov7 (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/KubernetesManager.kt
```diff
@@ -0,0 +1,147 @@
...
+            .withName(jobNameForExecution(executionId))
+            .get()
+        job?.let {
+            kc.batch().v1().jobs()
+                .withName(jobNameForExecution(executionId))
+                .delete()
+        }
+    }
+
+    private fun jobNameForExecution(executionId: Long) = "save-execution-$executionId"
```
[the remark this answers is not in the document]
petertrr: We name our containers `save-execution-$executionId-$index` now, so it's for consistency. Do you think we should rename?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/ConfigProperties.kt
petertrr: Added

Issue #141, opened 2021-06-22
Frontend: js file is too large


petertrr:
The `frontend.js` bundle is too large and it takes quite a lot of time to load.

```
> Task :save-frontend:browserProductionWebpack
Cache saved successfully
asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  save-frontend.js (3.17 MiB)
entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (3.17 MiB)
      save-frontend.js
webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

PR #796, opened 2022-05-26
Import individual font-awesome icons to reduce bundle size

petertrr:
Related discussion can be found at https://github.com/FortAwesome/react-fontawesome/issues/70.
We can't use babel to transform imports in existing code, because Kotlin (or Webpack?) hides them and provides already resolved modules to the entrypoint. So instead I'm using a marker interface for FA icons imported from individual modules.

This change leads to the following size reduction: 
- old: `save-frontend-vendors.js (1.79 MiB)`
- new `save-frontend-vendors.js (723 KiB)`

Related issue: #141


sanyavertolet (approved):
lgtm

PR #797, opened 2022-05-27
Use React error boundary

### LINKED ISSUE #613: Frontend: Add error boundary
https://reactjs.org/docs/error-boundaries.html
This should be a component that is shown when unhandled exception is thrown. Something like `Something went wrong, please try again or report on our github`.

petertrr:
* Add ErrorBoundary (https://reactjs.org/docs/error-boundaries.html)
* Add props for FallbackView

Closes #613 

image

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/ErrorBoundary.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/ErrorBoundary.kt
@@ -1,0 +1,69 @@
+/**
+ * Support for [error boundaries](https://reactjs.org/docs/error-boundaries.html)
+ */
+
+package com.saveourtool.save.frontend.components
+
+import com.saveourtool.save.frontend.components.views.FallbackView
+import com.saveourtool.save.frontend.topBarComponent
+
+import csstype.ClassName
+import react.Component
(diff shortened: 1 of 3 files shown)


akuleshov7:
WOW!!!!

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/ErrorBoundary.kt
```diff
@@ -0,0 +1,69 @@
...
+            hasError = false
+        }
+    }
+
+    override fun render(): ReactNode? = if (state.hasError == true) {
+        div.create {
+            className = ClassName("container-fluid")
+            topBarComponent()
+            FallbackView::class.react {
+                bigText = "Error"
```
akuleshov7, line 47: We do not pass an error number and error text here?

akuleshov7 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/ErrorBoundary.kt
petertrr: In this case the error is usually some sort of internal stack trace, so I just add a generic text

PR #799, opened 2022-05-27
Bugfix for wrong redirects

### LINKED ISSUE #784: Bug: wrong redirect when super admin confirms his rights in projectSettingsMenu

sanyavertolet:
This PR closes #784 

### What's done:
- [x] Fixed a bug with wrong redirect
- [x] Improved modals to have changeable button labels

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
@@ -131,6 +131,11 @@ external interface OrganizationViewState : State {
      * Users in organization
      */
     var usersInOrganization: List<UserInfo>?
+
+    /**
+     * Label that will be shown on Close button of modal windows
+     */
+    var closeButtonLabel: String?
 }
 
 /**
@@ -167,6 +172,7 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
         state.organization = Organization("", OrganizationStatus.CREATED, null, null, null)
         state.selectedMenu = OrganizationMenuBar.INFO
         state.projects = emptyArray()
+        state.closeButtonLabel = null
     }
 
     private fun deleteOrganization() {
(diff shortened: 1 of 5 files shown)


petertrr (commented):
What was the bug with redirect and how has it been fixed? It's not clear from code and there is no description in the issue

sanyavertolet:
> What was the bug with redirect and how has it been fixed? It's not clear from code and there is no description in the issue

Yes, I used wrong enum element processing notification. Now there is a pure notification that is shown to super admin.
Now I use errorModalWindow because it has only one button.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -184,10 +190,10 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
 
     private fun showNotification(notificationLabel: String, notificationMessage: String) {
         setState {
-            confirmationType = ConfirmationType.GLOBAL_ROLE_CONFIRM
-            isConfirmWindowOpen = true
-            confirmLabel = notificationLabel
-            confirmMessage = notificationMessage
+            isErrorOpen = true
```
petertrr, line 193: So basically you're now reusing an existing modal window not only for errors, but for confirmations too?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RBuilderUtils.kt
```diff
@@ -32,13 +31,15 @@ enum class ConfirmationType {
  * @param isErrorOpen flag to handle error
  * @param errorLabel label of error
  * @param errorMessage message of error
+ * @param closeButtonLabel label that will be shown on the only button
  * @param handler handler to close
  * @return modal
  */
 fun RBuilder.runErrorModal(
```
petertrr, line 38: These things can be refactored: converted to a `FC` with props containing all these settings, and renamed from `run...Modal` to something like `singleButtonModal` and `doubleButtonModal`. Can be done as a separate PR

petertrr (approved):

PR #811, opened 2022-06-02
Build and cache base images for executions

### LINKED ISSUE #344: Orchestrator: Cache base images for test execution

petertrr:
* Orchestrator: Support building base image (without additional test resources)
* Orchestrator: Support changing docker registry url via properties
* Backend: Add scheduled job to build base image for all supported SDKs
* Backend: Refactor scheduler to use spring-boot's autoconfiguration

Note for the future (about functional controllers):
`RouterFunctionDsl` doesn't play well with `springdoc`: they offer `SpringdocRouteBuilder` which can be used with `RouterFunctions` API, but `Operation` description should be provided manually.
Moreover, full path registered in `RouterFunctionDsl` is corrupted in actuator's `mappings` endpoint

Closes #344


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesController.kt
```diff
@@ -28,8 +29,10 @@ typealias ResponseListTestSuites = ResponseEntity<List<TestSuiteDto>>
 @RestController
 class TestSuitesController(
     private val testSuitesService: TestSuitesService,
-    private val scheduler: Scheduler,
 ) {
+    @Autowired(required = false)
+    private lateinit var quartzScheduler: Optional<Scheduler>
```
nulls, line 34: why it's optional now?

petertrr: I wanted to create this bean only in certain profiles, but reverted these changes eventually. I've changed it to be required here too.

nulls (approved):

nulls (approved):

PR #824, opened 2022-06-08
Feature/agent in same docker network

petertrr:
The idea of this change was to simplify setup of agents: now we are passing explicitly IP of host network so that agent can connect to backend's and orchestrator's exposed ports. If agents would be connected to the same docker network, this would be simplified as they could access upstream services by their DNS name. This change in its current state adds this option to docker compose.
Drawbacks:
* `COMPOSE_PROJECT_NAME` should be set explicitly - I couldn't find any reliable way to pass name of compose project to compose services (unlike Kubernetes, where a lot of useful info can be passed as environment variables)
* It works differently for Docker Swarm: there is a network which name defaults to swarm stack name.

So it would make sense to simply pass network name for agents as a property of save-orchestrator. It could be added into `~/configs/orchestrator/application.properties` and reference an existing network on a particular machine

### DIFF
--- a/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
+++ b/buildSrc/src/main/kotlin/org/cqfn/save/buildutils/DockerStackConfiguration.kt
@@ -164,6 +164,7 @@ fun Project.createStackDeployTask(profile: String) {
         require(componentName in allprojects.map { it.name }) { "Component name should be one of gradle subproject names" }
         val buildTask = project(componentName).tasks.named<BootBuildImage>("bootBuildImage")
         dependsOn(buildTask)
+        dependsOn("generateComposeFile")
         val serviceName = when (componentName) {
             "save-backend", "save-orchestrator", "save-preprocessor" -> "save_${componentName.substringAfter("save-")}"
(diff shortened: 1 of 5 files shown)


nulls (approved):

save-deploy/README.md
```diff
@@ -24,7 +24,7 @@ Deployment is performed on server via docker swarm or locally via docker-compose
 * Secrets should be added to the swarm as well as to `$HOME/secrets` file.
 * If custom SSL certificates are used, they should be installed on the server and added into JDK's truststore inside images. See section below for details.
 * Loki logging driver should be added to docker installation: [instruction](https://grafana.com/docs/loki/latest/clients/docker-driver/#installing)
-* Pull new changes to the server and run `./gradlew -Psave.profile=prod deployDockerStack`.
+* Pull new changes to the server and run `./gradlew -Psave.profile=prod -PdockerNetwork=save_default deployDockerStack`.
```
nulls, line 27: is it a suggestion or we should always provide -PdockerNetwork on prod deploying?

petertrr: We should always provide it, because network name is built from stack name (created with `docker stack deploy` command) and can't be deduced based on compose file location, as for docker-compose

PR #828, opened 2022-06-10
`useEffect`: pass dependencies as vararg everywhere

### LINKED ISSUE #827: `useEffect`: pass dependencies as vararg everywhere
For example, `useEffect(arrayOf(tableInstance.state.pageSize, pageCount))` should be `useEffect(tableInstance.state.pageSize, pageCount)`

Also there are some places where dependencies should be reconsidered, e.g. in TableComponent we set `pageCount` state from within effect, which uses it as a dependecny - seems incorrect

petertrr:
* pass dependencies as vararg everywhere
* Remove `pageCount` from `useEffect` in `TableComponent`

Closes #827

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
@@ -133,7 +133,7 @@ fun <D : Any> tableComponent(
         additionalOptions()
     }, plugins = plugins)
 
-    useEffect(arrayOf<dynamic>(tableInstance.state.pageSize, pageCount)) {
+    useEffect(tableInstance.state.pageSize) {
         if (useServerPaging) {
             scope.launch {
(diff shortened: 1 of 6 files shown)


sanyavertolet (approved):
finally we got rid of it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
```diff
@@ -133,7 +133,7 @@ fun <D : Any> tableComponent(
         additionalOptions()
     }, plugins = plugins)
 
-    useEffect(arrayOf<dynamic>(tableInstance.state.pageSize, pageCount)) {
+    useEffect(tableInstance.state.pageSize) {
```
nulls, line 136: pageCount is not needed?

petertrr: It created a circular dependency, i.e. we would query page count and update it whenever page count was changed. Also, page count depends on page size, so it will get updated whenever page size changes

PR #840, opened 2022-06-15
Rework agents stop

### LINKED ISSUE #829: Orchestrator: Rework `stop` and `cleanup` operations
* In rerun mode we are currently performing a cleanup before starting execution once again. We remove containers and images used for the previous attempt, because we reuse same IDs and names. However, in Kubernetes there shouldn't be such problem: if we delete a Job, then we will be able to create a new one with the same name. We may delete only images, because they'll be rebuilt anyway
* `stop` operation should accept `executionId` and not individual agent IDs. For k8s it naturally should delete a Job, and then all pods get stopped by the job controller (same, as can be done now with manual `/cleanup` operation).
* We may want to switch back to shutting agents down from inside the container, i.e. agent exits after receiving a certain heartbeat response from orchestrator. It will make our setup more suitable for the concept of Kubernetes Job. Now Job controller sometimes recreates deleted pods, even with `restartPolicy` set to `never`

petertrr:
* Terminate agents by using `TerminationResponse`
* Extract `HeartBeatInspector` into a separate file along with some logic
* Minor improvements in charts
  * Support setting `nodeName` for orchestrator and preprocessor
  * Don't use container ports; make all services accepts connections on port 80
* Quick fix for `Sdk#toSdk` method
* Tests
  * Add `kotest-assertions`
  * Fix `HeartbeatControllerTest.kt`

Closes #829 and establishes foundation for #843 (`AgentService#markExecutionBasedOnAgentStates`)


akuleshov7:
@icemachined please have a brief look on this

save-orchestrator/src/test/kotlin/com/saveourtool/save/orchestrator/controller/heartbeat/HeartbeatControllerTest.kt
```diff
@@ -191,26 +191,46 @@ class HeartbeatControllerTest {
             testSuite = null,
             mockAgentStatuses = true,
             {
+                // set agents status to TERMINATED
```
nulls, line 199: looks like leftovers

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -101,122 +94,76 @@ class HeartbeatController(private val agentService: AgentService,
...
-            agentService.markAgentsAndTestExecutionsCrashed(crashedAgentsList)
-            logger.warn("All agents are crashed, initialize shutdown sequence")
-            if (agentsLatestHeartBeatsMap.keys().toList() == crashedAgentsList.toList()) {
-                initiateShutdownSequence(crashedAgentsList.first(), areAllAgentsCrashed = true)
-            }
-        } else {
-            logger.warn("Crashed agents $crashedAgentsList are not stopped after stop command")
+    private fun handleIllegallyOnlineAgent(agentId: String, state: AgentState) {
+        logger.warn("Agent sent $state status, but should be offline in that case!")
+        if (agentId !in crashedAgentsList) {
```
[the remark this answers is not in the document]
petertrr: Changed, thanks

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/domain/Sdk.kt
```diff
@@ -52,8 +55,8 @@ fun String.toSdk(): Sdk {
     require(splitSdk.size == 2) { "Cant find correct sdk and version" }
     val (sdkType, sdkVersion) = splitSdk.run { this.first() to this.last() }
     return when (sdkType) {
-        Jdk.NAME -> Jdk(sdkVersion)
-        Python.NAME -> Python(sdkVersion)
+        Jdk.NAME, Jdk("-1").name -> Jdk(sdkVersion)
```
nulls, line 58: there is `Jdk.NAME` and `Python.NAME` -- can we use them? I mean that there are two values which mean similar option

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -101,122 +93,76 @@ class HeartbeatController(private val agentService: AgentService,
...
-            agentService.markAgentsAndTestExecutionsCrashed(crashedAgentsList)
-            logger.warn("All agents are crashed, initialize shutdown sequence")
-            if (agentsLatestHeartBeatsMap.keys().toList() == crashedAgentsList.toList()) {
-                initiateShutdownSequence(crashedAgentsList.first(), areAllAgentsCrashed = true)
-            }
-        } else {
-            logger.warn("Crashed agents $crashedAgentsList are not stopped after stop command")
+    private fun handleIllegallyOnlineAgent(agentId: String, state: AgentState) {
+        logger.warn("Agent id=$agentId sent $state status, but should be offline in that case!")
+        if (agentId !in crashedAgents) {
```
nulls, line 138: we can simplify checking -- Set guarantees that collection doesn't contain duplicate

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/domain/Sdk.kt
petertrr: Yeah, there are string values for two purposes: display name, which is currently `Companion.NAME`, and docker image name, which is in the property `name`. This is indeed confusing, but let's create a separate issue to refactor it, because this PR is already pretty large

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
```diff
@@ -128,50 +128,120 @@ class AgentService(
...
-                AgentStatusDto(LocalDateTime.now(), AgentState.CRASHED, agentId)
+    fun markExecutionBasedOnAgentStates(
+        executionId: Long,
+        agentIds: List<String>,
+    ): Mono<*> {
+        // all { STOPPED_BY_ORCH || TERMINATED } -> FINISHED
+        // all { CRASHED } -> ERROR; set all test executions to CRASHED
+        return webClientBackend
+            .get()
+            .uri("/getAgentsStatusesForSameExecution?agentId=${agentIds.first()}")
```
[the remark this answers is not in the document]
petertrr: Added a new method in backend to get statuses by a list of ids

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -74,24 +73,17 @@ class HeartbeatController(private val agentService: AgentService,
...
+                        handleFinishedAgent(heartbeat.agentId, isSavingSuccessful)
                     }
-                    AgentState.BUSY -> Mono.just(ContinueResponse)
-                    AgentState.BACKEND_FAILURE, AgentState.BACKEND_UNREACHABLE, AgentState.CLI_FAILED, AgentState.STOPPED_BY_ORCH -> Mono.just(WaitResponse)
-                    AgentState.CRASHED -> {
-                        logger.warn("Agent sent CRASHED status, but should be offline in that case!")
+                    BUSY -> Mono.just(ContinueResponse)
+                    BACKEND_FAILURE, BACKEND_UNREACHABLE, CLI_FAILED, STOPPED_BY_ORCH -> Mono.just(WaitResponse)
+                    CRASHED, TERMINATED -> {
+                        handleIllegallyOnlineAgent(heartbeat.agentId, heartbeat.state)
```
nulls, line 86: maybe wrap it by Mono?

nulls (approved):

PR #841, opened 2022-06-15
Fix Gradle build on kotlin 1.7.0

nulls:


petertrr (approved):

petertrr (changes requested):
Oh, sorry, could you please also update cache keys for konan in all github workflow files to contain `-1.7.0`?

buildSrc/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
-    `kotlin-dsl`
+    // workaround https://github.com/gradle/gradle/issues/16345
```
akuleshov7, line 2: btw - the issue is from the February... 
Very strange

akuleshov7 (approved):

buildSrc/build.gradle.kts
nulls: The issue is from February 2021. it's long live issue.
I think that a fix is possible only in Gradle 8.
But comment about `Kotlin 1.7.0` is from June 9th (release date :))

nulls:
> also maybe upgrade:
> 
> ```toml
> # kx.datetime 0.3.3 is built with Kotlin 1.7, so currently not compatible with our build
> kotlinx-datetime = { require = "0.3.2", rejects = "0.3.3" }
> ```

done

petertrr (approved):

PR #842, opened 2022-06-15
Github credentials should be set in Settings tab of ProjectView

### LINKED ISSUE #812: Github credentials should be set in Settings tab of ProjectView
If User has not provided or has provided invalid git credentials to private git repository - he should be able to add them to project (on settings bar).

Cheshiriks:
* Added url settings in ProjectView
Closes #812


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -699,6 +712,44 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
         }
     }
 
+    private fun updateGit() {
```
petertrr, line 715: Why do we need all this logic here and not inside settings view? It'll make code more isolated if you pass only the callback to update state of parent component instead of placing the entire logic of the request

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ProjectController.kt
```diff
@@ -165,6 +162,36 @@ class ProjectController(
...
+        git?.apply {
+            url = gitDto.url
+            username = gitDto.username
+            password = gitDto.password
+            branch = gitDto.branch
+        }
+
+        git?.let {
+            val saveGit = gitService.save(it)
+            log.info("Update git id = ${saveGit.id}")
```
petertrr, line 190: I think it should be `debug`

petertrr (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
```diff
@@ -172,6 +174,19 @@ fun projectSettingsMenu(
...
+                div("row d-flex align-items-center mt-2 mr-2 ml-2") {
+                    div("col-5 text-left") {
+                        +"User settings:"
+                    }
+                    div("col-7 row") {
+                        button(type = ButtonType.button, classes = "btn btn-sm btn-primary") {
+                            attrs.onClickFunction = {
+                                updateGit(true)
+                            }
+                            +"Add"
```
akuleshov7, line 186: Please rename it properly “Add” -> “Add github credentials”

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -567,6 +570,16 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
...
         }
 
+        runSettingGitWindow(
+            state.isOpenGitWindow,
+            gitDto,
+            changeFields = { type, element -> changeGitFields(type, element) },
+            handlerCancel = { setState { isOpenGitWindow = false } }
+        ) {
+            updateGit()
+            // setState { isOpenGitWindow = false }
```
akuleshov7, line 580: Why commented?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -699,6 +712,44 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
...
 
+    private fun updateGit() {
+        scope.launch {
+            val headers = Headers().also {
+                it.set("Accept", "application/json")
+                it.set("Content-Type", "application/json")
+            }
+
+            val git = GitDto(
+                fieldsGitMap[InputTypes.GIT_URL]?.trim() ?: gitDto!!.url,
```
akuleshov7, line 723: Please rename “fieldsGitMap” is a bad name. What did you mean? gitCredentials? gitInfo? fieldsWithGitInfo?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ProjectController.kt
```diff
@@ -165,6 +162,36 @@ class ProjectController(
...
+        }
+        return ResponseEntity.ok("Git updated successfully")
```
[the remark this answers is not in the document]
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
Cheshiriks: moved

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ProjectController.kt
```diff
@@ -165,6 +162,36 @@ class ProjectController(
...
+     * @param gitDto
+     * @param authentication
+     * @return response
+     */
+    @PostMapping("/update/git")
+    @Suppress("UnsafeCallOnNullableType")
+    fun updateGit(
+        @RequestParam projectId: Long,
+        @RequestBody gitDto: GitDto,
+        authentication: Authentication
```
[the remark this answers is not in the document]
Cheshiriks: Added

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ProjectController.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
Cheshiriks: Renamed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
Cheshiriks: Removed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
Cheshiriks: Renamed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
```diff
@@ -68,6 +73,21 @@ fun projectSettingsMenu(
     val projectPath = props.project.let { "${it.organization.name}/${it.name}" }
 
     val (wasConfirmationModalShown, setWasConfirmationModalShown) = useState(false)
+    val (isOpenGitWindow, setOpenGitWindow) = useState(false)
+    val (gitDto, setGitDto) = useState(props.gitInitDto)
```
petertrr, line 77: Doesn't having a separate copy of `git` object here break data synchronization with parent component? I think the more correct solution (if parent component requires `git`) is to pass in props both `git` and a callback to update its value in parent's state

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ProjectController.kt
```diff
@@ -165,6 +162,36 @@ class ProjectController(
         return ResponseEntity.ok(projectStatus.message)
     }
 
+    /**
+     * @param projectId
+     * @param gitDto
+     * @param authentication
+     * @return response
+     */
+    @PostMapping("/update/git")
```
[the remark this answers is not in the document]
petertrr: OK, that makes sense.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
```diff
@@ -66,6 +73,21 @@ fun projectSettingsMenu(
...
+
+    child(
+        runSettingGitWindow(
```
petertrr, line 80: Also, here you make the same error with functional components, that we have made multiple times already. You create a new instance of `fc` by calling the function on every render - in the past it used to lead to lost focus after every character input. The correct way is to create an `fc` once as a top-level property and then add is as a child. If you need to pass some fields that are only available inside the parent component (`projectSettingsView`) you should add the into `props` of the child component instead of passing as function parameters.

This way the `fc` will be created only once and then only re-rendered.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ProjectSettingsMenu.kt
Cheshiriks: Fixed

petertrr: Not exactly this - you've simply moved component creation from `child` function several lines up. It's still created inside the render logic (i.e. all logic in `fc` is called on every render, only hooks persist between those calls). It should be created outside of `fc`, e.g. as a top level property in this file

Cheshiriks: Yes. Changed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RBuilderUtils.kt
```diff
@@ -27,6 +36,36 @@ enum class ConfirmationType {
...
+
+    /**
+     * Lambda to close window
+     */
+    var handlerCancel: (Event) -> Unit
+
+    /**
+     * Lambda to set new git dto in state
+     */
+    var handler: (GitDto) -> Unit
```
petertrr, line 66: Maybe name it something like `onGitUpdate`?

petertrr (approved):

PR #856, opened 2022-06-27
path issues in local start on windows closes #855

### LINKED ISSUE #855: path issues in local start on windows
when starting on windows 
some paths may use use different separators,
also windows don't understand some symbols in paths so there is a problems whan starting on windows

icemachined:


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
```diff
@@ -325,10 +326,10 @@ class AgentService(
             }
             .map { (isStandardMode, testPaths) ->
                 val cliArgs = if (!isStandardMode) {
-                    suitesToArgs.values.first()
+                    FilenameUtils.separatorsToUnix(suitesToArgs.values.first())
```
petertrr, line 329: I can see that you are now saving tests in `TestService` with Unix separators. Do we still need conversion here? I believe these values are constructed using existing data from the database

petertrr (commented):
Should we also add validation in the class that these fields (paths in `Execution`, `Test` and `TestSuite`) contain only Unix separators? Or at least mention it in properties' KDocs?

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestDataFilesystemRepository.kt
```diff
@@ -95,6 +93,9 @@ class TestDataFilesystemRepository(configProperties: ConfigProperties,
      * @return path to file with additional data
      */
     internal fun getLocation(executionId: Long, testResultLocation: TestResultLocation) = with(testResultLocation) {
-        root / executionId.toString() / pluginName / testSuiteName / testLocation / "$testName-debug.json"
+        root / executionId.toString() / pluginName / sanitizePathName(testSuiteName) / testLocation / "$testName-debug.json"
+    }
+    fun sanitizePathName(name: String): String {
```
nulls: please split by new line
also, do we need this method as public?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestService.kt
```diff
@@ -59,7 +60,7 @@ class TestService(
                     val testSuiteStub = TestSuite(testRootPath = "N/A").apply {
                         id = testDto.testSuiteId
                     }
-                    Test(testDto.hash, testDto.filePath, testDto.pluginName, LocalDateTime.now(), testSuiteStub, testDto.tags.joinToString(";"))
+                    Test(testDto.hash, FilenameUtils.separatorsToUnix(testDto.filePath), testDto.pluginName, LocalDateTime.now(), testSuiteStub, testDto.tags.joinToString(";"))
```
nulls: what's about existed values in database? maybe need to ad liquibase migration script?

save-backend/src/main/resources/application-dev.properties
```diff
@@ -1,7 +1,9 @@
 spring.liquibase.enabled=false
 # `mysql` is resolved when running mysql in docker-compose, change to `datasource.dev.url` if running backend via `bootRun` or from IDE
 spring.datasource.url=jdbc:mysql://localhost:3306/save_cloud
+#spring.datasource.url=jdbc:mysql://mysql:3306/save_cloud
```
nulls, line 4: looks like as test leftovers

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
nulls: `suitesToArgs` is taken from `TestSuiteDto.testRootPath` -- maybe normalize separators on TestSuite level? 
I guess it can be normalize on saving (TestService and TestSuiteService)

nulls (changes requested):

nulls (approved):

PR #860, opened 2022-06-28
Support save-cli 0.4.0

nulls:
Support changes after: https://github.com/saveourtool/save-cli/pull/404

### What's done:
* removed reading save.properties -- testRoot cannot be set by this file

The old behavior: `testFiles` is `emptyList()` by default. `testRootDir` cannot be set by `save.properteis`, since we need to have `testRootDir` to read this property

### DIFF
--- a/save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
+++ b/save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
@@ -1,8 +1,6 @@
 package com.saveourtool.save.preprocessor.controllers
 
-import com.saveourtool.save.core.config.SaveProperties
 import com.saveourtool.save.core.config.TestConfig
-import com.saveourtool.save.core.config.defaultConfig
 import com.saveourtool.save.domain.FileInfo
 import com.saveourtool.save.entities.Execution
 import com.saveourtool.save.entities.ExecutionRequest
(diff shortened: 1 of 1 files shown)


save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -548,20 +546,8 @@ class DownloadProjectController(
 
     @Suppress("UnsafeCallOnNullableType")
     private fun getTestResourcesRootAbsolutePath(testRootPath: String,
-                                                 projectRootRelativePath: String): String {
-        // TODO: File should be provided without explicit naming of `save.propeties`
-        val propertiesFile = File(configProperties.repository, projectRootRelativePath)
-            .resolve("$testRootPath/save.properties")
-        val saveProperties: SaveProperties = if (propertiesFile.exists()) {
-            decodeFromPropertiesFile<SaveProperties>(propertiesFile)
```
nulls, line 556: `decodeFromPropertiesFile` is used only here -- maybe remove it?

petertrr: I believe we initially had the following logic: user can provide `save.properties` file with `testFiles` set to custom path. We would then evaluate their properties file and use either custom or default value as `testFiles`. Then we sort of simplified it in save-cli, making `save.properties` and top-level `save.toml` always reside in the same directory.

petertrr: And I believe now we don't need `save.properties` in preprocessor at all

PR #869, opened 2022-06-30
Agent: use `report-dir` parameter from save-cli

petertrr:
After https://github.com/saveourtool/save-cli/issues/404 save-cli started to write reports into `save-reports/save.out.json` because of `--report-dir` parameter, which has been previousely ignored. Now save-agent expects report from another location. This PR fixes this behavior.

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
@@ -167,15 +167,22 @@ class SaveAgent(internal val config: AgentConfiguration,
     }
 
     @Suppress("MagicNumber")
-    private fun runSave(cliArgs: String): ExecutionResult = ProcessBuilder(true, FileSystem.SYSTEM)
-        .exec(
-            config.cliCommand.let {
-                "$it $cliArgs"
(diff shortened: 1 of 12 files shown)


save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -167,15 +167,22 @@ class SaveAgent(internal val config: AgentConfiguration,
...
-            } + " --report-type json --result-output file --log all",
-            "",
-            config.logFilePath.toPath(),
-            1_000_000L
-        )
+    private fun runSave(cliArgs: String): ExecutionResult {
+        val fullCliCommand = buildString {
+            append(config.cliCommand)
+            append(" $cliArgs")
+            append(" --report-type json")
```
nulls, line 174: Maybe also move to SaveCliConfig?

nulls (approved):

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
petertrr: Ok, moved

PR #870, opened 2022-07-01
Contest enhancement

### LINKED ISSUE #250: Save should have the functionality for contests in the area of static analysis
This is more like a milestone.

nulls:
### What's done:
* Split loading\discovering tests and running of them
* Extract Github information from project entity to a dedicated one

It's part of #250


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/GithubLocation.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Entity class with github repository information
+ * @property organization
+ * @property repository
+ * @property token it can be null for public repositories
+ * @property branch a specific branch name with test, otherwise default branch (main or master)
+ * @property subDirectory a path to test suites in repository, otherwise root directory
+ */
+@Entity
+class GithubLocation(
```
petertrr, line 14: I think we should support multiple git providers from the very beginning. Maybe we can make something like `GitRemoteLocation` with implementations for different services? I think the only difference would be in how URL is constructed based on repo name and organization name

nulls: > I think we should support multiple git providers from the very beginning.
GithubLocation is extracted for this purpose on future, I thought about interface `Location`, which encapsulates location of tests (GIT, SVN and etc.). Now it's not required and I added only `GithubLocation`

> I think the only difference would be in how URL is constructed based on repo name and organization name
also credentials provider -- GitHub supports `token` without user\password.

db/v-2/tables/test_suites_source.xml
```diff
@@ -0,0 +1,28 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="create-test-suites-source-database" author="nulls">
+        <createTable tableName="test_suites_source">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="name" type="varchar(255)">
+                <constraints nullable="false" unique="true" uniqueConstraintName="uq_test_suites_source_name"/>
```
nulls, line 14: it should be unique per organization

PR #879, opened 2022-07-05
Fix dependencies of `useRequest` in `TestExecutionDetailsView`

### LINKED ISSUE #866: flood of dummy requests when opening additional info
when open additional info link, 
the following requests to backend from  ui performed in infinite loop

http://localhost:8081/api/v1/testExecutions?executionId=36&checkDebugInfo=true
http://localhost:8081/api/v1/testExecutions?executionId=36&checkDebugInfo=true
http://localhost:8081/api/v1/files/get-debug-info

borned by the following piece of code 

```
    // fixme: after https://github.com/saveourtool/save-cloud/issues/364 can be passed via history state to avoid requests
    useRequest(arrayOf(status, testResultDebugInfo, executionId, testResultLocation), isDeferred = false) {
        val testExecutionDtoResponse = post(
            "$apiUrl/testExecutions?executionId=$executionId&checkDebugInfo=true",
            Headers().apply {
                set("Content-Type", "application/json")
            },
            Json.encodeToString(testResultLocation),
        )
        if (testExecutionDtoResponse.ok) {
            val testResultDebugInfoResponse = getDebugInfoFor(testExecutionDtoResponse.decodeFromJsonString())
            if (testResultDebugInfoResponse.ok) {
                setTestResultDebugInfo(
                    testResultDebugInfoResponse.decodeFromJsonString<TestResultDebugInfo>()
                )
            } else {
                setStatus("Additional test info is not available (code ${testResultDebugInfoResponse.status})")
            }
        } else {
            setStatus("Additional test info is not available (code ${testExecutionDtoResponse.status})")
        }
    }()

```

petertrr:
We shouldn't have put local variables into array of dependencies for react's effect. `params` on the other hand, is an object managed by React and suited better to serve as a dependency.

Closes #866


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/TestExecutionDetailsView.kt
```diff
@@ -112,7 +112,7 @@ fun testExecutionDetailsView() = fc<Props> {
     val (testResultDebugInfo, setTestResultDebugInfo) = useState<TestResultDebugInfo?>(null)
 
     // fixme: after https://github.com/saveourtool/save-cloud/issues/364 can be passed via history state to avoid requests
-    useRequest(arrayOf(status, testResultDebugInfo, executionId, testResultLocation), isDeferred = false) {
+    useRequest(arrayOf(params), isDeferred = false) {
```
sanyavertolet, line 115: Why do we use `arrayOf` here? Far as I remember `dependencies` is `vararg` of `dynamic`

sanyavertolet: Oh ok

sanyavertolet (approved):

Issue #778, opened 2022-05-18
Orchestrator: Proper handling for SAVE-cli issues reported by SAVE-agent


akuleshov7:
1. Right now if the processing fails we simply mark agent execution as `CLI_FAILED` 
2. In this case orchestrator simply stops ALL agents (if all of them are `IDLE` or `CLI_FAILED`)
3. But tests are not marked as `FAILED` - they are still in RUNNING state

We need to do the following:
- Rename `CLI_FALED` to `INTERNAL_SAVE_CLI_ERROR` 
- Mark tests `FAILED` (fail all of them)
- Think of how to give extra information to the user of what happened and why Agents failed

akuleshov7:
There is already some work have been done in https://github.com/analysis-dev/save-cloud/pull/776

In this PR we fail execution. But we need to get the reason of failure and provide it for the user

akuleshov7:
https://github.com/saveourtool/save-cloud/pull/2635/files#r1341720749

PR #886, opened 2022-07-07
Orchestrator: Proper handling for SAVE-cli issues reported by SAVE-agent #778

icemachined:
What's done:
* add new field failReason in ExecutionUpdateDto
* in case of failure ExecutionUpdateDto is saved in debugInfo folder in execution root
* new endpoint created for ui /get-execution-info
* in case of debug-info is unavailable for the test then in tries to extract and show execution-info

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
@@ -158,9 +158,7 @@ class SaveAgent(internal val config: AgentConfiguration,
                 state.value = AgentState.CLI_FAILED
             } else {
                 handleSuccessfulExit().invokeOnCompletion { cause ->
-                    if (cause == null) {
-                        state.value = AgentState.FINISHED
-                    }
+                    state.value = if (cause == null) AgentState.FINISHED else AgentState.CLI_FAILED
                 }
             }
             else -> {
--- a/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
@@ -158,7 +158,7 @@ class SaveAgent(internal val config: AgentConfiguration,
                 state.value = AgentState.CLI_FAILED
             } else {
                 handleSuccessfulExit().invokeOnCompletion { cause ->
-                    state.value = if (cause == null) AgentState.FINISHED else AgentState.CLI_FAILED
+                    state.value = if (cause == null) AgentState.FINISHED else AgentState.CRASHED
                 }
             }
             else -> {
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
@@ -192,11 +192,7 @@ class DownloadFilesController(
     fun getDebugInfo(
         @RequestBody testExecutionDto: TestExecutionDto,
     ): String {
-        val agentContainerId = testExecutionDto.agentContainerId
-            ?: throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Request body should contain agentContainerId")
-        val execution = agentRepository.findByContainerId(agentContainerId)?.execution
-            ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Execution for agent $agentContainerId not found")
-        val executionId = execution.id!!
+        val executionId = getExecutionId(testExecutionDto)
         val testResultLocation = TestResultLocation.from(testExecutionDto)
         val debugInfoFile = testDataFilesystemRepository.getLocation(
             executionId,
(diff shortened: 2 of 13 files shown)


save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -158,9 +158,7 @@ class SaveAgent(internal val config: AgentConfiguration,
                 state.value = AgentState.CLI_FAILED
             } else {
                 handleSuccessfulExit().invokeOnCompletion { cause ->
-                    if (cause == null) {
-                        state.value = AgentState.FINISHED
-                    }
+                    state.value = if (cause == null) AgentState.FINISHED else AgentState.CLI_FAILED
```
petertrr, line 161: I think a more suitable status here would be `CRASHED` instead of `CLI_FAILED`, because at this point we know that save-cli exited with status 0, but save-agent failed to save some of the execution data inside `handleSuccessfulExit`.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
```diff
@@ -188,6 +190,12 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
                                 val trdi = getDebugInfoFor(te)
                                 if (trdi.ok) {
                                     cellProps.row.original.asDynamic().debugInfo = trdi.decodeFromJsonString<TestResultDebugInfo>()
+                                } else {
+                                    val trei = getExecutionInfoFor(te)
```
petertrr, line 194: This information should be displayed not only when debug info is unavailable, but also somewhere in the general description of the execution. I think it  would make more sense to see error text right away if tests didn't even begin.

petertrr (changes requested):

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
icemachined: fixed

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -364,11 +365,12 @@ class DownloadProjectController(
         .toBodilessEntity()
         .then(initializeAgents(this))
         .onErrorResume { ex ->
+            val failReason = "Error during preprocessing"
```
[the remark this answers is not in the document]
icemachined: fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
icemachined: fixed

petertrr (approved):
LGTM

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -210,6 +206,42 @@ class DownloadFilesController(
         }
     }
 
+    private fun getExecutionId(testExecutionDto: TestExecutionDto): Long {
+        testExecutionDto.executionId?.let { return it }
```
nulls, line 210: why `executionId` can be null for `testExecutionDto`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -210,6 +206,42 @@ class DownloadFilesController(
...
+        return execution?.id
+            ?: throw ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                "Execution for agent $agentContainerId not found"
+            )
+    }
+
+    /**
+     * @param testExecutionDto
+     * @return [Mono] with response
```
nulls, line 224: this method returns `String`, not `Mono<String>`

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -210,6 +206,42 @@ class DownloadFilesController(
...
+    @Suppress("ThrowsCount", "UnsafeCallOnNullableType")
+    @PostMapping(path = ["/api/$v1/files/get-execution-info"])
+    fun getExecutionInfo(
+        @RequestBody testExecutionDto: TestExecutionDto,
+    ): String {
+        logger.debug("Processing getExecutionInfo : $testExecutionDto")
+        val executionId = getExecutionId(testExecutionDto)
+        val executionInfoFile = testDataFilesystemRepository.getExecutionInfoFile(executionId).file
+        return if (executionInfoFile.exists()) {
+            val text = executionInfoFile.readText()
```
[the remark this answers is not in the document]
icemachined: 1. it was made like in getDebugInfo
2. file is quite small
3. we always parse the whole object
so this optimization would not be effective here

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
icemachined: because then we build debugInfo to TestExecutionDto( on SaveAgent it can't bind it to executionId yet because it doesn't have access to the database

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
icemachined: It is wrapped to mono (Mono.just(result)) by webflux in InvocableHandlerMethod.invoke

nulls: interesting, but here we return String. KDoc describes this method

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestDataFilesystemRepository.kt
```diff
@@ -4,28 +4,29 @@ import com.saveourtool.save.agent.TestExecutionDto
...
-import kotlin.io.path.name
 
 /**
  * A repository for storing additional data associated with test results
  */
 @Repository
 class TestDataFilesystemRepository(
     configProperties: ConfigProperties,
-    private val objectMapper: ObjectMapper,
+    private val objectMapper: ObjectMapper
```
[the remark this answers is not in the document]
nulls: I see that it's old logic

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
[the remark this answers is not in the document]
nulls: my comment is more about unification, not about optimization.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
nulls: can we add executionId to debugInfo then?

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestDataFilesystemRepository.kt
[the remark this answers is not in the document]
icemachined: this is separate task. #908 was created

PR #889, opened 2022-07-08
Standard test suites loading

nulls:
### What's done:
* fixed collecting of initialized tests for marking old test suites


save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
```diff
@@ -242,15 +241,17 @@ class DownloadProjectController(
                         log.error("Error to update test suite with url=$testSuiteUrl, path=${testSuiteRepoInfo.testSuitePaths}")
                     }
             }
+                .flatMap { Flux.fromIterable(it) }
```
petertrr, line 244: Will there be any difference between `flatMap { Flux.fromIterable(it) }` and `flatMapIterable { it }`?

petertrr (approved):

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectController.kt
nulls: I don't see any difference, will change to `flatMapIterable` , I guess it's more optimized for iterable

akuleshov7 (approved):

PR #891, opened 2022-07-08
Syntax highlight for public tests

sanyavertolet:
This PR is connected with #250 

### What's done:
- [x] Added RehypeHighlight plugin based on highlight.js
- [x] Added intellij-light.css
- [x] Added syntax highlight for public tests

This is how it looks like now:
screenshot


save-frontend/src/main/resources/highlight-css/intellij-light.css
```diff
@@ -0,0 +1,117 @@
+/*
+Intellij-light style (c) Pegasis <me@pegasis.site>
```
petertrr, line 2: Why do we need code, when it's loaded as `<link>`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
```diff
@@ -89,12 +89,17 @@ internal class ContestController(
...
             ?: return Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
-        return TestFilesRequest(test.toDto(), testSuite.testRootPath).let {
+        return TestFilesRequest(test.toDto(), testSuite.testRootPath).let<TestFilesRequest, Mono<TestFilesContent>> {
             preprocessorWebClient.post()
                 .uri("/tests/get-content")
                 .bodyValue(it)
                 .retrieve()
                 .bodyToMono()
+        }.map { testFilesContent ->
+            testFilesContent.apply {
```
petertrr, line 99: I think it's better to make fields immutable and call `copy` here

sanyavertolet: Yes, that is better

save-frontend/src/main/resources/highlight-css/intellij-light.css
sanyavertolet: Oh, will replace local css with css from cloudflare

petertrr (approved):

PR #897, opened 2022-07-11
Merge SARIF reports in Github Actions

petertrr:
* Use `ReportMergeTask` to merge SARIF reports

### DIFF
--- a/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DetektConfiguration.kt
+++ b/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DetektConfiguration.kt
@@ -9,6 +9,7 @@ import io.gitlab.arturbosch.detekt.DetektPlugin
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import io.gitlab.arturbosch.detekt.report.ReportMergeTask
 import org.gradle.api.Project
+import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DetektConfiguration.kt
```diff
@@ -7,23 +7,40 @@ package com.saveourtool.save.buildutils
...
         config = rootProject.files("detekt.yml")
         basePath = rootDir.canonicalPath
         buildUponDefaultConfig = true
     }
+    if (path == rootProject.path) {
+        tasks.register("mergeDetektReports", ReportMergeTask::class) {
+            output.set(buildDir.resolve("detekt-sarif-reports/detekt-merged.sarif"))
+        }
+    }
+    val reportMerge: TaskProvider<ReportMergeTask> = rootProject.tasks.named<ReportMergeTask>("mergeDetektReports") {
```
nulls, line 35: do we need to add these changes to all our project to have merged reports there?

nulls (approved):

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DetektConfiguration.kt
petertrr: Yes. One more reason to consider adding shared Gradle logic :)

PR #899, opened 2022-07-11
Say no to legacy react wrappers

sanyavertolet:
### What's done:
 - [x] Moved all `fc`s to `FC`s
 - [ ] Partly moved all `RComponent`s to `Component`s

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
@@ -10,23 +10,28 @@ import com.saveourtool.save.entities.*
 import com.saveourtool.save.frontend.components.basic.InputTypes
 import com.saveourtool.save.frontend.components.basic.inputTextFormRequired
 import com.saveourtool.save.frontend.utils.*
+import csstype.ClassName
 
 import org.w3c.dom.HTMLInputElement
 import org.w3c.dom.events.Event
 import org.w3c.fetch.Headers
-import react.Props
-import react.RBuilder
-import react.State
 import react.dom.*
-import react.setState
 
 import kotlinx.browser.window
 import kotlinx.coroutines.launch
 import kotlinx.datetime.LocalDateTime
 import kotlinx.datetime.Month
-import kotlinx.html.ButtonType
+import react.dom.html.ButtonType
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
+import react.*
+import react.dom.html.ReactHTML.button
+import react.dom.html.ReactHTML.div
+import react.dom.html.ReactHTML.form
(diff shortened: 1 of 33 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/modal/Modal.kt
```diff
@@ -25,6 +25,7 @@ private val defaultModalStyle = Styles(
 /**
  * @param handler builder for modal component
  */
+@Deprecated("Use non-legacy code")
```
petertrr, line 28: I believe in this PR we'll need to completely remove dependency on `-legacy` artifact, if it's possible :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/chart/PieChartBuilder.kt
```diff
@@ -14,15 +14,15 @@ import kotlin.random.Random
...
     data: Array<DataPieChart>,
-    handler: RHandler<PieChartProps> = {},
-) = child(PieChart::class) {
-    attrs.data = data
-    attrs.animate = false
-    attrs.segmentsShift = 0
-    attrs.viewBoxSize = intArrayOf(100, 100)
-    attrs.radius = 50
+    handler: ChildrenBuilder.(props: PieChartProps) -> Unit = {},
+) = PieChart::class.react.create {
```
petertrr, line 20: I don't think it should be `create`, but rather extension `ElemenType.invoke` from `ChildrenBuilder` should be used

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -127,30 +132,43 @@ class CreateOrganizationView : AbstractView<Props, OrganizationSaveViewState>(tr
...
+                                    height = 260.0
+                                    width = 260.0
                                 }
-                                form(classes = "needs-validation") {
+                                form {
+                                    className = ClassName("needs-validation")
                                     div {
                                         inputTextFormRequired(InputTypes.ORGANIZATION_NAME, state.isValidOrganizationName!!, "", "Organization name", true) {
-                                            changeFields(InputTypes.ORGANIZATION_NAME, it)
+                                            changeFields(InputTypes.ORGANIZATION_NAME, it as Event)
```
petertrr, line 164: What is the inferred type of `it` here?

petertrr (changes requested):
Since the proposed change is too large, I suggest to split it into two parts: right now we can merge `fc -> FC` migration and add `fc` into Detekt's forbidden list (`ForbiddenMethodCall`, `ForbiddenImport` and similar rules), and then we can migrate class components

sanyavertolet:
As far as this PR was split into two different (#907 and #913), it should be considered as done.

Issue #900, opened 2022-07-11
Extract public info from RUN tab


sanyavertolet:
Now we have some public info in project's `RUN` tab. We need to move it to `INFO` tab which is gonna be public also and make `RUN` tab private.

Also need to
- [x] Add project global score label
- [x] Improve visual part
- [x] Think of private projects participating in contests

Task to move `RUN` block from `ProjectView` to separate `FC` will be done in #904

sanyavertolet:
Done by #902

PR #902, opened 2022-07-11
Info tab in ProjectView

sanyavertolet:
This PR is a part of #900 

### What's done:
- [x] Added `INFO` tab in project info
- [x] Refactored `ProjectScoreCard` to make it more general
- [x] Disabled `RUN` tab for those who do not have enough permission

### TODO:
 * Add project global score label
 * Improve visual part
 * Move `RUN` block from `ProjectView` to separate `FC` 

### How it looks like:
Screenshot

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
@@ -8,7 +8,6 @@
 package com.saveourtool.save.backend.controllers
 
 import com.saveourtool.save.backend.service.LnkContestProjectService
-import com.saveourtool.save.entities.ContestDto
 import com.saveourtool.save.entities.ContestResult
 import com.saveourtool.save.v1
 
@@ -66,6 +65,12 @@ class LnkContestProjectController(
             it.toContestResult()
         }
 
+    /**
+     * @param organizationName
+     * @param projectName
+     * @param numberOfContests
+     * @return best [numberOfContests] contests of a project with name [projectName]
+     */
     @GetMapping("/{organizationName}/{projectName}/best")
     @PreAuthorize("permitAll()")
     fun getBestProjectsContests(
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
@@ -8,10 +8,12 @@
 package com.saveourtool.save.backend.controllers
 
 import com.saveourtool.save.backend.service.LnkContestProjectService
+import com.saveourtool.save.entities.ContestDto
 import com.saveourtool.save.entities.ContestResult
 import com.saveourtool.save.v1
 
 import org.springframework.http.HttpStatus
+import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
 import org.springframework.web.bind.annotation.*
 import org.springframework.web.server.ResponseStatusException
@@ -61,6 +63,19 @@ class LnkContestProjectController(
         lnkContestProjectService.getByContestName(contestName)
     )
         .map {
-            ContestResult(it.project.name, it.project.organization.name, it.score)
+            it.toContestResult()
+        }
+
+    @GetMapping("/{organizationName}/{projectName}/best")
+    @PreAuthorize("permitAll()")
+    fun getBestProjectsContests(
+        @PathVariable organizationName: String,
+        @PathVariable projectName: String,
+        @RequestParam(defaultValue = "4") numberOfContests: Int,
+    ): Flux<ContestResult> = Flux.fromIterable(
+        lnkContestProjectService.getBestContestsByProject(projectName, organizationName, numberOfContests)
+    )
+        .map {
+            it.toContestResult()
         }
 }
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
@@ -17,11 +17,11 @@ import kotlinx.js.jso
 /**
  * ProjectScoreCardProps component props
  */
-external interface ProjectScoreCardProps : Props {
+external interface ScoreCardProps : Props {
     /**
-     * Name of a current project
+     * Name of a current project or contest
      */
-    var projectName: String
+    var name: String
 
     /**
      * Score of a project in a contest
@@ -35,7 +35,7 @@ external interface ProjectScoreCardProps : Props {
  * @return ReactElement
  */
 @Suppress("TOO_LONG_FUNCTION", "LongMethod")
-fun projectScoreCard() = FC<ProjectScoreCardProps> { props ->
+fun scoreCard() = FC<ScoreCardProps> { props ->
     div {
         className = ClassName("card border-left-info shadow h-70 py-2")
         div {
@@ -82,7 +82,7 @@ fun projectScoreCard() = FC<ProjectScoreCardProps> { props ->
                                     alignItems = AlignItems.center
                                     alignSelf = AlignSelf.center
                                 }
-                                +props.projectName
+                                +props.name
                             }
                         }
                         div {
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
@@ -15,6 +15,11 @@ import react.dom.html.ReactHTML.h6
 import kotlinx.js.jso
 
 /**
+ * ReactElement that represents scorecard
+ */
+val scoreCard = scoreCard()
+
+/**
  * ProjectScoreCardProps component props
  */
 external interface ScoreCardProps : Props {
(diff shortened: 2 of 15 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
```diff
@@ -61,6 +62,25 @@ class LnkContestProjectController(
...
+        }
+
+    /**
+     * @param organizationName
+     * @param projectName
+     * @param numberOfContests
+     * @return best [numberOfContests] contests of a project with name [projectName]
+     */
+    @GetMapping("/{organizationName}/{projectName}/best")
+    @PreAuthorize("permitAll()")
```
petertrr, line 75: How will we handle results for private projects?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
```diff
@@ -17,11 +17,11 @@ import kotlinx.js.jso
 /**
  * ProjectScoreCardProps component props
  */
-external interface ProjectScoreCardProps : Props {
+external interface ScoreCardProps : Props {
     /**
-     * Name of a current project
+     * Name of a current project or contest
```
petertrr, line 22: > project or contest

Kind of confusing

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
```diff
@@ -35,7 +35,7 @@ external interface ProjectScoreCardProps : Props {
  * @return ReactElement
  */
 @Suppress("TOO_LONG_FUNCTION", "LongMethod")
-fun projectScoreCard() = FC<ProjectScoreCardProps> { props ->
+fun scoreCard() = FC<ScoreCardProps> { props ->
```
petertrr, line 38: Maybe we should start making these `val`s instead of `fun`s, especially if they don't accept parameters

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectInfoMenu.kt
```diff
@@ -0,0 +1,227 @@
...
+                            className = ClassName("col-md-4")
+                            figure {
+                                img {
+                                    className = ClassName("img-fluid")
+                                    src = user.avatar?.let { path ->
+                                        "/api/$v1/avatar$path"
+                                    }
+                                        ?: run {
+                                            "img/user.svg"
+                                        }
```
petertrr, line 155: We must've already added similar logic for avatars somewhere. Can it be extracted into a reusable function?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -559,7 +563,9 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
         div("row align-items-center justify-content-center") {
             nav("nav nav-tabs mb-4") {
                 ProjectMenuBar.values()
-                    .filter { it != ProjectMenuBar.SETTINGS || state.selfRole.isHigherOrEqualThan(Role.ADMIN) }
+                    .filter {
+                        (it != ProjectMenuBar.SETTINGS && it != ProjectMenuBar.RUN) || state.selfRole.isHigherOrEqualThan(Role.ADMIN)
+                    }
```
petertrr, line 568: Maybe, invert condition and use `filterNot`? I think it would be more clear that we forbid certain tabs for certain roles

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
sanyavertolet: You mean something like this?
```
val card = FC<Props> { ... } ()
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
sanyavertolet: The point is that this component is used either to display the results of a given project in some contest (in this way the name of a contest is shown) or the results in a given contest (in this way the name of a project is shown)

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: The point is that contest info is public for everybody no matter if it is private or public

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectInfoMenu.kt
sanyavertolet: Yes, will do it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
sanyavertolet: Oh ok

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
petertrr: Yes, then we won't need things like `val cardComponent = card()` before using it

sanyavertolet: Talking about cards, we cannot do it because cards still have arguments
But there are some other components that could do so

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
sanyavertolet: To my mind there is no difference
```
.filterNot {
    (it == ProjectMenuBar.RUN || it == ProjectMenuBar.SETTINGS) && !state.selfRole.isHigherOrEqualThan(Role.ADMIN)
}
```
and
```
.filter {
    (it != ProjectMenuBar.SETTINGS && it != ProjectMenuBar.RUN) || state.selfRole.isHigherOrEqualThan(Role.ADMIN)
}
```

petertrr: For me the first one is 'We are filtering out elements if it's RUN or SETTING tab and role doesn't grant enough permissions'. And for the latter it's 'We are keeping elements if they are neither SETTINGS nor RUN OR if role is high enough'. Idk, it's just personal preference probably, but for me the first variant seems more clear and explicit.

sanyavertolet: Ok, let it be so

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
```diff
@@ -35,7 +40,7 @@ external interface ScoreCardProps : Props {
  * @return ReactElement
  */
 @Suppress("TOO_LONG_FUNCTION", "LongMethod")
-fun scoreCard() = FC<ScoreCardProps> { props ->
+private fun scoreCard() = FC<ScoreCardProps> { props ->
```
petertrr, line 43:
```suggestion
val scoreCard = FC<ScoreCardProps> { props ->
```
I meant simply like this

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
petertrr: Feels really weird that we hide private projects from everyone, but they can be linked to public contests and be uncovered.. Need to think more carefully about it and create a warning screen for private projects entering a public contest. (in a separate issue, of course)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
petertrr: Yes, no need to create a separate comnponent, but maybe at least explain a bit more:
```suggestion
     * Name of a current project or contest (acts as a card header)
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
sanyavertolet: To my mind it might be convenient to keep private fun in order if we would like to add params later

petertrr: I think it's almost always more correct to pass all parameters as props, for older components we have parameters because of our lack of understanding of React back then :) I think we can safely assume that these new components can be extended using props if it'll be needed

sanyavertolet: Some things like style can be left as arguments (for example as it is in `cardComponent`) because it defines type of a component, I do not think it should be in props. For example we might want scoreCard component to have a different background or different shape. In this case it might be good to pass a param like `hasBg` or `isBoarded`. To my mind it should not be in props cause props seem to be more functional set of attributes.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: Mentioned in #900

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ScoreCard.kt
petertrr: Ok, that makes sense

petertrr (approved):

Issue #905, opened 2022-07-12
Get rid of `kotlin-react-legacy`


sanyavertolet:
We need to get rid of `kotlin-react-legacy` in order to use new `ChildrenBuilder` API.
This task should be done in different PRs:
- [x] Move from old-style `fc`s to new `FC`s
- [x] Get rid of `RComponent`s

PR #907, opened 2022-07-12
`fc`s to `FC`s

sanyavertolet:
This PR is a part of #905

### What's done:
- [ ] Completely got rid of `fc`s
- [ ] Added `fc` to ForbiddenMethodCall in detekt.yml (seems like it doesn't work because of type resolution)

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/FileUploader.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/FileUploader.kt
@@ -13,22 +13,32 @@ import com.saveourtool.save.frontend.utils.ConfirmationType
 import com.saveourtool.save.frontend.utils.toPrettyString
 import com.saveourtool.save.v1
 
+import csstype.ClassName
 import csstype.Width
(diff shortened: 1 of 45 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/FileUploader.kt
```diff
@@ -207,45 +282,6 @@ fun fileUploader(
...
 
-/**
- * A component for file icon that changes depending on executable flag
- *
- * @param fileInfo a [FileInfo] to base the icon on
- * @param onExecutableChange a handler that is invoked when icon is clicked
- * @return a functional component
- */
-@Suppress("TYPE_ALIAS")
-internal fun fileIconWithMode(fileInfo: FileInfo, onExecutableChange: (file: FileInfo, checked: Boolean) -> Unit) = fc<Props> {
```
akuleshov7, line 218: why deleted?

akuleshov7: ah, moved

petertrr (approved):

Issue #839, opened 2022-06-14
k8s: Support building images in distributed environment


petertrr:
Support sharing execution image among nodes (either push it into an existing registry or deploy a registry for this purpose only). Moreover, with current setup (using `hostDir` for downloaded resources), orchestrator and preprocessor have to be on the same node in order to successfully build base image for execution.

Possible solutions: use volume mount that can be shared among many pods on many nodes, then push everything to a registry

petertrr:
Closing in favor of direct download of test resources by agents

PR #911, opened 2022-07-13
Improvements for running in Kubernetes

petertrr:
* Set RuntimeClassName only if `useGvisor` flag is set
* Run save-execution as non-privileged user
* Set deployment strategy for backend
* Run save-agent on the same node as orchestrator :( - required unitl #839 
* Add profile for running orchestrator in minikube


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -243,20 +243,17 @@ class DockerService(private val configProperties: ConfigProperties,
             log.info("Base image [$baseImage] for execution ${execution.id} doesn't exists, will build it first")
             buildBaseImage(sdk)
         }
-        val aptCmd = "apt-get ${configProperties.aptExtraFlags}"
         val imageId = dockerContainerManager.buildImageWithResources(
             baseImage = baseImage,
             imageName = imageName(execution.id!!),
             baseDir = resourcesPath,
             resourcesTargetPath = executionDir,
-            runCmd = """RUN $aptCmd update && env DEBIAN_FRONTEND="noninteractive" $aptCmd install -y \
```
kgevorkyan, line 252: Could you please describe the reason of changing the order of calling these `RUN` commands? now they will be called only at the end of function

petertrr: I've moved them into base image build only, because these commands are required for every execution. I.e. from logs of `DockerServiceTest`, first build is 
```
13:47:41.345 [Test worker] DEBUG c.s.s.o.d.DockerContainerManager - Using generated Dockerfile FROM docker.io/library/ubuntu:latest
RUN apt-get  update && env DEBIAN_FRONTEND="noninteractive" apt-get  install -y \
libcurl4-openssl-dev tzdata
RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
RUN rm -rf /var/lib/apt/lists/*

RUN useradd --create-home --shell /bin/sh save-agent
WORKDIR /home/save-agent/save-execution
```
and the second is 
```
13:48:01.049 [Test worker] DEBUG c.s.s.o.d.DockerContainerManager - Using generated Dockerfile FROM docker.io/library/save-base-ubuntu:latest

COPY resources /home/save-agent/save-execution
RUN chmod +x /home/save-agent/save-execution/save-agent.kexe
RUN chmod +x /home/save-agent/save-execution/save-0.3.2-linuxX64.kexe
RUN chown -R save-agent .
USER save-agent
```

kgevorkyan (approved):

PR #913, opened 2022-07-13
RComponent to CComponent

### LINKED ISSUE #905: Get rid of `kotlin-react-legacy`

sanyavertolet:
This PR closes #905 

### What's done:
- [ ] Added CComponent which is ChildrenBuilder-based component
- [ ] Completely removed legacy wrappers from dependencies


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/CComponent.kt
```diff
@@ -0,0 +1,61 @@
...
+        state = jso { init(props) }
+    }
+
+    @Suppress(
+        "WRONG_OVERLOADING_FUNCTION_ARGUMENTS",
+        "EMPTY_BLOCK_STRUCTURE_ERROR",
+        "MISSING_KDOC_CLASS_ELEMENTS",
+        "MISSING_KDOC_ON_FUNCTION",
+    )
+    open fun S.init() {}
```
petertrr, line 29: We didn't ever use similar method of `RComponent`, I think we should either remove it or rewrite state initializations from `init` to `override fun init`

petertrr (approved):
Looks OK, I guess. We should also remember all our architectural flaws that we've realized during this migration :)

PR #926, opened 2022-07-15
Fix flaky frontend tests

petertrr:
Kotlin generates JS tests without correct handling of `@BeforeTest` returning `Promise`. For now it looks like its better to keep creation and stopping of MSW worker inside each test.


sanyavertolet (approved):
lol, it worked!

PR #928, opened 2022-07-15
Fix for single testRootPath

nulls:
### What's done:
- added distinct()


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -157,6 +157,7 @@ class AgentsController(
             .bodyToMono<List<String>>()
             .toFlux()
             .flatMap { Flux.fromIterable(it) }
```
petertrr, line 159: `.bodyToFlux`?
```suggestion
            .bodyToFlux<String>()
```

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
nulls: thanks, done

PR #935, opened 2022-07-18
Redirect missing 404 requests to a proper 404 page on the frontend

### LINKED ISSUE #912: Redirect missing 404 requests to a proper 404 page on the frontend
If you go to saveourtool.com/#/hui you should be redirected to 404 page.
Same with projects, with executions and ALL pages that are missing.

Cheshiriks:
* Added redirect for 404 requests
Closes #912


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
```diff
@@ -46,18 +47,30 @@ val requestModalHandler: FC<PropsWithChildren> = FC { props ->
...
                 errorLabel = response?.status.toString(),
+                status = response?.status,
             )
         }
         setModalState(newModalState)
     }
 
     modal { modalProps ->
+        if (modalState.status == 404.toShort()) {
+            window.location.href = "${window.location.origin}/error"
```
petertrr, line 72: We shouldn't redirect to the generic fallback page. We should clearly indicate that it's a 404 Not Found situation

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
```diff
@@ -46,18 +47,30 @@ val requestModalHandler: FC<PropsWithChildren> = FC { props ->
...
                 isErrorModalOpen = response != null,
                 errorMessage = "${response?.status} ${response?.statusText}",
                 errorLabel = response?.status.toString(),
+                status = response?.status,
             )
         }
         setModalState(newModalState)
     }
 
     modal { modalProps ->
```
petertrr, line 70: I had in mind something more like
```kotlin
if (modalState.status == 404.toShort()) {
    Redirect { to = ... }
} else {
    modal {}
}
```
For me the current version (when location is changed inside rendering of `modal`) looks like an unexpected side effect

petertrr (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
```diff
@@ -161,28 +161,28 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
                                 }
 
                                 Route {
-                                    path = "/:user/settings/profile"
+                                    path = "/${state.userInfo?.name}/settings/profile"
```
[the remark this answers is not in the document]
petertrr: Ok, makes sense. However, does that mean that the page `/null/settings/profile` will be present for anonymous users?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
Cheshiriks: Yes, agree

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
Cheshiriks: Changed it

petertrr (approved):

Issue #803, opened 2022-06-01
Rework modal windows


sanyavertolet:
We need to refactor existing modals:

> These things can be refactored: converted to a `FC` with props containing all these settings, and renamed from `run...Modal` to something like `singleButtonModal` and `doubleButtonModal`. Can be done as a separate PR

_Originally posted by @petertrr in https://github.com/saveourtool/save-cloud/pull/799#discussion_r884958740_

akuleshov7:
not critical, can be done later

sanyavertolet:
Now there are two variants of how I see it:
1. We should try using browser windows like `window.alert()` or `window.confirm()` - this is a nice option, but the only problem is that `Chrome` does it not that good
2. We should implement one `modalBuilder` that uses bootstrap styles with custom button amount, generated by `buttonBuilder`

For now I like the second idea more.

PR #936, opened 2022-07-18
Contest registrations on frontend

sanyavertolet:
This PR is connected with #250

### What's done:
- [x] Added several endpoints to fetch contests/projects avaliable for registration
- [x] Added `contestEnrollerComponent` that is used for selecting contest or project in order to register for a contest
- [x] Added modal with `contestEnrollerComponent`
- [x] Added modal with `contestEnrollerComponent` to `ProjectView` and `ContestListView`

### How it looks:
Screenshot
Screenshot
Screenshot

### UPD:
Screenshot


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
```diff
@@ -83,6 +82,49 @@ class LnkContestProjectController(
             lnkContestProject.toContestResult(score)
         }
 
+    /**
+     * @param contestName
+     * @param authentication
+     * @return list of user's [Project]s that can participate in contest with name [contestName]
+     */
+    @GetMapping("/{contestName}/avaliable")
```
petertrr, line 90:
```suggestion
    @GetMapping("/{contestName}/eligible-projects")
```
Maybe something like this? At first I thought that this endpoint would return Boolean whether the caller can participate in the contest

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
```diff
@@ -0,0 +1,197 @@
...
+    "TOO_LONG_FUNCTION",
+    "MAGIC_NUMBER",
+    "ComplexMethod",
+)
+private fun contestEnrollerComponent() = FC<ContestEnrollerProps> { props ->
+    val (organizationName, setOrganizationName) = useState(props.organizationName)
+    val (projectName, setProjectName) = useState(props.projectName)
+    val (contestName, setContestName) = useState(props.contestName)
+
+    val (availabilityUrl, _) = useState(contestName?.let {
```
petertrr, line 121: Does it need to be in state?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
```diff
@@ -0,0 +1,197 @@
...
+        )
+            .text()
+            .await()
+        props.onResponseChanged(responseFromBackend)
+    }
+
+    useEffect(selectedOption) {
+        if (selectedOption.isNotBlank()) {
+            val splitResult = selectedOption.split("/")
+            if (splitResult.size == 1) {
```
petertrr, line 155: Too unclear and error-prone. I'd rather mark this more explicitly, like flag that is set at the start of render method, `isContest` or something like this. Or even make it a prop and set when creating the component

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -628,6 +628,15 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
                     testingType = state.testingType
                     isSubmitButtonPressed = state.isSubmitButtonPressed
                     gitDto = gitDto
+                    projectName = props.name
+                    organizationName = props.owner
+                    onContestEnrollerResponse = {
+                        setState {
+                            isErrorOpen = true
+                            errorMessage = it
+                            errorLabel = "Contest enrollment"
```
petertrr, line 637: Why are you opening an error modal?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestListView.kt
```diff
@@ -68,18 +93,63 @@ class ContestListView : AbstractView<ContestListViewProps, State>(false) {
...
+                    td {
+                        button {
+                            className = ClassName("btn btn-primary")
+                            onClick = {
+                                setState {
+                                    selectedContestName = cellProps.value
+                                    isProjectSelectorModalOpen = true
+                                }
+                            }
+                            +"Participate in ${cellProps.value}"
```
petertrr, line 107: Isn't it too much text? Why not just `Participate`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: Seems like yes because if it is not in state but is just a val, everything breaks after the first render

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: oh ok let it be so

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: Well it is a bit unclear, but to my mind another flag will make it even more complicated because we will need to check if all the data initialized in a correct way

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
sanyavertolet: Because this is a modal with the only button
#803 :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestListView.kt
sanyavertolet: Oh it seemed like more clear
Ok, will fix it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: Ok, have an idea, will fix it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: Done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: Done

petertrr: Have you considered `sealed class` as a props field, that will be set to one of two variants depending on what we want to render?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
petertrr: OK

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: Well, it seems to be a nice idea from a first glance but you see raise of complexity while implementing

sanyavertolet: Fixed )

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
```diff
@@ -114,11 +117,23 @@ fun ChildrenBuilder.showContestEnrollerModal(
     "ComplexMethod",
 )
 private fun contestEnrollerComponent() = FC<ContestEnrollerProps> { props ->
-    val (organizationName, setOrganizationName) = useState(props.organizationName)
-    val (projectName, setProjectName) = useState(props.projectName)
-    val (contestName, setContestName) = useState(props.contestName)
-    val (isContestSelector, _) = useState(props.contestName == null)
-    require(isContestSelector && props.organizationName != null && props.projectName != null || !isContestSelector && props.contestName != null)
+    val (isContestSelector, _) = useState(props.nameProps is ProjectNameProps)
```
petertrr, line 120: Now you can use `when` with smart casts instead of boolean flag and `as`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
```diff
@@ -144,6 +144,12 @@ class LnkContestProjectController(
         Mono.justOrEmpty<Project>(projectService.findByNameAndOrganizationName(projectName, organizationName)),
```
petertrr, line 144: `findWithPermission` and `Permission.READ`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
```diff
@@ -144,6 +144,12 @@ class LnkContestProjectController(
...
         Mono.justOrEmpty(contestService.findByName(contestName)),
     )
+        .switchIfEmpty {
+            Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
+        }
+        .filter { (project, _) ->
+            project.public
+        }
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
```
petertrr, line 154: If you use `findWithPermission` in the start of the method, that by this point we can assume that the user has at least read access to the project and we can send a more meaningful response, e.g. `409 Conflict: Private projects cannot be enrolled in the contest`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ContestEnroller.kt
sanyavertolet: We cannot do it with `var`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: Ok, done

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
petertrr: Yeah, but why at first check `READ` and then filter unless `WRITE`?

sanyavertolet: In order to provide better error code and message

petertrr (approved):

PR #938, opened 2022-07-20
Fix setting default values for inputs on ProjectView

petertrr:
Old logic that was relying on class property values and that used state values right after `setState` must've got broken after update to React 18 (haven't tested with 17 though).

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
@@ -342,18 +342,22 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
                 jsonProject,
                 loadingHandler = ::noopLoadingHandler,
             ).decodeFromJsonString()
+            gitDto = gitDtoInit
             val currentUserRole: Role = get(
                 "$apiUrl/projects/${state.project.organization.name}/${state.project.name}/users/roles",
                 headers,
                 loadingHandler = ::classLoadingHandler,
             ).decodeFromJsonString()
             setState {
-                gitDto = gitDtoInit
                 selfRole = getHighestRole(currentUserRole, props.currentUserInfo?.globalRole)
             }
             when {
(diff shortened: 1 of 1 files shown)


sanyavertolet:
In @Arrgentum's PR he tried to use state's value right after `setState` and the state did't change. According to React 18 release note, now states change with batches.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -342,18 +342,22 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
                 jsonProject,
                 loadingHandler = ::noopLoadingHandler,
             ).decodeFromJsonString()
+            gitDto = gitDtoInit
```
sanyavertolet, line 345: Why do we have `gitDto` as a class property instead of state variable?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -342,18 +342,22 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
...
                 headers,
                 loadingHandler = ::classLoadingHandler,
             ).decodeFromJsonString()
             setState {
-                gitDto = gitDtoInit
                 selfRole = getHighestRole(currentUserRole, props.currentUserInfo?.globalRole)
             }
             when {
-                state.gitUrlFromInputField.isBlank() && gitDto?.url != null -> state.gitUrlFromInputField = gitDto?.url ?: ""
-                state.gitBranchOrCommitFromInputField.isBlank() && gitDto?.branch != null -> state.gitBranchOrCommitFromInputField = gitDto?.branch ?: ""
```
sanyavertolet, line 356: lol and it worked?

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
petertrr: `¯\_(ツ)_/¯`  Horrible legacy. Probably we'll refactor ProjectView once

PR #940, opened 2022-07-20
Fix for testRootPath calculating

nulls:
### What's done:
- replaced from Flux<String> to Mono<List<String>>


sanyavertolet (approved):

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -153,9 +154,8 @@ class AgentsController(
             .contentType(MediaType.APPLICATION_JSON)
             .body(BodyInserters.fromValue(execution))
             .retrieve()
-            .bodyToFlux<String>()
-            .distinct()
-            .single()
+            .bodyToMono<List<String>>()
```
akuleshov7, line 157: it won't work

nulls: it's previous logic

PR #944, opened 2022-07-20
Automate updates of OpeanAPI spec backend-api-docs.json

kgevorkyan:
### What's done:
* Automate updates of OpeanAPI spec backend-api-docs.json


save-backend/build.gradle.kts
```diff
@@ -8,13 +8,18 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
     id("org.springdoc.openapi-gradle-plugin") version "1.3.4"
 }
 
 openApi {
     apiDocsUrl.set("http://localhost:5800/internal/v3/api-docs/latest")
+    outputDir.set(file(rootDir))
     outputFileName.set("$rootDir/save-backend/backend-api-docs.json")
+    waitTimeInSeconds.set(120)
+
+    val bootRun = project.tasks["bootRun"] as org.springframework.boot.gradle.tasks.run.BootRun
```
petertrr, line 21: Shouldn't it be `forkedBootRun`? Also, it's preferrable to use `tasks.named` instead of `tasks.get`

save-backend/build.gradle.kts
```diff
@@ -8,13 +8,18 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
     kotlin("jvm")
     // this plugin will generate generateOpenApiDocs task
-    // running this task, it will writes the OpenAPI spec into a backend-api-docs.json file in save-backend dir.
+    // running this task, it will write the OpenAPI spec into a backend-api-docs.json file in save-backend dir.
     id("org.springdoc.openapi-gradle-plugin") version "1.3.4"
 }
 
 openApi {
     apiDocsUrl.set("http://localhost:5800/internal/v3/api-docs/latest")
+    outputDir.set(file(rootDir))
```
petertrr, line 17:
```suggestion
    outputDir.set(file(projectDir))
```
Did you try it like this?

.github/workflows/backend-api-spec-update.yml
```diff
@@ -0,0 +1,42 @@
+name: Update backend api spec generated by Kotlin Gradle Plugin
+
+on:
+  pull_request:
+    paths:
+      - 'save-backend/**/*Controller'
```
petertrr, line 6: Should it be `*Controller.kt` or `*Controller*` or does it match this as a substring?

.github/workflows/backend-api-spec-update.yml
```diff
@@ -0,0 +1,42 @@
...
+  update_backend_api_spec:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Checkout source branch directly, without merging base into it
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Set up Java
+        uses: actions/setup-java@v1
```
petertrr, line 18: Why `v1`?

kgevorkyan: Changed to `v3`

.github/workflows/backend-api-spec-update.yml
kgevorkyan: Fixed

save-backend/build.gradle.kts
kgevorkyan: I tried, but for some reason, this didn't worked, it writes, that there is no any `backend-api-docs.json` file, even if it exist and failed with this error

save-backend/build.gradle.kts
kgevorkyan: In current version of plugin there is no `forkedProperties`, the readme of this plugin is outdated

Changed to `tasks.named`

.github/workflows/backend-api-spec-update.yml
petertrr: I believe `v3` requires `distribution` parameter, please compare to other workflows

save-backend/build.gradle.kts
petertrr: Should be `project.tasks`, not `rootProject.tasks`

.github/workflows/backend-api-spec-update.yml
kgevorkyan: 🤦‍♂️ Thanks

save-backend/build.gradle.kts
kgevorkyan: Yeah, thanks

save-backend/build.gradle.kts
petertrr: What about combination of `outputDir.set(file(projectDir))` and `outputFileName.set("backend-api-docs.json")`? Since the plugin generates full file name relative to `outputDir` (as we've seen from source) there is no need to set `outputFileName` as an absolute path

kgevorkyan: Seems, that it's works

petertrr (approved):

PR #948, opened 2022-07-20
[Gradle] Move precompiled script plugins to an included build

petertrr:
When changes are made to code in buildSrc, Gradle re-evaluates the whole build (https://docs.gradle.org/7.5.1/userguide/composite_builds.html, https://docs.gradle.org/7.2/userguide/organizing_gradle_projects.html#sec:build_sources). Putting precompiled plugins in an included build is a new recommedned practice

- `*.gradle.kts` moved from `buildSrc` to `gradle/plugins`
- Fixed CLI parameter in workflows
- `VersioningConfiguration` is moved and converted. Other files can be moved later
- Switched back on K/N cache as it seems that it works now


.run/Frontend-LocalRun.run.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="save-frontend:run --continuous" type="GradleRunConfiguration" factoryName="Gradle">
+  <configuration default="false" name="Frontend-LocalRun" type="GradleRunConfiguration" factoryName="Gradle">
```
nulls, line 2: maybe create a runner for mysql and select JRE-11 there?

petertrr: I guess it was a change from a merge commit 🤔

nulls (approved):

PR #951, opened 2022-07-21
Minor improvements in logging

petertrr:


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/TestController.kt
```diff
@@ -31,7 +32,8 @@ class TestController(
      */
     @PostMapping("/initializeTests")
     fun initializeTests(@RequestBody testDtos: List<TestDto>) {
-        log.debug { "Received the following tests for initialization: $testDtos" }
+        log.debug { "Received ${testDtos.size} tests for initialization" }
+        log.trace { "Received the following tests for initialization: $testDtos" }
```
kgevorkyan, line 36: Finally

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
```diff
@@ -114,7 +115,7 @@ class TestSuitesService(
                 testSuiteDto.testRootPath,
                 testSuiteDto.testSuiteRepoUrl,
             )
-            log.info("Mark test suite ${testSuite.name} with id ${testSuite.id} as obsolete")
```
kgevorkyan, line 117: It's should be kind of rare and scheduled task, does this message anyway annoying?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
petertrr: Ok, will change back to `info`

PR #953, opened 2022-07-21
Run executions on mac

sanyavertolet:
### What's done:
 * Specified base agent docker image architecture to be `linux/amd64`
 * Updated `mac` profile for `save-orchestrator`
 * Updated check in dependencies from `save-orchestrator` (now it checks that current platform is NOT linux)
 * Improved `ContestEnroller` on frontend - now you cannot participate in stub contest
 * Updated `CONTRIBUTING.md` and `save-deploy/README.md`
 * Added shell script for docker tcp listening on Mac

### DIFF
--- a/save-deploy/README.md
+++ b/save-deploy/README.md
@@ -78,12 +78,17 @@ If you run on windows, dependency `save-agent` is omitted because of problems wi
 To run on windows you need to compile save-agent on wsl and put saveAgentDistroFilepath to %USERPROFILE%\.gradle\gradle.properties <br/>
 For example: <br/> 
 `saveAgentDistroFilepath=file:\\\\\\\\wsl$\\Ubuntu\\home\\username\\projects\\save-cloud\\save-agent\\build\\libs\\save-agent-0.3.0-alpha.0.48+1c1fd41-distribution.jar` <br/>
-If you need to test changes in save-cli also you can compile snapshot version of save-cli on wsl <br/> 
+If you need to test changes in save-cli also you can compile snapshot version of save-cli on wsl <br/>
 and set saveCliPath and saveCliVersion in %USERPROFILE%\.gradle\gradle.properties <br/>
 For example:<br/>
 `saveCliPath=file:\\\\\\\\wsl$\\Ubuntu\\home\\username\\projects\\save-cli\\save-cli\\build\\bin\\linuxX64\\releaseExecutable` <br/> 
 `saveCliVersion=0.4.0-alpha.0.42+78a24a8` <br/>
 the version corresponds to the file `save-0.4.0-alpha.0.42+78a24a8-linuxX64.kexe` <br/>
+
+#### Some workarounds:
+If setting `save-agnet`'s path in `gradle.properties` didn't help you (something doesn't work on Mac), you still can place all the files from `save-agent-*-distribution.jar` into `save-orchestrator/build/resources/main`.
+Moreover, if you use Mac with Apple Silicon, you should run `docker-mac-settings.sh` in order to let docker be avaliable via TCP.
+Do not forget to use `mac` profile.
 
 #### Note: 
 * This works only if snapshot version of save-core is set in lib.version.toml. 
--- a/save-deploy/README.md
+++ b/save-deploy/README.md
@@ -86,7 +86,7 @@ For example:<br/>
 the version corresponds to the file `save-0.4.0-alpha.0.42+78a24a8-linuxX64.kexe` <br/>
 
 #### Some workarounds:
-If setting `save-agnet`'s path in `gradle.properties` didn't help you (something doesn't work on Mac), you still can place all the files from `save-agent-*-distribution.jar` into `save-orchestrator/build/resources/main`.
+If setting `save-agent`'s path in `gradle.properties` didn't help you (something doesn't work on Mac), you still can place all the files from `save-agent-*-distribution.jar` into `save-orchestrator/build/resources/main`.
 Moreover, if you use Mac with Apple Silicon, you should run `docker-mac-settings.sh` in order to let docker be avaliable via TCP.
 Do not forget to use `mac` profile.
 
--- a/save-deploy/docker-mac-settings.sh
+++ b/save-deploy/docker-mac-settings.sh
@@ -1,0 +1,1 @@
+socat TCP-LISTEN:2375,range=127.0.0.1/32,reuseaddr,fork UNIX-CLIENT:/var/run/docker.sock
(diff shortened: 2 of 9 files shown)


save-deploy/docker-mac-settings.sh
```diff
@@ -0,0 +1 @@
+socat TCP-LISTEN:2375,range=127.0.0.1/32,reuseaddr,fork UNIX-CLIENT:/var/run/docker.sock
```
petertrr, line 1: Why can't orchestrator use UNIX socket directly? Also, can HTTP endpoint be enabled on docker daemon directly, without proxying script? On Linux we can add `-H 0.0.0.0:2375` to flags of `dockerd` command

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManager.kt
```diff
@@ -66,6 +66,7 @@ class DockerContainerManager(
         val buildImageResultCallback: BuildImageResultCallback = try {
             val buildCmd = dockerClient.buildImageCmd(dockerFile)
                 .withBaseDirectory(tmpDir)
+                .withPlatform("linux/amd64")
```
petertrr, line 69: Does it essentially act as no-op when platform is the same as host platform?

save-deploy/README.md
```diff
@@ -78,13 +78,18 @@ If you run on windows, dependency `save-agent` is omitted because of problems wi
...
-If you need to test changes in save-cli also you can compile snapshot version of save-cli on wsl <br/> 
+If you need to test changes in save-cli also you can compile snapshot version of save-cli on wsl <br/>
 and set saveCliPath and saveCliVersion in %USERPROFILE%\.gradle\gradle.properties <br/>
 For example:<br/>
 `saveCliPath=file:\\\\\\\\wsl$\\Ubuntu\\home\\username\\projects\\save-cli\\save-cli\\build\\bin\\linuxX64\\releaseExecutable` <br/> 
 `saveCliVersion=0.4.0-alpha.0.42+78a24a8` <br/>
 the version corresponds to the file `save-0.4.0-alpha.0.42+78a24a8-linuxX64.kexe` <br/>
 
+#### Some workarounds:
+If setting `save-agnet`'s path in `gradle.properties` didn't help you (something doesn't work on Mac), you still can place all the files from `save-agent-*-distribution.jar` into `save-orchestrator/build/resources/main`.
```
petertrr, line 89:
```suggestion
If setting `save-agent`'s path in `gradle.properties` didn't help you (something doesn't work on Mac), you still can place all the files from `save-agent-*-distribution.jar` into `save-orchestrator/build/resources/main`.
```

petertrr, line 89: Could you provide more details (as a comment or in PR description, to keep it for future reference) on what exactly doesn't work on Mac? I.e. how you specified `-PsaveAgentDistroFilepath` option and other details

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManager.kt
sanyavertolet: Well, `amd64` is, let's say, the most popular architecture. I mean mostly all the laptops and PCs have it. It seems like it should work the same as if `withPlatform` would not be present because `amd64` seems to be is a "default" architecture. And of cause we need linux.

save-deploy/README.md
sanyavertolet: As always

sanyavertolet: I just did everything that @icemachined advised (taking into consideration `save-deploy/README.md`). Folder `save-orchestrator/build/agentDistro` is created, it contains `save-agent-*-distribution.jar`, but it never extracts to `save-orchestrator/build/resources` as well as `agentDistro` is never found in `-classpath`

save-deploy/docker-mac-settings.sh
sanyavertolet: The point it that @icemachined suggested to use docker with `TCP` instead of `UNIX socket`. Talking about `-H` flag it seems that it should work, but there is no `dockerd` command on Mac, probably you meant `docker`

save-orchestrator/src/main/resources/application-mac.properties
```diff
@@ -1,2 +1,3 @@
 orchestrator.executionLogs=/Users/Shared/cnb/executionLogs/
 orchestrator.testResources.basePath=/Users/Shared/cnb/repositories
+orchestrator.adjustResourceOwner=false
```
akuleshov7, line 3: What is this for?

akuleshov7 (approved):
Very good job! Thank you

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManager.kt
petertrr: >  `amd64` seems to be is a "default" architecture

There can be a bunch of different images under the same tag but for different architectures, and what Docker does is it reads manifest file associated with the tag to find the image for the current host architecture. So on `linux/amd64` hosts setting `--platform linux/amd64` should indeed match the default behavior of Docker.

save-deploy/docker-mac-settings.sh
petertrr: > probably you meant `docker`

I honestly know nothing about specifics of Mac. On Linux where Docker is started as a Docker daemon `dockerd` there is a flag; on Windows there is a toggle in UI of Docker Desktop. If Docker for Mac works only with UI too, maybe there is a toggle as well?

save-orchestrator/src/main/resources/application-mac.properties
petertrr: This settings controls whether orchestrator `chown`s resources when packaging them into a container. I'm also curious why it should be disabled on Mac, maybe permission control differs from Linux?

save-deploy/docker-mac-settings.sh
sanyavertolet: As discussed, I can see no other workaround

save-deploy/README.md
sanyavertolet: I got it via GitHub (since #949 we build save-agent using Actions). The point is that something doesn't work - save agent is located in `$buildDir/agentDistro`, but when creating a docker in runtime, I got `FileNotFoundException` because `save-agent.kexe` could not be found

petertrr (approved):

PR #957, opened 2022-07-22
Add user avatar in top bar

### LINKED ISSUE #952: Avatar upload broken
Previously, before uploading a new avatar, we cleaned the folder. Now all the avatars are loaded into the folder and we cannot choose the ones we need.

Cheshiriks:
* Added user avatar in top bar
* Added delate folder with avatar users
 Closes #952 
avatarr

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
@@ -192,7 +192,7 @@ class DownloadFilesController(
             part.filename()
         )
         val content = part.content().map { it.asByteBuffer() }
-        avatarStorage.upload(avatarKey, content).map {
+        avatarStorage.upsert(avatarKey, content).map {
             logger.info("Saved $it bytes of $avatarKey")
             ImageInfo(avatarKey.getRelativePath())
         }
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
@@ -8,6 +8,9 @@ import com.saveourtool.save.backend.configs.ConfigProperties
 import com.saveourtool.save.storage.AbstractFileBasedStorage
 import com.saveourtool.save.utils.AvatarType
 import org.springframework.stereotype.Service
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+import java.nio.ByteBuffer
 import java.nio.file.Path
 import kotlin.io.path.div
 import kotlin.io.path.name
@@ -31,14 +34,27 @@ class AvatarStorage(configProperties: ConfigProperties) :
     )
 
     /**
+     * @param key
+     * @param content
+     * @return `Mono` with file size
+     */
+    fun upsert(key: AvatarKey, content: Flux<ByteBuffer>): Mono<Long> {
+        list()
+            .filter { it.imageName == key.folderUserName }
+            .singleOrEmpty()
+            .flatMap { delete(it) }
+        return upload(key, content)
+    }
+
+    /**
      * @param rootDir
      * @param key
      * @return [Path] is built by [AvatarKey] object
      */
     override fun buildPathToContent(rootDir: Path, key: AvatarKey): Path = rootDir
         .let { if (key.type == AvatarType.USER) it.resolve(USERS_DIRECTORY) else it }
+        .resolve(key.folderUserName)
         .resolve(key.imageName)
-        .resolve(key.filename)
 
     companion object {
         const val USERS_DIRECTORY = "users"
@@ -47,13 +63,13 @@ class AvatarStorage(configProperties: ConfigProperties) :
 
 /**
  * @property type
+ * @property folderUserName
  * @property imageName
- * @property filename
  */
 data class AvatarKey(
     val type: AvatarType,
+    val folderUserName: String,
     val imageName: String,
-    val filename: String,
 ) {
     /**
      * Added for backward compatibility
@@ -61,7 +77,7 @@ data class AvatarKey(
      * @return relative path to avatar image
      */
     fun getRelativePath(): String = when (type) {
-        AvatarType.ORGANIZATION -> "/$imageName/$filename"
-        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$imageName/$filename"
+        AvatarType.ORGANIZATION -> "/$folderUserName/$imageName"
+        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$folderUserName/$imageName"
     }
 }
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
@@ -4,6 +4,7 @@
 
 package com.saveourtool.save.frontend
 
+import com.saveourtool.save.domain.ImageInfo
 import com.saveourtool.save.domain.Role
 import com.saveourtool.save.domain.TestResultStatus
 import com.saveourtool.save.frontend.components.*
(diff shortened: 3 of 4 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -30,15 +33,28 @@ class AvatarStorage(configProperties: ConfigProperties) :
...
     )
 
+    /**
+     * @param key
+     * @param content
+     * @return `Mono` with file size
+     */
+    fun upsert(key: AvatarKey, content: Flux<ByteBuffer>): Mono<Long> {
+        list()
+            .filter { it.imageName == key.folderUserName }
```
nulls, line 43: why did you compare `imageName` with `folderUserName`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
```diff
@@ -121,18 +123,25 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
...
+                    loadingHandler = ::noopLoadingHandler,
+                )
+                    .unsafeMap {
+                        it.decodeFromJsonString<ImageInfo>()
+                    }
+            }
             userInfoNew?.let {
                 setState {
-                    userInfo = userInfoNew.copy(globalRole = globalRole)
+                    userInfo = userInfoNew.copy(globalRole = globalRole, avatar = avatarNew?.path)
```
nulls, line 139: can we remove `avatar` on `userInfo` if we don't use it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -192,7 +192,7 @@ class DownloadFilesController(
             part.filename()
```
nulls, line 192: I though we decide to remote filename from avatar name?

nulls (changes requested):

petertrr (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
Cheshiriks: We then decided to leave the full path to the avatar in the database and not change the name

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
Cheshiriks: Yes. Thank you. fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
Cheshiriks: We still decided to use the path to the avatar from the user object

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
```diff
@@ -97,9 +98,10 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
     }
 
     init {
-        state.userInfo = null
+        getUser()
```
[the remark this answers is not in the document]
Cheshiriks: Ok. Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
nulls: then you need to update this path in database when upsert happens

Cheshiriks: Yes, it is updated

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -30,15 +33,28 @@ class AvatarStorage(configProperties: ConfigProperties) :
...
+     * @param key
+     * @param content
+     * @return `Mono` with file size
+     */
+    fun upsert(key: AvatarKey, content: Flux<ByteBuffer>): Mono<Long> {
+        return list()
+            .filter { it.folderUserName == key.folderUserName }
+            .singleOrEmpty()
+            .map { delete(it) }
+            .flatMap { upload(key, content) }
```
akuleshov7, line 46: Here you will execute `upload of key and content` for each and every element from the `list`. If you will have 1000 elements you will do it 1000 times. Are you sure that you really want that? I doubt.

Cheshiriks: Why? I call the _singleOrEmpty()_ method before this

petertrr: Will `flatMap` be called if flux is empty?

Cheshiriks: Yes. Will not. Already checked

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -47,21 +62,21 @@ class AvatarStorage(configProperties: ConfigProperties) :
 
 /**
  * @property type
+ * @property folderUserName
  * @property imageName
- * @property filename
  */
 data class AvatarKey(
     val type: AvatarType,
+    val folderUserName: String,
```
petertrr, line 70: Maybe call it differently, because for `AvatarType.ORGNAIZATION` it won't be user name

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
Cheshiriks: Ok. Renamed

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -47,21 +62,21 @@ class AvatarStorage(configProperties: ConfigProperties) :
 
 /**
  * @property type
+ * @property folderName
  * @property imageName
- * @property filename
  */
 data class AvatarKey(
     val type: AvatarType,
+    val folderName: String,
```
nulls, line 70: can we rename it to 'objectName' for example. because folder name is implementation details which can be hidden here (it will be irrelevant after migration to S3)

Cheshiriks: Ok. Fixed it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
nulls: why we do it, if `User.avatar` is set on uploading a new avatar. No need to request it additionally

Cheshiriks: No. In App.kt, we do not look for a user in the database, we only take the name in authentication

nulls: I see, can we then fetch UserInfo from `/api/$v1/users` instead of fetching avatar info separately

nulls: similar in `UserSettingsView.kt`, we request avatar and user info separately, but avatar is part of user info

nulls: also, `/sec/user` returns only `userName`, but return object is `UserInfo` -- can we refactor it to String to avoid getting partially filled UserInfo object

Cheshiriks: OK. Now I'm returning a string and not UserInfo, and I'm requesting a User, not an avatar.
I will change the UserSetting later as part of another issue

nulls (approved):

Issue #959, opened 2022-07-22
CreateContestView


sanyavertolet:
We need to be able to create contests both on
- [x]  backend
- [ ] frontend

PR #964, opened 2022-07-25
Contest creation on backend

sanyavertolet:
This PR is connected with #959 

### What's done:
- [x] Added endpoint for contest creation
- [x] Added column for `organization` table - a flag that indicates whether an organization can create contests or not
- [x] Added column for `contest` table - id of an organization that has created this contest
- [x] Reworked `LnkUserOrganizationController` - made it reactive
- [x] Added swagger annotations for `LnkUserOrganizationController`
- [x] Added swagger annotations for `ContestController`
- [x] Updated dev database insertions


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
```diff
@@ -42,7 +61,18 @@ internal class ContestController(
      * @return contest with name [contestName]
      */
     @GetMapping("/{contestName}")
+    @RequiresAuthorizationSourceHeader
     @PreAuthorize("permitAll()")
+    @Operation(
+        method = "GET",
+        summary = "Get contest by name.",
```
kgevorkyan, line 68: Do we really need both `summary` and `description`, while they are the same?
Isn't enough only one of them?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ContestService.kt
```diff
@@ -82,4 +82,17 @@ class ContestService(
         ?.let {
             testSuitesService.findTestSuiteById(it)?.getOrNull()
         }
+
+    /**
+     * @param newContest
+     * @return true if contest was saved, else otherwise
```
kgevorkyan, line 88:
```suggestion
     * @return true if contest was saved, false otherwise
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
```diff
@@ -101,4 +162,54 @@ internal class ContestController(
...
+        .switchIfEmpty {
+            Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
+        }
+        .filter {
+            organizationPermissionEvaluator.canCreateContests(it, authentication)
+        }
+        .switchIfEmpty {
+            Mono.error(ResponseStatusException(HttpStatus.FORBIDDEN))
+        }
+        .map {
```
kgevorkyan, line 204: Couldn't we do it in one step?
```kotlin
filter {
contestService.createContestIfNotPresent(contestDto.toContest(it))
}
```
instead of `map` + `filter`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
sanyavertolet: `summary` is said to be less than 120 characters, I am not sure, but it might be displayed in a different way. Moreover it looks prettier because we have three arguments in every `@Operation`

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ContestService.kt
sanyavertolet: of course

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
sanyavertolet: That is a good idea, thanks
Will do it

sanyavertolet: Done

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -42,125 +68,223 @@ class LnkUserOrganizationController(
      * @param organizationName
      * @param authentication
      * @return list of users with their roles, connected to the organization with [organizationName]
-     * @throws ResponseStatusException
      */
```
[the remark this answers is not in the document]
sanyavertolet: Ok, done and updated `diktat-analysis.yml`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ContestController.kt
sanyavertolet: screenshot
Take a look here: the endpoint at the top has only `description` set and the endpoint at the bottom has both of them set.

PR #965, opened 2022-07-25
Problem of username selection: improvements

### LINKED ISSUE #915: Problem of username selection: improvements
Right now we give user exactly the same name/id, that he has in the initial system (codehub/github). 
We need to support two main things on the frontend:
1) "Registration"-like page, where user will be able to change/update his nickname OR use one we have suggested;
2) Possibility to change nick-name (if it is not yet used by other people).

Cheshiriks:
* added Registration view
Closes #915


api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
```diff
@@ -128,7 +128,7 @@ class WebSecurityConfig(
             it.authenticationSuccessHandler(
                 DelegatingServerAuthenticationSuccessHandler(
                     StoringServerAuthenticationSuccessHandler(configurationProperties),
-                    RedirectServerAuthenticationSuccessHandler("/#/projects"),
+                    RedirectServerAuthenticationSuccessHandler("/#/registration"),
```
sanyavertolet, line 131: Will it work only when it is the first time that the user is authenticated with his GitHub account?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+    /* companion object :
+        RStatics<RegistrationProps, RegistrationViewState, RegistrationView, Context<RequestStatusContext>>(
+            RegistrationView::class
+        ) {
+        init {
+            RegistrationView.contextType = requestStatusContext
+        }
+    }*/
```
sanyavertolet, line 266: Why is this companion object commented?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,267 @@
...
+    @Suppress(
+        "TOO_LONG_FUNCTION",
+        "EMPTY_BLOCK_STRUCTURE_ERROR",
+        "LongMethod",
+        "MAGIC_NUMBER",
+        "PARAMETER_NAME_IN_OUTER_LAMBDA",
+    )
+    override fun ChildrenBuilder.render() {
+        if (state.userInfo?.isActive == false) {
+            ReactHTML.main {
```
sanyavertolet, line 151: Remove `ReactHTML` please

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,267 @@
...
+                                            }
+                                                ?: run {
+                                                    "img/undraw_profile.svg"
+                                                }
+                                            height = 260.0
+                                            width = 260.0
+                                        }
+                                    }
+                                    ReactHTML.form {
+                                        className = ClassName("needs-validation")
```
sanyavertolet, line 193: Not sure if we need this because it should already be inside `inputTextFormRequired`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,267 @@
...
+    var isValidUserName: Boolean?
+}
+
+/**
+ * A Component for registration view
+ */
+@JsExport
+@OptIn(ExperimentalJsExport::class)
+class RegistrationView : AbstractView<RegistrationProps, RegistrationViewState>() {
+    private val fieldsMap: MutableMap<InputTypes, String> = mutableMapOf()
```
sanyavertolet, line 84: I strongly recommend you to use state here instead of private field of class as far as it is more react style (afaik)

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/domain/UserSaveStatus.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * New user
+     */
+    NEW("User saved successfully"),
+
+    /**
+     * Update user
+     */
+    UPDATE("User update successfully"),
```
sanyavertolet, line 24:
```suggestion
    UPDATE("User successfully updated"),
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: It's strange that the Diktat does not swear at this

Cheshiriks: Created issue https://github.com/saveourtool/diktat/issues/1510

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
Cheshiriks: Yes. We check if the user is registered. If so, move it to _projects_ page

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
sanyavertolet: Why did you completely remove companion object? Don't we need it here? I thought you might want to process responses using `responseHandlerWithValidation`.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -44,33 +50,53 @@ class UsersDetailsController(
     @GetMapping("/{userName}")
     @PreAuthorize("permitAll()")
     fun findByName(@PathVariable userName: String): Mono<UserInfo> =
-            userRepository.findByName(userName).toMonoOrNotFound().map { it.toUserInfo() }
+            userRepository.findByName(userName)?.let { user ->
+                user.toMonoOrNotFound().map { it.toUserInfo() }
+            } ?: run {
```
petertrr, line 55:
```suggestion
            userRepository.findByName(userName)
                ?.toMonoOrNotFound()?.map { it.toUserInfo() }
                ?:
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
```diff
@@ -44,11 +47,12 @@ class UsersController(
...
     fun saveNewUser(@RequestBody user: User) {
         val userName = requireNotNull(user.name) { "Provided user $user doesn't have a name" }
-        userRepository.findByName(userName)?.let {
+        originalLoginRepository.findByName(userName)?.user?.let {
             logger.debug("User $userName is already present in the DB")
         } ?: run {
             logger.info("Saving user $userName to the DB")
-            userRepository.save(user)
+            val newUser = userRepository.save(user)
+            originalLoginRepository.save(OriginalLogin(user.name, newUser, user.source))
```
petertrr, line 55: This should be encapsulated in `UserDetailsService`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
```diff
@@ -44,11 +47,12 @@ class UsersController(
     @PostMapping("/new")
     fun saveNewUser(@RequestBody user: User) {
         val userName = requireNotNull(user.name) { "Provided user $user doesn't have a name" }
-        userRepository.findByName(userName)?.let {
+        originalLoginRepository.findByName(userName)?.user?.let {
```
petertrr, line 50: Why are you checking `originalLoginRepository` here? What if we need to save a user with a name that exists in original logins but from a new authorization source?

petertrr, line 50: Why are you checking `originalLoginRepository` here? What if we need to save a user with a name that exists in original logins but from a new authorization source?

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/ValidateRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+     */
+    @Query("""select if (count(*) = 0, true, false) from save_cloud.high_level_names where name = :name""", nativeQuery = true)
+    fun validateName(@Param("name") name: String): Long
+
+    /**
+     * @param name
+     */
+    @Query("""insert into save_cloud.high_level_names set name = :name""", nativeQuery = true)
+    @Modifying
+    fun saveHighName(@Param("name") name: String)
```
petertrr, line 23:
```suggestion
    fun saveHighLevelName(@Param("name") name: String)
```

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/OrganizationControllerTest.kt
```diff
@@ -48,7 +48,7 @@ import java.util.*
     ProjectPermissionEvaluator::class,
     LnkUserProjectService::class,
     UserDetailsService::class,
-
+    OriginalLoginRepository::class,
```
petertrr, line 51: Should be either as `@Import` or `@MockBean`

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
```diff
@@ -19,10 +21,17 @@ import kotlinx.serialization.Serializable
...
+ * @property isActive
+ * @property oldName
+ * @property originalLogins
  */
 @Serializable
 data class UserInfo(
     val name: String,
+    val id: Long? = null,
+    val oldName: String? = null,
+    val originalLogins: List<String?> = emptyList(),
```
petertrr, line 34: Should be list of pairs, together with source?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
```diff
@@ -19,10 +21,17 @@ import kotlinx.serialization.Serializable
...
+ * @property id
+ * @property isActive
+ * @property oldName
+ * @property originalLogins
  */
 @Serializable
 data class UserInfo(
     val name: String,
+    val id: Long? = null,
+    val oldName: String? = null,
```
petertrr, line 33: What is `oldName`? Is it the previous login?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Execution.kt
```diff
@@ -43,7 +42,7 @@ import javax.persistence.ManyToOne
 @Entity
 class Execution(
 
-    @ManyToOne(fetch = FetchType.EAGER)
+    @ManyToOne
```
petertrr, line 45: Why did you remove it? Also, isn't `EAGER` the default value?

petertrr (changes requested):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: Yes. This is necessary to change the name. This is the new name and the previous one.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
sanyavertolet: or even a map

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
petertrr: Why do we store only the previous name then? User may change the name multiple times.

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
petertrr: But why do we need to open registration page every time? We should distinguish fisrt-time users from already registered ones

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: We need the old name only in the wrapper for the front, and only in order to find the user in the database by it.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: No. OriginalLogins contains both the name and the source

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/OrganizationControllerTest.kt
Cheshiriks: Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/ValidateRepository.kt
Cheshiriks: Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/UsersDetailsController.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Removed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
Cheshiriks: This is checked further. We will open a registration page and ask the user to change their login.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: I think it's unnecessary here. The only check required is to display the response in the Input form itself.

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
Cheshiriks: Yes, on the registration view there is a check that the user is registered. And redirect the user to the project view even before rendering.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
petertrr: I can see that you are trying to retrieve a user by their original username and if it's not found you start saving the user as a new user. What I'm asking is, suppose we have a user `[name = "johndoe", source = "github"]`, and then another user, `[name = "johndoe", source = "facebook"]` tries to authenticate. Seems like with current logic you would treat the new user as if they were already present, because there is a row with the same name in `original_login` table

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
petertrr: In `UserInfo` you store only names of these logins

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
Cheshiriks: Fixed

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
petertrr: Ah, OK, so this field should be `null` except for the renaming process. Could you please add this to class KDoc?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
Cheshiriks: Ok. Agree. Changed on _findByNameAndSource_

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,276 @@
...
+import kotlinx.serialization.json.Json
+
+/**
+ * `Props` retrieved from router
+ */
+external interface RegistrationProps : PropsWithChildren {
+    /**
+     * Currently logged in user or null
+     */
+    var userName: String?
```
petertrr, line 54: Can we pass `userInfo` from `App.kt` instead of just name?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: Added

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,276 @@
...
+                                    }
+                                    label {
+                                        input {
+                                            type = InputType.file
+                                            hidden = true
+                                            onChange = { event ->
+                                                postImageUpload(event.target)
+                                            }
+                                        }
+                                        ariaLabel = "Change organization's avatar"
```
petertrr, line 197:
```suggestion
                                        ariaLabel = "Change user's avatar"
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,276 @@
...
+                        className = ClassName("row justify-content-center")
+                        div {
+                            className = ClassName("col-sm-4")
+                            div {
+                                className = ClassName("container card o-hidden border-0 shadow-lg my-2 card-body p-0")
+                                div {
+                                    className = ClassName("p-5 text-center")
+                                    h1 {
+                                        className = ClassName("h4 text-gray-900 mb-4")
+                                        +"Registration new user"
```
petertrr, line 187:
```suggestion
                                        +"Registration of a new user"
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,276 @@
...
+                                        ariaLabel = "Change organization's avatar"
+                                        img {
+                                            className =
+                                                    ClassName("avatar avatar-user width-full border color-bg-default rounded-circle")
+                                            src = state.image?.path?.let {
+                                                "/api/$v1/avatar$it"
+                                            }
+                                                ?: run {
+                                                    "img/undraw_profile.svg"
+                                                }
```
petertrr, line 206:
```suggestion
                                                ?: "img/undraw_profile.svg"
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,276 @@
...
+
+    @Suppress(
+        "TOO_LONG_FUNCTION",
+        "EMPTY_BLOCK_STRUCTURE_ERROR",
+        "LongMethod",
+        "MAGIC_NUMBER",
+        "PARAMETER_NAME_IN_OUTER_LAMBDA",
+    )
+    override fun ChildrenBuilder.render() {
+        if (state.userInfo?.isActive == false) {
```
petertrr, line 169: Can we maybe move this check into `getUser()` in `App.kt`? I.e., when any page is opened, we get `UserInfo` and if `UserInfo.isActive == false`, redirect the user to registration page. This way we won't need to start authorization from the registration page, even if it's not visible

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
petertrr: https://github.com/saveourtool/save-cloud/pull/965#discussion_r953695679

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Moved

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Ok. Fixed

akuleshov7:
Logic changed (bug?): Now I see my name instead of login!

akuleshov7 (changes requested):
incorrect login processing

djhdjsdkjhnkjs:
This PR does not work at all:

image

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
```diff
@@ -163,6 +163,7 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
         HashRouter {
             requestModalHandler {
                 userInfo = state.userInfo
+
```
petertrr, line 166: I tried to tinker with your PR a little and found that this solution doesn't lead to endless loop:
```kotlin
                withRouter<Props> { location, _ ->
                    if (state.userInfo?.isActive == false && !location.pathname.startsWith("/${FrontendRoutes.REGISTRATION.path}")) {
                        Navigate {
                            to = "/${FrontendRoutes.REGISTRATION.path}"
                            replace = false
                        }
                    }
                }()
```

Basically, while `Navigate` is mounted in components tree, during each render it leads to another redirect. I'm still not sure, but I think that actual `window.location` and internal location representation in react-router's state are not fully synced. In my example we are asserting on location from `HashRouter`'s state.

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
```diff
@@ -24,6 +25,7 @@ class StoringServerAuthenticationSuccessHandler(
 ) : ServerAuthenticationSuccessHandler {
     private val logger = LoggerFactory.getLogger(javaClass)
     private val objectMapper = ObjectMapper()
+        .configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, false)
```
petertrr, line 28: Is it still needed? I think it causes empty arrays to be completely absent in JSONs, but how are non-empty arrays represented?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
petertrr: But there is still a check here

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/welcome/WelcomeView.kt
```diff
@@ -113,10 +118,11 @@ class WelcomeView : AbstractView<WelcomeProps, IndexViewState>(true) {
                 }
 
                 div {
-                    Particles::class.react {
+                    // FixMe: Particles are superimposed on top of the view in some transitions
```
petertrr, line 121: Maybe more details / note that they block user interactions?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * Image to owner avatar
+     */
+    var image: ImageInfo?
+
+    /**
+     * Currently logged in user or null
+     */
+    var userInfo: UserInfo?
```
petertrr, line 79: Why do we need a copy of UserInfo in component's state? Can we use value from props?

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
Cheshiriks: At the time of user creation, this array will always be empty. Otherwise, we will receive the Kotlin.EmptyClass class in Json and will not be able to serialize it.

petertrr: Please add a comment in code about why we need this feature (btw its `EmptyList`, not `EmptyClass`) and add explicit initialization of `originalLogins` to `emptyList` in `fun Authentication.toUser()`. This way we probably won't accidentally change it to something different

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: This is necessary so that you cannot enter the view by url if isActive = false and the user is redirected to the project view

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Removed it

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/StoringServerAuthenticationSuccessHandler.kt
Cheshiriks: Ok. Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
```diff
@@ -22,22 +26,30 @@ import java.util.*
...
     private val userRepository: UserRepository,
+    private val originalLoginRepository: OriginalLoginRepository,
 ) : ReactiveUserDetailsService {
     private val logger = LoggerFactory.getLogger(javaClass)
 
     override fun findByUsername(username: String): Mono<UserDetails> =
-            { userRepository.findByName(username) }
-                .toMono()
-                .getIdentitySourceAwareUserDetails(username)
+            userRepository.findByName(username)?.let {
```
akuleshov7, line 34: 1) All the logic should be unified:
```Mono.just()
          .map {
                
          }
```
2) You duplicate same call twice: ` .getIdentitySourceAwareUserDetails(username)`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,258 @@
...
+                                    }
+                                    ariaLabel = "Change user's avatar"
+                                    img {
+                                        className =
+                                                ClassName("avatar avatar-user width-full border color-bg-default rounded-circle")
+                                        src = state.image?.path?.let {
+                                            "/api/$v1/avatar$it"
+                                        }
+                                            ?: "img/undraw_profile.svg"
+                                        height = 260.0
```
akuleshov7, line 192: NEVER ever set width with pixels

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,258 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    private fun postImageUpload(element: HTMLInputElement) =
+            scope.launch {
+                setState {
+                    isUploading = true
```
akuleshov7, line 238: Why isUploading needed here? Do you show any animation or what?

akuleshov7 (changes requested):
говнокод опять

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,258 @@
...
+                }
+                div {
+                    className = ClassName("row justify-content-center")
+                    div {
+                        className = ClassName("col-sm-4")
+                        div {
+                            className = ClassName("container card o-hidden border-0 shadow-lg my-2 card-body p-0")
+                            div {
+                                className = ClassName("p-5 text-center")
+                                h1 {
```
akuleshov7, line 172: please split to components

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -0,0 +1,258 @@
...
+            val response = post(
+                "$apiUrl/users/save",
+                jsonHeaders,
+                Json.encodeToString(newUserInfo),
+                loadingHandler = ::classLoadingHandler,
+                responseHandler = ::classComponentResponseHandlerWithValidation,
+            )
+            if (response.ok) {
+                window.location.href = "#/${FrontendRoutes.PROJECTS.path}"
+                window.location.reload()
```
akuleshov7, line 139: why do you need to reload?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
akuleshov7: you concatenated it?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Need to update the states in App.kt otherwise the redirect will not let us leave this view

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Removed it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

akuleshov7 (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: No. These are two different fields.

PR #977, opened 2022-07-28
Implement `KubernetesPersistentVolumeService` and support using tests from PV

### LINKED ISSUE #924: Orchestrator: Use volumes with data instead of bundling it into an image
In orchestrator we bundle tests and binaries into a single docker image. Instead, we can simply create a volume, copy data into it and mount in to every container with running agent. Things to consider:
- Bundling several SDKs into an image: we'll have to have some of such images already prepared (or use JIB to build them without Docker)

Part of #727, because for Kubernetes + Docker it's impossible to use alternative Docker runtimes, and with containerd we can't use `docker build`

Considerations:
* It's hard to completely get rid of custom images: we need not only test resources, but some system packages like `libcurl` for save-agent
* Even if we migrate build from Docker to JIB there is an open question how to import the image into Kubernetes. However there's nothing stopping us from building ~20 base images and keeping them in any OCI registry.

Tasks:
* [x] Implementation for Docker with Docker volumes: #962, #979
* [ ] Implementation for Kubernetes with Kubernetes PVCs

petertrr:
* Implement `KubernetesPersistentVolumeService` and support using tests from PV. Resources are copied from an intermediate location into a newly created PV
* [Helm] Attach an intermediate volume to orchestrator (using `hostPath` for Minikube, configurable)
* [Helm] Add cluster-scoped role for orchestrator to be able to create PVs
* [Helm] Make orchestrator's `application.properties` configurable
* Move properties for Kubernetes and Minikube into `ConfigMap`s and Helm values instead of properties files
* [Helm] Make `PullPolicy` configurable
* [Helm] Increase startup failure threshold for all spring-boot services
* Add notes on running in Minikube
* Use latest version of `chart-testing` for linting Helm chart
* Fix typo from the previous PR (`mount` -> `volume` in DockerPersistentVolumeService.kt)
* [Gradle] Replace all symbols in docker tag, that are not allowed, with dashes

This change implemets a temporary solution: Delete temporary directory only after agents are started (or failed to do so). Should be discarded in favor of:
* (Subsequent PR) Download resources from backend/storage into the volume directly (by init contaienr or by agent)
* (subsequent PR) Implement deletion in PersistentVolumeManagers

Closes #924


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
```diff
@@ -67,20 +70,43 @@ class KubernetesManager(
...
+                                )
+                            }
+                        )
                         containers = listOf(
                             Container().apply {
                                 name = "save-agent-pod"
                                 metadata = ObjectMeta().apply {
                                     labels = mapOf(
                                         "executionId" to executionId.toString(),
-                                        "baseImageId" to baseImageId,
```
petertrr, line 76: Docker image ID doesn't match format of appropriate Kubernetes label

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
```diff
@@ -67,20 +76,48 @@ class KubernetesManager(
                         }
                         // FixMe: Orchestrator doesn't push images to a remote registry, so agents have to be run on the same host.
                         nodeName = System.getenv("NODE_NAME")
+                        initContainers = listOf(
```
kgevorkyan, line 79: May be we can extract configuration of this fields for this job into some separated methods, what do you think?
like `initContainers = configureContainers()` and so on?

save-cloud-charts/save-cloud/Chart.lock
```diff
@@ -1,15 +1,15 @@
 dependencies:
 - name: grafana
   repository: https://grafana.github.io/helm-charts
-  version: 6.31.1
+  version: 6.32.10
```
nulls, line 4: does renovate able to upgrade it? I don't remember PRs for it

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -327,11 +329,13 @@ class DockerService(
      * @property imageId ID of an image which should be used for a container
      * @property runCmd command that should be run as container's entrypoint
      * @property pvId ID of a persistent volume that should be attached to a container
+     * @property resourcesPath FixMe: needed only until agents download test by themselves
```
nulls, line 332: and evaluated tools too?

save-cloud-charts/save-cloud/Chart.lock
petertrr: I was hoping that's what this setting is for
https://github.com/saveourtool/save-cloud/blob/0cd164a405d9440d4c9003d5cb21af1a16126f07/renovate.json#L45-L47

But apparently, it needs something else, like global toggle for lock files maintenance. The problem is that we also have a `yarn.lock`, which shouldn't be updated automatically, so this needs some more experiments.

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
petertrr: Extracted this one

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -125,17 +126,23 @@ class AgentsController(
...
                 }
                 .subscribe()
         }
     }
 
-    // if some additional file is archive, unzip it into proper destination:
-    // for standard mode into STANDARD_TEST_SUITE_DIR
-    // for Git mode into testRootPath
+    /**
+     * if some additional file is archive, unzip it into proper destination:
```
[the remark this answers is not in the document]
petertrr: Removed

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
```diff
@@ -98,23 +135,54 @@ class KubernetesManager(
                                         }
                                     }
                                 )
-                                command = agentRunCmd.split(" ")
+
+                                // `agentRunCmd` looks like `sh -c "rest of the command"`
+                                val (command, args) = agentRunCmd
+                                    .let {
+                                        it.substringBefore('"').trim().split(" ") + "\"${it.substringAfter('"')}"
```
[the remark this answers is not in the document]
petertrr: I ended up changing type of `Configuration.runCmd` to List, because there is only a single place where we use it as a string.

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
petertrr: Changed

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesPersistentVolumeService.kt
```diff
@@ -19,22 +23,61 @@ import kotlin.io.path.createTempDirectory
...
     private val kc: KubernetesClient,
+    private val configProperties: ConfigProperties,
 ) : PersistentVolumeService {
+    @Suppress("TOO_LONG_FUNCTION")
     override fun createFromResources(resources: Collection<Path>): KubernetesPvId {
-        val tmpDir = createTempDirectory()
-        resources.forEach {
-            it.copyTo(tmpDir)
+        requireNotNull(configProperties.kubernetes)
+        if (resources.size > 1) {
```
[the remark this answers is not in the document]
petertrr: Changed

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
kgevorkyan: But why only this one?

nulls (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
petertrr: Extracted agent container too

kgevorkyan (approved):

Issue #363, opened 2021-10-27
Create ProjectDTO


Cheshiriks:
It's necessary to split the Project entity into ProjectDTO

nulls:
duplicates #527

PR #978, opened 2022-07-28
Add Can Create Contests checkbox

sanyavertolet:
### What's done:
- [x] Added checkbox in OrganizationSettingsMenu.kt for canCreateContests changing (SUPER_ADMIN only)
- [x] Added endpoint on backend for making an organization to be contest creator
- [x] Forbade to change canCreateContests in update method

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -73,7 +73,7 @@ internal class OrganizationController(
 
     /**
      * @param organizationName name of an organization
-     * @param newCanCreateContests new value of flag Organization.canCreateContests
+     * @param isAbleToCreateContests new value of flag Organization.canCreateContests
      * @param authentication an [Authentication] representing an authenticated request
      * @return response
      */
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -99,7 +99,7 @@ internal class OrganizationController(
             Mono.error(ResponseStatusException(HttpStatus.FORBIDDEN))
         }
         .map { organization ->
-            organizationService.setAbilityToCreateContest(
+            organizationService.updateOrganization(
                 organization.copy(canCreateContests = isAbleToCreateContests).apply { id = organization.id }
             )
             ResponseEntity.ok("Organization updated")
@@ -135,26 +135,27 @@ internal class OrganizationController(
     }
 
     /**
-     * @param organization updateOrganization
+     * @param organizationName name of an organization that should be changed
+     * @param organization draft organization that should be saved as an organization with name [organizationName]
      * @param authentication an [Authentication] representing an authenticated request
      * @return response
      */
     @PostMapping("/{organizationName}/update")
-    @Suppress("UnsafeCallOnNullableType")
     fun updateOrganization(
+        @PathVariable organizationName: String,
         @RequestBody organization: Organization,
         authentication: Authentication,
     ): Mono<StringResponse> = Mono.just(
-        organization.name
+        organizationName
     )
         .flatMap {
-            organizationService.findByName(organization.name).toMono()
+            organizationService.findByName(it).toMono()
         }
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.NOT_FOUND))
         }
         .filter {
-            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, organization.name, Role.OWNER)
+            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.OWNER)
         }
         .switchIfEmpty {
             Mono.error(ResponseStatusException(HttpStatus.FORBIDDEN))
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -23,6 +23,7 @@ import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
 import reactor.kotlin.core.publisher.switchIfEmpty
 import java.time.LocalDateTime
+import java.util.*
 
 /**
  * Controller for working with organizations.
@@ -70,6 +71,32 @@ internal class OrganizationController(
     }
 
     /**
+     * @param organizationName name of an organization
+     * @param newCanCreateContests new value of flag Organization.canCreateContests
+     * @param authentication an [Authentication] representing an authenticated request
+     * @return response
+     */
+    @PostMapping("/{organizationName}/manage-contest-permission")
+    @Suppress("UnsafeCallOnNullableType")
+    fun changeCanCreateContests(
+        @PathVariable organizationName: String,
+        @RequestParam newCanCreateContests: Boolean,
+        authentication: Authentication
+    ): Mono<StringResponse> = Mono.justOrEmpty(
+        Optional.ofNullable(
+            organizationService.findByName(organizationName)
+        )
+    )
+        .filter {
+            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.SUPER_ADMIN)
+        }
+        .map { organization ->
+            organizationService.changeCanCreateContests(organization.copy(canCreateContests = newCanCreateContests).apply { id = organization.id })
+            ResponseEntity.ok("Organization updated")
+        }
+        .defaultIfEmpty(ResponseEntity.status(HttpStatus.FORBIDDEN).build())
+
+    /**
      * @param newOrganization newOrganization
      * @param authentication an [Authentication] representing an authenticated request
      * @return response
@@ -105,9 +132,13 @@ internal class OrganizationController(
      */
     @PostMapping("/{organizationName}/update")
     @Suppress("UnsafeCallOnNullableType")
-    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> = Mono
-        .just(organization)
-        .filter { organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.OWNER) }
+    fun updateOrganization(
+        @RequestBody organization: Organization,
+        authentication: Authentication,
+    ): Mono<StringResponse> = Mono.just(organization)
+        .filter {
+            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.OWNER)
+        }
         .map {
             organizationService.updateOrganization(organization)
             ResponseEntity.ok("Organization updated")
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -22,6 +22,7 @@ import org.springframework.web.server.ResponseStatusException
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
 import reactor.kotlin.core.publisher.switchIfEmpty
+import reactor.kotlin.core.publisher.toMono
 import java.time.LocalDateTime
 import java.util.*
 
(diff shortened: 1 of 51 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -105,9 +132,13 @@ internal class OrganizationController(
      */
     @PostMapping("/{organizationName}/update")
```
nulls, line 133: organizationName is provided but it's not used, suggestion to search organization by name and replace by provided value

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -69,7 +69,18 @@ class OrganizationService(
...
+        organization.name
+    )!!.let { organizationFromDb ->
+        organizationRepository.save(organizationFromDb.copy(canCreateContests = organization.canCreateContests).apply { id = organizationFromDb.id })
+    }
+
+    /**
+     * @param organization
+     * @return organization
+     */
+    fun changeCanCreateContests(organization: Organization): Organization = organizationRepository.save(organization)
```
nulls, line 83: why we need a separate method for it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -69,6 +70,32 @@ internal class OrganizationController(
...
+     */
+    @PostMapping("/{organizationName}/manage-contest-permission")
+    @Suppress("UnsafeCallOnNullableType")
+    fun changeCanCreateContests(
+        @PathVariable organizationName: String,
+        @RequestParam newCanCreateContests: Boolean,
+        authentication: Authentication
+    ): Mono<StringResponse> = Mono.justOrEmpty(
+        Optional.ofNullable(
+            organizationService.findByName(organizationName)
```
nulls, line 87: we can search on next step
also you can use method 'toMono' instead of Optional

nulls (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -504,6 +504,8 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                 }
             }
             updateNotificationMessage = ::showNotification
+            organization = state.organization ?: Organization.stub(-1)
```
nulls, line 507: stub was created for tests purpose -- can we request a real organization here? 
also, such update change all fields don't it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -69,7 +69,18 @@ class OrganizationService(
      * @param organization
      * @return organization
      */
-    fun updateOrganization(organization: Organization): Organization = organizationRepository.save(organization)
+    @Suppress("UnsafeCallOnNullableType")
+    fun updateOrganization(organization: Organization): Organization = organizationRepository.findByName(
+        organization.name
+    )!!.let { organizationFromDb ->
+        organizationRepository.save(organizationFromDb.copy(canCreateContests = organization.canCreateContests).apply { id = organizationFromDb.id })
```
nulls, line 76: so `updateOrganization` updates only `canCreateContests`?

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: Because in one method we provide changing `canCreateContests` only and in another we provide changing other parameters. To my mind we should forbid changing canCreateContests flag separately

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: Oh, my bad.
It should change everything except canCreateContests. Will fix it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: We can, but it seems that there is no difference

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -69,6 +70,32 @@ internal class OrganizationController(
...
 
+    /**
+     * @param organizationName name of an organization
+     * @param newCanCreateContests new value of flag Organization.canCreateContests
+     * @param authentication an [Authentication] representing an authenticated request
+     * @return response
+     */
+    @PostMapping("/{organizationName}/manage-contest-permission")
+    @Suppress("UnsafeCallOnNullableType")
+    fun changeCanCreateContests(
```
akuleshov7, line 81: `changeCanCreateContests` -> `setAbilityToCreateContest`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
akuleshov7: code is more accurate

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -69,6 +70,32 @@ internal class OrganizationController(
...
+        @PathVariable organizationName: String,
+        @RequestParam newCanCreateContests: Boolean,
```
akuleshov7, line 83: `isAbleToCreateContests`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Ok, will do it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Will do it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Will do it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
sanyavertolet: That is just a replacement of `null` in order to ignore null checks or unsafe null casts later

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: ok, will do it

akuleshov7 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -105,14 +141,30 @@ internal class OrganizationController(
...
-        .map {
-            organizationService.updateOrganization(organization)
+    fun updateOrganization(
+        @RequestBody organization: Organization,
+        authentication: Authentication,
+    ): Mono<StringResponse> = Mono.just(
+        organization.name
+    )
+        .flatMap {
+            organizationService.findByName(organization.name).toMono()
```
Cheshiriks, line 151: We are already getting the "organization" object from the frontend. Why do you give a name and look for it again in the database?

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: In case of update we need to be sure that canCreateContests is not changed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -105,14 +141,30 @@ internal class OrganizationController(
...
-    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> = Mono
-        .just(organization)
-        .filter { organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.OWNER) }
-        .map {
-            organizationService.updateOrganization(organization)
+    fun updateOrganization(
+        @RequestBody organization: Organization,
+        authentication: Authentication,
+    ): Mono<StringResponse> = Mono.just(
+        organization.name
```
akuleshov7, line 148: yeah, @Cheshiriks is right - this is very strange - you do the following:
1) Get organization object
2) Get name from it 
3) Find this organization again by name 
4) Save Object 
5) get a name from an object 
6) Find object again 
7) Save results

akuleshov7 (changes requested):
pls look at controller logic

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -69,7 +69,18 @@ class OrganizationService(
      * @param organization
      * @return organization
      */
-    fun updateOrganization(organization: Organization): Organization = organizationRepository.save(organization)
+    @Suppress("UnsafeCallOnNullableType")
+    fun updateOrganization(organization: Organization): Organization = organizationRepository.findByName(
```
Cheshiriks, line 73: This is a duplicate of your logic from the Controller. She shouldn't be here at all. The service should not contain any additional logic and should only call the Save method.

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: Agree, forgot to remove it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Fixed this. Hope everything else is alright

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -99,20 +135,43 @@ internal class OrganizationController(
...
      */
     @PostMapping("/{organizationName}/update")
-    @Suppress("UnsafeCallOnNullableType")
-    fun updateOrganization(@RequestBody organization: Organization, authentication: Authentication): Mono<StringResponse> = Mono
-        .just(organization)
-        .filter { organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.OWNER) }
-        .map {
-            organizationService.updateOrganization(organization)
+    fun updateOrganization(
+        @PathVariable organizationName: String,
```
Cheshiriks, line 145: I don't understand why we need an PathVariable _organizationName_ if we already have an Organization object that contains the name.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -99,20 +135,43 @@ internal class OrganizationController(
...
+        @RequestBody organization: Organization,
+        authentication: Authentication,
+    ): Mono<StringResponse> = Mono.just(
+        organizationName
+    )
+        .flatMap {
+            organizationService.findByName(it).toMono()
```
Cheshiriks, line 152: In OrganizationView in componentDidMount we get the _organization_ from the database. We change something in it and return it to the backend. The organization already contains an ID. We don't have to look it up again in the database by name.
The _save(Organization)_ method will not create a new entry, but will update the old one by ID

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: It might be more user-friendly to use organizationName instead of id
In this case id is not required.
But perhaps I am wrong.
If you find it to be a bad practice, I can remove `organizationName` from path

sanyavertolet: That also might be good if we want to provide an ability to change organization name

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
petertrr: We generally can't rely on *all* fields of an object received from a network request. For example, to check `canCreateContests` for a certain organization we have to look it up in the database by name or by ID, because user may have sent a request with maliciously altered value. Same goes for the combination of name and ID. So we still need to use some key to look up the real data for an entity and operate with it.

Cheshiriks: Ok. Can we then create an OrganizationDTO entity to pass data between the frontend and backend?

sanyavertolet: I guess there is an issue for that

sanyavertolet: There is a similar one https://github.com/saveourtool/save-cloud/issues/363

Cheshiriks: Yes. I forgot. OK. Let's leave it as is for now. But we will have to think about creating this layer in the future for the project and organization in the future.

akuleshov7 (approved):

PR #962, opened 2022-07-25
Attach test resources as volumes to save-agents' containers

petertrr:
* Introduce interface `PersistentVolumeService` to create volumes with test resources
* Implement `PersistentVolumeService` for Docker
* Mark save-cli executable from the agent itself
* Update properties for orchestrator in application-win.properties
* Move `KubernetesClient#close` from `@PreDestroy` to `@Bean(destroyMethod = )`
* Create `.env` file with agent id in directory that is not overridden when volume is mounted

Part of #924: establishes foundation and adds implementation for Docker. Implementation for Kubernetes will be in a separate PR.

PR #979, opened 2022-07-29
Fix path to the directory for temporary resources copying in save-orchestrator with Docker

petertrr:
* Copy resources for execution into a dedicated path on an intermediate volume, mount an intermediate volume when creating a volume for an execution
* Create an intermediate volume in tmpfs in `docker-compose.yaml`
* Change case of properties in save-orchestrator's application.properties

Actual value of `testResourcesVolumeName` should be set externally for a particular env, because both compose and swarm add a prefix to the volume name.
Test for `DockerService` are executed outside Docker, so we can't use intermediate volume mount. Same goes for local running as a plain JVM process. Using bind mount is supported for this purpose.

Follow-up after #962, for Kubernetes will be implemented as part of #977


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -199,7 +197,11 @@ class DockerService(
             configProperties.testResources.basePath,
             execution.resourcesRootPath!!,
         )
-        val resourcesForExecution = createTempDirectory(prefix = "save-execution-${execution.id}")
+        val resourcesForExecution = createTempDirectory(
+            directory = Paths.get(configProperties.testResources.tmpPath),
+            prefix = "save-execution-${execution.id}"
+        )
+        log.debug { "Copying resources from $originalResourcesPath into $resourcesForExecution" }
         originalResourcesPath.copyRecursively(resourcesForExecution.toFile())
```
nulls, line 205: btw, we create a tmp folder here and then create a new tmp folder in KubernetesPersistentVolumeService (which is not clear).
do we need two temp folders?

nulls (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
petertrr: I'll change this is #977. The logic will be like in Docker: create a tmp folder on a mount that can be shared with a temporary container, copy data from original storage to the tmp folder, copy from the tmp folder into a volume created for a certain execution,

PR #982, opened 2022-07-30
Frontend updates: welcome view and project view

akuleshov7:
### What's done:
- updated welcome view animation
- updated error and warning logic for project vie

https://user-images.githubusercontent.com/58667063/184942326-d6b0933f-84f9-4ac5-8620-0f0790cbd887.mov


sanyavertolet:
After I took a look at how different browsers (firefox, chrome and safari) draw alert or confirmation windows, I started to like an idea of moving to browser modal windows in some cases.

akuleshov7:
> After I took a look at how different browsers (firefox, chrome and safari) draw alert or confirmation windows, I started to like an idea of moving to browser modal windows in some cases.

I think we should give a try for it and later compare how it works

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -455,19 +424,6 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
         submitRequest("/submitExecutionRequest", Headers(), formData)
     }
 
-    private fun submitExecutionRequestByContest() {
```
akuleshov7, line 458: very strange resolving of conflicts by idea

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
akuleshov7: done

PR #985, opened 2022-08-01
Minor fix in top bar and UserSettingsView

Cheshiriks:
* fixed avatar in top bar
* removed download avatar in bd

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/TopBar.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/TopBar.kt
@@ -230,30 +230,31 @@ fun topBar() = FC<TopBarProps> { props ->
                     asDynamic()["data-toggle"] = "dropdown"
 
                     div {
-                        className = ClassName("row")
+                        className = ClassName("d-flex flex-row")
                         div {
+                            className = ClassName("d-flex flex-column")
                             span {
                                 className = ClassName("mr-2 d-none d-lg-inline text-gray-600")
                                 +(props.userInfo?.name ?: "")
                             }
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/TopBar.kt
```diff
@@ -230,30 +230,31 @@ fun topBar() = FC<TopBarProps> { props ->
...
                             }
-                            props.userInfo?.avatar?.let {
-                                img {
-                                    className =
-                                            ClassName("avatar avatar-user width-full border color-bg-default rounded-circle fas fa-lg fa-fw mr-2")
-                                    src = "/api/$v1/avatar$it"
-                                    height = 26.0
-                                    width = 26.0
+                            val globalRole = props.userInfo?.globalRole ?: Role.VIEWER
+                            if (globalRole.priority >= Role.ADMIN.priority) {
```
sanyavertolet, line 240: I guess there is a function for that

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/TopBar.kt
nulls:
```suggestion
                            if (globalRole.isHigherOrEqualThan(Role.ADMIN)) {
```

Cheshiriks: Fixed. Thank you!

petertrr (approved):

Issue #918, opened 2022-07-13
Prohibit some names for projects and organisations


akuleshov7:
In SAVE we should prohibit names of organisations that are equal to some USER name that is already there in the system.
Also we should prohibit project names, like some_organization/"settings"

sanyavertolet:
We might also want to forbid some other names. I think it might be possible to combine these checks at the same place.

sanyavertolet:
Done by #988 and #991

PR #988, opened 2022-08-02
General Validation: Origins

sanyavertolet:
This PR is connected to #918 

### What's done:
- [x] Added validations on frontend when creating `Organization`
- [x] Added validations on frontend when creating `Project`
- [x] Added interface `Validatable` which should be implemented by everything that needs validation
- [x] Added `ProjectDto`
- [x] Added `OrganizationDto`
- [x] Disabled "create" buttons when current DTO is not valid
- [x] Improved swagger docs for `ProjectController`
- [x] Improved swagger docs for `OrganizationController`

### How it looks like now:
screenshot
screenshot
screenshot
screenshot


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettingsview/UserSettingsOrganizationsMenuView.kt
```diff
@@ -37,22 +37,22 @@ class UserSettingsOrganizationsMenuView : UserSettingsView() {
                                 className = ClassName("align-items-center ml-3")
                                 img {
                                     className = ClassName("avatar avatar-user width-full border color-bg-default rounded-circle")
-                                    src = organizationInfo.avatar?.let {
+                                    src = organizationDto.avatar.let {
```
Cheshiriks, line 40: _organizationDto.avatar?.let {_

Cheshiriks (changes requested):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/ProjectDto.kt
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @property name name of a project
+ * @property organizationName name of an organization that is connected to the project
+ * @property description description of a project
+ * @property isPublic flag is project public or not
+ * @property url url of a project
+ * @property email email of a project
+ */
+@Serializable
+data class ProjectDto(
```
petertrr, line 20: `Project` should be moved to `jvmMain` and have `@Serializable` removed. Will you implement this?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/ValidationUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package com.saveourtool.save.validation
+
+/**
+ * Default amount of characters allowed for names
+ */
+const val NAMING_ALLOWED_LENGTH = 64
+
+/**
+ * Check if name is valid.
```
petertrr, line 13: Name for which entities?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -61,42 +56,24 @@ external interface OrganizationSaveViewState : State {
...
                         "$apiUrl/organization/save",
-                        headers,
-                        Json.encodeToString(newOrganizationRequest),
+                        jsonHeaders,
+                        Json.encodeToString(state.organizationDto),
                         loadingHandler = ::classLoadingHandler,
                     )
             if (responseFromCreationOrganization.ok) {
-                window.location.href = "${window.location.origin}#/${organizationName.replace(" ", "%20")}/"
+                window.location.href = "${window.location.origin}#/${state.organizationDto.name.replace(" ", "%20")}/"
```
petertrr, line 76: `String.replace(" ", "%20")` can be extracted to an Utils file

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -165,14 +129,22 @@ class CreateOrganizationView : AbstractView<Props, OrganizationSaveViewState>(tr
...
+                                            setState {
+                                                organizationDto = organizationDto.copy(name = it.target.value)
+                                            }
                                         }
                                     }
                                     button {
                                         type = ButtonType.button
                                         className = ClassName("btn btn-info mt-4 mr-3")
                                         +"Create organization"
+                                        disabled = !state.organizationDto.validate()
```
petertrr, line 147: Any hints for a user that input data is invalid?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/Validatable.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * [Validatable] should be implemented by all the classes where we need validation
+ */
+interface Validatable {
+    /**
+     * @return true if object is valid, false otherwise
+     */
+    @Suppress("FUNCTION_BOOLEAN_PREFIX")
+    fun validate(): Boolean = true
```
petertrr, line 15: On a second thought, this is quite poor API in terms of expressiveness. Consider `java.io` API: method `File.delete` returns `false` if file has not been deleted. Now consider `java.nio`: `Files.delete` throws an exception in case deletion was not successful with detailed reasons.

petertrr (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettingsview/UserSettingsOrganizationsMenuView.kt
sanyavertolet: sure, fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
sanyavertolet: As far as we forbade using whitespaces in names, we do not need it anymore. The only reason why it is still here is that we have in dev db examples with whitespaces. In next PR I will change `project-insert.xml` and `organization.csv` in order to get rid of those examples. In the same PR I will remove this replace

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/ValidationUtils.kt
sanyavertolet: For now just for three main entities: `Project`, `Organization` and `Contest`, but generally I guess we will have the same requirements for any names

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/ProjectDto.kt
sanyavertolet: Will do it later, now there are more important tasks

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
sanyavertolet: My idea is to show errors on certain fields, not on the button. Guess for button it is ok

sanyavertolet: For now those error messages are not that good, but later it will be fixed

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/ProjectDto.kt
petertrr: * #363

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/Validatable.kt
petertrr: Pls take a look at https://github.com/jquense/yup. Will this help us on the frontend?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
petertrr: So every time you change a field of the object, two validations are calculated - field validation and the whole object validation, right? Maybe, we can sort of `isFormValid` state that would be updated from inside `inputTextFormRequired` if any of fields is invalid? And then this value can be used to disable the button.

sanyavertolet: Decided to come back to this idea in phase 2 of this problem

Cheshiriks (approved):

petertrr (approved):

PR #989, opened 2022-08-03
Redirect missing 404 requests to a proper 404 page on the frontend

### LINKED ISSUE #912: Redirect missing 404 requests to a proper 404 page on the frontend

Cheshiriks:
- [x] - fixed redirect for project and organization view
- [ ] - topBar fixed


Cheshiriks:
I propose to make a merge without fixes with a TopBar for now

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
```diff
@@ -193,7 +198,19 @@ internal fun ComponentWithScope<*, *>.classComponentResponseHandler(
...
     if (hasResponseContext) {
-        this.withModalResponseHandler(response)
+        this.withModalResponseHandler(response, false)
+    }
+}
+
+/**
+ * @param response
+ */
+internal fun ComponentWithScope<*, *>.classComponentResponseRedirectHandler(
```
petertrr, line 208:
```suggestion
internal fun ComponentWithScope<*, *>.classComponentRedirectOnErrorResponseHandler(
```
Maybe like this?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
```diff
@@ -257,9 +274,13 @@ private suspend fun ComponentWithScope<*, *>.loadingHandler(request: suspend ()
 
 private fun ComponentWithScope<*, *>.withModalResponseHandler(
     response: Response,
+    isNeedRedirect: Boolean
 ) {
     if (!response.ok) {
         val statusContext: RequestStatusContext = this.asDynamic().context
+        if (isNeedRedirect) {
+            statusContext.setIsNeedRedirect(isNeedRedirect)
+        }
```
petertrr, line 283:
```suggestion
    isNeedRedirect: Boolean = false,
) {
    if (!response.ok) {
        val statusContext: RequestStatusContext = this.asDynamic().context
        statusContext.setIsNeedRedirect(isNeedRedirect)
```
It seems that with the current approach after `responseHandlerWithRedirect` gets called once, all consecutive `404`s will cause redirect, because the flag hasn't been unset.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
```diff
@@ -160,12 +165,24 @@ val requestModalHandler: FC<PropsWithChildren> = FC { props ->
...
+}
+
 /**
  * @property setResponse [StateSetter] for response error handler
  * @property setLoadingCounter [StateSetter] for active request counter
+ * @property setIsNeedRedirect
  */
 data class RequestStatusContext(
     val setResponse: StateSetter<Response?>,
+    val setIsNeedRedirect: StateSetter<Boolean>,
```
petertrr, line 185: Maybe call it `redirectToFallbackView` or `redirectToErrorPage`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
```diff
@@ -135,11 +138,13 @@ val requestModalHandler: FC<PropsWithChildren> = FC { props ->
         arrayOf(statusContext)
     ) { statusContext }
 
-    val reactNode = if (modalState.status == 404.toShort()) {
+    val reactNode = if (modalState.status == 404.toShort() && modalState.isNeedRedirect) {
```
petertrr, line 141: It seems a little strange, that the flag is set in the response handler if `!response.ok`, however, it's only used if status is 404 as well. It would be more clear if the logic would be in a single place. For example, set flag in response handler only if status is 404, and then, in the render method, if the flag is set, render FallbackView with text based on current status in the context.

petertrr:
> without fixes with a TopBar for now

Could you please remind, what are the problems with TopBar? I can't remember, and they're not mentioned in the issue, as far as I can see.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
Cheshiriks: Renamed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
Cheshiriks: Yes, agree. Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
Cheshiriks: Renamed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
Cheshiriks: Fixed

Cheshiriks:
> > without fixes with a TopBar for now
> 
> Could you please remind, what are the problems with TopBar? I can't remember, and they're not mentioned in the issue, as far as I can see.

After the redirect works, it will be impossible to go anywhere from the TopBar. Most likely because the HasErrorModal in the app.kt is higher than the Router and it does not work. I made a wrapper for a router with a location check, but the router itself no longer works this way.

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/HasErrorModal.kt
petertrr: Error status on the page is still always `404`, should we use status from context?

PR #1001, opened 2022-08-05
Single test suites source per execution

nulls:
Agreed with @akuleshov7 to avoid generating `save.toml` on orchestrator level and run test suites as is. It can be done only for a single test suites source

### What's done:
- supported case when agent sends empty result to backend
- added new endpoint to download test suites source by execution
- modified TestBatch: now it's only list of tests
- removed synthetic toml, will be handled properly in https://github.com/saveourtool/save-cli/issues/430


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestExecutionController.kt
```diff
@@ -226,7 +226,9 @@ class TestExecutionController(private val testExecutionService: TestExecutionSer
      */
     @PostMapping(value = ["/internal/saveTestResult"])
     fun saveTestResult(@RequestBody testExecutionsDto: List<TestExecutionDto>) = try {
-        if (testExecutionService.saveTestResult(testExecutionsDto).isEmpty()) {
+        if (testExecutionsDto.isEmpty()) {
```
kgevorkyan, line 229: But in which case they will sent empty results?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestExecutionController.kt
nulls: originally, I found it after issue in logic, but we have ability to exclude tests on test suite level. so we can select test suite with disabled tests and they can be run on a single agent (theoretically 😄 )

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -222,13 +220,8 @@ class DockerService(
...
     )
     private fun prepareImageAndVolumeForExecution(resourcesForExecution: Path, execution: Execution): RunConfiguration<PersistentVolumeId> {
-        // create stub toml config in aim to execute all test suites directories from `testSuitesDir`
-        resourcesForExecution.resolve(TEST_SUITES_DIR_NAME)
-            .resolve("save.toml")
-            .apply { createFile() }
-            .writeText(createSyntheticTomlConfig(execution.execCmd, execution.batchSizeForAnalyzer))
         // collect test suite names, which were selected by user
-        val saveCliExecFlags = " $TEST_SUITES_DIR_NAME --include-suites \"${execution.getTestSuiteNames()}\""
+        val saveCliExecFlags = " --include-suites \"${execution.getTestSuiteNames().joinToString(DATABASE_DELIMITER)}\" $TEST_SUITES_DIR_NAME"
```
nulls, line 224: the main problem -- `joinToString` wasn't called

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -328,6 +314,41 @@ class TestSuitesSourceController(
...
+    @ApiResponse(responseCode = "409", description = "IDs of test suites were not set on requested execution or weren't found by ID.")
+    @ApiResponse(responseCode = "404", description = "Test suites source with such name in organization name was not found.")
+    @ApiResponse(responseCode = "409", description = "Organization was not found by provided name.")
+    fun downloadByExecutionId(
+        @RequestParam executionId: Long
+    ): Mono<ByteBufferFluxResponse> = blockingToMono {
+        val execution = executionService.findExecution(executionId)
+            .orNotFound { "Execution (id=$executionId) not found" }
+        val testSuiteId = execution.parseAndGetTestSuiteIds()?.firstOrNull()
+            .orConflict { "Execution (id=$executionId) doesn't contain testSuiteIds" }
```
kgevorkyan, line 343: Why `conflict` and not `not found`?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -328,6 +314,41 @@ class TestSuitesSourceController(
         }
         .map { it.toDto() }
 
+    /**
+     * @param executionId
+     * @return selected [TestSuitesSourceDto] with versions for [com.saveourtool.save.entities.Execution] found by provided values
+     */
+    @PostMapping("/download-snapshot-by-execution-id", produces = [MediaType.APPLICATION_OCTET_STREAM_VALUE])
```
kgevorkyan, line 321: Btw, we decide to use swagger annotations, but it clogs up so much, as for me 👎

nulls: +1

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
nulls: I think it's a conflict situation, when we request download snapshot from execution which doesn't contain testSuiteIds. It means that we request this endpoint before execution is properly populated

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
kgevorkyan: lines 317-336 - blablabla
lines 337-350 - code

PR #1004, opened 2022-08-05
Removed download avatar in Organization view

Cheshiriks:
* removed download avatar in Organization view

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
@@ -228,14 +228,13 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
     override fun componentDidMount() {
         super.componentDidMount()
         scope.launch {
-            val avatar = getAvatar()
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -228,14 +228,13 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
         scope.launch {
-            val avatar = getAvatar()
             val organizationLoaded = getOrganization(props.organizationName)
             val projectsLoaded = getProjectsForOrganization()
             val role = getRoleInOrganization()
             val users = getUsers()
             setState {
-                image = avatar
                 organization = organizationLoaded
+                image = ImageInfo(organizationLoaded.avatar)
```
sanyavertolet, line 237: Why don't we store `ImageInfo` in `Organization`?

Cheshiriks: Because it is a wrapper and the database will not understand how to properly process it for writing.

petertrr (approved):

PR #1013, opened 2022-08-08
OrganizationSettingsMenu: Deletion of Git linked with TestSuiteSource

kgevorkyan:
### What's done:
* Also delete `foreign key constraints` fields, which depends on Git - test suite sources, test executions, tests


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -362,8 +367,17 @@ internal class OrganizationController(
...
-        .map {
-            gitService.delete(it, url)
+        .map { organization ->
+            // Find and remove all corresponding data to the current git repository
+            val git = gitService.getByOrganizationAndUrl(organization, url)
+            val testSuitesSources = testSuitesSourceService.findByGit(git)
+            testSuitesSources.forEach { testSuitesSource ->
+                val testSuites = testSuitesService.getBySource(testSuitesSource)
+                testSuitesService.deleteTestSuiteDto(testSuites.map { it.toDto() })
+                testSuitesSourceService.delete(testSuitesSource)
```
nulls, line 377: we also have `TestSuitesSourceStorage` which also should be cleaned after deleting `TestSuitesSource`

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -362,8 +367,17 @@ internal class OrganizationController(
...
         }
-        .map {
-            gitService.delete(it, url)
+        .map { organization ->
+            // Find and remove all corresponding data to the current git repository
+            val git = gitService.getByOrganizationAndUrl(organization, url)
+            val testSuitesSources = testSuitesSourceService.findByGit(git)
+            testSuitesSources.forEach { testSuitesSource ->
+                val testSuites = testSuitesService.getBySource(testSuitesSource)
+                testSuitesService.deleteTestSuiteDto(testSuites.map { it.toDto() })
```
nulls, line 376: Also, maybe check that there is no Execution in 'PENDING' state with TestSuites selected to be removed

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestExecutionRepository.kt
```diff
@@ -208,12 +208,12 @@ interface TestExecutionRepository : BaseEntityRepository<TestExecution>, JpaSpec
     @Transactional
     fun deleteAllByExecutionIdAndTestId(executionId: Long, testId: Long)
 
-    /** Returns a TestExecution matched by the test id
+    /** Returns a TestExecutions matched by the test id
      *
      * @param testId test id
      * @return Optional TestExecution
```
Cheshiriks, line 214: @return _List_ TestExecution**s**

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -363,11 +372,34 @@ internal class OrganizationController(
...
         }
 
+    private fun cleanupStorageData(testSuite: TestSuite) {
+        testSuitesSourceSnapshotStorage.findKey(
+            testSuite.source.organization.name,
+            testSuite.source.name,
+            testSuite.version,
+        ).flatMap { key ->
+            testSuitesSourceSnapshotStorage.delete(key)
+        }.subscribe()
```
nulls, line 400: why do you call subscribe here? maybe return `Mono<*>` and zip with Mono from method `deleteGit`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
kgevorkyan: Actually, we didn't remove executions at all, fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestExecutionRepository.kt
kgevorkyan: Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
kgevorkyan: Changed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -377,17 +377,26 @@ internal class OrganizationController(
...
+            val testSuitesList: MutableList<TestSuite?> = mutableListOf()
             testSuitesSources.forEach { testSuitesSource ->
                 val testSuites = testSuitesService.getBySource(testSuitesSource)
-                // Remove data from storage,
-                // since it is common for all test suites from one test suite source, it's enough to take any one of them
-                testSuites.firstOrNull()?.let {
-                    cleanupStorageData(it)
-                }
+                // Since storage data is common for all test suites from one test suite source, it's enough to take any one of them
+                testSuitesList.add(testSuites.firstOrNull())
```
nulls, line 385:
```suggestion
                testSuites.firstOrNull()?.let { testSuitesList.add(it) }
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -377,17 +377,26 @@ internal class OrganizationController(
             // Find and remove all corresponding data to the current git repository from DB and file system storage
             val git = gitService.getByOrganizationAndUrl(organization, url)
             val testSuitesSources = testSuitesSourceService.findByGit(git)
+            // List of test suites for removing data from storage at next step
+            val testSuitesList: MutableList<TestSuite?> = mutableListOf()
```
nulls, line 381: actually, we don't need `testSuite`, we need `TestSuitesSource` and version from `TestSuite`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
nulls: I'm about case when Execution is created with some test suites source but not started yet and then someone else remove TestSuitesSource.
I think it's rare case

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -377,17 +377,26 @@ internal class OrganizationController(
...
             testSuitesSources.forEach { testSuitesSource ->
```
nulls, line 382: mutable list can be avoided by returning result from `forEach` (by changing to `map`)

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
kgevorkyan: Yeah, changed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
kgevorkyan: We need `organization.name`, `TestSuitesSource.name` and `TestSuite.version`, I think it more convenient to pass `testSuite` instead of this tuple, since it contain all necessary fields

PR #1014, opened 2022-08-08
TestSuiteSelector

sanyavertolet:
This PR is a part of #250 
### What's done:
 - [x] Implemented TestSuiteSelector for TestSuite selection when creating contests
 - [x] Added several endpoints on backend for TestSuiteSelector
 - [x] Fixed a bug with BAD REQUEST when updating Project settings

### How contest creation looks like now:
https://user-images.githubusercontent.com/35039155/183642748-27fb6488-a6a0-4627-b078-0792472fe945.mov


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -56,7 +56,14 @@ class AgentsController(
     private val configProperties: ConfigProperties,
     @Qualifier("webClientBackend") private val webClientBackend: WebClient,
 ) {
-    private val tmpDir = Paths.get(configProperties.testResources.tmpPath).createDirectories()
+    // Somehow simple path.createDirectories() doesn't work on macOS, probably due to Apple File System features
```
petertrr, line 59: WDYM by `doesn't work`? Does it throw an exception?

sanyavertolet: Yes, especially for `Path("/tmp").createDirectories()` I got `FileAlreadyExistsException`

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuiteDto.kt
```diff
@@ -18,4 +18,16 @@ data class TestSuiteDto(
     val version: String,
     val language: String? = null,
     val tags: List<String>? = null,
-)
+) {
+    /**
+     * Identificator of a TestSuite for easier test suite management
+     */
+    var id: Long? = null
```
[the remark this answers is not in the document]
sanyavertolet: Yes, you are right, `equals()` compares only properties from primary constructor. Seems like it is better to move it to constructor parameter.

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
petertrr: According to the docs, `Files.createDirectories`, which is  actually used by the Kotlin extension, throws `FileAlreadyExistsException – if dir exists but is not a directory (optional specific exception)`. Maybe, tmpfs is implemented differently on Mac?

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
sanyavertolet: Probably it is so

PR #1015, opened 2022-08-08
Fix for additional files on Orchestrator

### LINKED ISSUE #1003: Backend: "/internal/saveTestResult" java.util.NoSuchElementException: List is empty.
Probably, connected with https://github.com/saveourtool/save-cloud/pull/1001

However, there are another warnings in `orchestrator`, which may cause this exeption:

```bash
[boundedElastic-1] WARN c.s.s.o.controller.AgentsController - Not found any tests for execution 8
[reactor-http-epoll-4] INFO c.s.s.o.controller.AgentsController - Downloaded all tests for execution 8 to /tmp/save-execution-84904926851769248940/test-suites
[reactor-http-epoll-4] WARN c.s.s.o.controller.AgentsController - Not found any additional files for execution 8
```

despite the fact, that additional files are provided and test repository is valid

nulls:
### What's done:
- fixed map(Mono)
- fixed logging in defaultIfEmpty
- additional logging

It closes #1003


save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/ReactorUtils.kt
```diff
@@ -94,14 +94,20 @@ fun <T> justOrNotFound(data: Optional<T>, message: String? = null) = Mono.justOr
     message
 }
 
+/**
+ * @param supplier it returns nullable [T]
+ * @return [Mono] from result or empty
+ */
+fun <T : Any> lazyToMono(supplier: () -> T?): Mono<T> = Mono.just(UInt)
```
petertrr, line 101: What is the difference with `Mono.fromCallable`?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -228,9 +230,16 @@ class AgentsController(
...
+        .lazyDefaultIfEmpty {
+            log.warn {
+                "Not found any tests for execution ${requiredId()}"
+            }
+        }
+
+    private fun Flux<DataBuffer>.writeTo(targetFile: Path): Mono<Path> =
+            DataBufferUtils.write(this, targetFile.outputStream())
+                .map { DataBufferUtils.release(it) }
+                .then(Mono.just(targetFile))
```
petertrr, line 242: Is there any advantage on wrapping the value instead of calling `thenReturn`?

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/ReactorUtils.kt
akuleshov7: Unit, not Uint :D

nulls: it checks on nullable. `Mono.fromCallable` expects only not null value. 
This method creates a mono or Mono.empty

akuleshov7 (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
nulls: funny, but Flux doesn't have `thenReturn` (I wanted to use it)

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/ReactorUtils.kt
petertrr: Quoting from docs,

> If the Callable resolves to null, the resulting Mono completes empty.

I believe, `Mono.fromCallable` extects a non-nullable *callable*, which *can* emit `null`. Or am I misreading it?

nulls: yep, you are right. confused by type resoling in idea in next map function (IDEA detects it as nullable).
actually, kotlin extension already has similar override. will re-use

PR #1016, opened 2022-08-08
Bugfix: Wrong error display when creating organization with existing name

sanyavertolet:
### What's done:
 * Changed error code sent from backend with conflicting names
 * Fixed NPE when unauthorized
 * Improved swagger doc for `LnkContestProjectController`


petertrr (approved):

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -121,10 +121,11 @@ internal class OrganizationController(
     @ApiResponse(responseCode = "200", description = "Successfully changed ability to create contests.")
     @ApiResponse(responseCode = "403", description = "Could not change ability to create contests due to lack of permission.")
     @ApiResponse(responseCode = "404", description = "Organization with such name was not found.")
+    @Suppress("UnsafeCallOnNullableType")
     fun setAbilityToCreateContest(
         @PathVariable organizationName: String,
         @RequestParam isAbleToCreateContests: Boolean,
-        authentication: Authentication
+        authentication: Authentication?,
```
petertrr, line 128: I don't think nullable parameter is needed here, because this method already has `PreAuthorize` which would filter out unauthorized requests

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -195,10 +196,11 @@ internal class OrganizationController(
     @ApiResponse(responseCode = "403", description = "Not enough permission for managing this organization.")
     @ApiResponse(responseCode = "404", description = "Could not find an organization with such name.")
     @ApiResponse(responseCode = "409", description = "Organization with such name already exists.")
+    @Suppress("UnsafeCallOnNullableType")
     fun updateOrganization(
         @PathVariable organizationName: String,
         @RequestBody organization: Organization,
-        authentication: Authentication,
+        authentication: Authentication?,
```
petertrr, line 203: Same here

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -135,7 +136,7 @@ internal class OrganizationController(
             "No organization with name $organizationName was found."
         }
         .filter {
-            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.SUPER_ADMIN)
+            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication!!, it.name, Role.SUPER_ADMIN)
```
nulls, line 139: can we property handle nullable here? we can use `Mono.zip(authentication.toMono())`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Oh sure, will fix all those moments

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
sanyavertolet: Will fix it

petertrr (approved):

PR #1020, opened 2022-08-10
Bugfix for 500 when adding user to group

### LINKED ISSUE #1018: 500 error when trying to add new user to organization

sanyavertolet:
This PR closes #1018 

### What's done:
- [x] Made all the used methods support reactivity
- [x] Reworked `LnkUserProjectController` to support reactivity
- [x] Improved swagger doc for `LnkUserProjectController`

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
@@ -292,21 +292,16 @@ class LnkUserOrganizationController(
         .map { users ->
             users.map { (user, _) -> user.requiredId() }.toSet()
         }
-        .flatMap { organizationUserIds ->
-            Mono.zip(
-                Mono.just(organizationUserIds),
-                lnkUserOrganizationService.getNonOrganizationUsersByName(prefix, organizationUserIds).toMono(),
-            )
+        .map { organizationUserIds ->
+            organizationUserIds to lnkUserOrganizationService.getNonOrganizationUsersByName(prefix, organizationUserIds)
         }
-        .flatMap { (organizationUserIds, exactMatchUsers) ->
-            Mono.zip(
-                Mono.just(exactMatchUsers),
-                lnkUserOrganizationService.getNonOrganizationUsersByNamePrefix(
-                    prefix,
-                    organizationUserIds + exactMatchUsers.map { it.requiredId() },
-                    PAGE_SIZE - exactMatchUsers.size,
-                ).toMono()
-            )
+        .map { (organizationUserIds, exactMatchUsers) ->
+            exactMatchUsers to
+                    lnkUserOrganizationService.getNonOrganizationUsersByNamePrefix(
+                        prefix,
+                        organizationUserIds + exactMatchUsers.map { it.requiredId() },
+                        PAGE_SIZE - exactMatchUsers.size,
+                    )
         }
         .map { (exactMatchUsers, prefixUsers) ->
             (exactMatchUsers + prefixUsers).map { it.toUserInfo() }
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
@@ -22,6 +22,7 @@ import com.saveourtool.save.entities.User
 import com.saveourtool.save.info.UserInfo
 import com.saveourtool.save.permission.Permission
 import com.saveourtool.save.permission.SetRoleRequest
+import com.saveourtool.save.utils.switchIfEmptyToNotFound
 import com.saveourtool.save.v1
 
 import io.swagger.v3.oas.annotations.Operation
@@ -31,7 +32,6 @@ import io.swagger.v3.oas.annotations.enums.ParameterIn
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
 import io.swagger.v3.oas.annotations.tags.Tags
-import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
@@ -41,6 +41,7 @@ import org.springframework.web.server.ResponseStatusException
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
 import reactor.kotlin.core.publisher.switchIfEmpty
+import reactor.kotlin.core.publisher.toMono
 import reactor.kotlin.core.util.function.component1
 import reactor.kotlin.core.util.function.component2
 import reactor.util.function.Tuple2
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -272,30 +271,47 @@ class LnkUserOrganizationController(
...
-        val exactMatchUsers = lnkUserOrganizationService.getNonOrganizationUsersByName(prefix, organizationUserIds)
-        // and then we need to get those whose names start with `prefix`
-        val prefixUsers = lnkUserOrganizationService.getNonOrganizationUsersByNamePrefix(
-            prefix,
-            organizationUserIds + exactMatchUsers.map { it.requiredId() },
-            PAGE_SIZE - exactMatchUsers.size,
-        )
-        return (exactMatchUsers + prefixUsers).map { it.toUserInfo() }
-    }
+    ): Mono<List<UserInfo>> = Mono.just(organizationName)
```
petertrr, line 279: `Flux<UserInfo>`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -272,30 +271,47 @@ class LnkUserOrganizationController(
...
+        }
+        .flatMap {
+            organizationService.findByName(it).toMono()
+        }
+        .switchIfEmptyToNotFound {
+            "No organization with name $organizationName was found."
+        }
+        .map { organization ->
+            lnkUserOrganizationService.getAllUsersAndRolesByOrganization(organization)
+        }
```
petertrr, line 291: Can be converted to Flux using `flatMapMany` and `Flux.fromIterable`

sanyavertolet: The point is that later we will need all the list we get here because we pass it later to `LnkUserOrganizationService`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
sanyavertolet: Well, it can be done so, but the final flux will be constructed only in the end of a method and I am not so sure if there is any difference

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
```diff
@@ -7,100 +7,156 @@
...
-        val project = projectService.findByNameAndOrganizationName(projectName, organizationName)
-            ?: throw ResponseStatusException(HttpStatus.NOT_FOUND)
-        val usersWithRoles = if (projectPermissionEvaluator.hasPermission(authentication, project, Permission.READ)) {
+    ): Mono<List<UserInfo>> = Mono.zip(
+        Mono.just(organizationName),
+        Mono.just(projectName),
+    )
+        .flatMap {
+            projectService.findByNameAndOrganizationName(projectName, organizationName).toMono()
+        }
```
petertrr, line 73: Why not start with simple `Mono.fromCallable { projectService... }`? You are not using parameters passed down from `zip`

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
petertrr: Ok, haven't noticed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
sanyavertolet: As far as I remember, @nulls considered it to be more readable, but maybe I got it wrong

sanyavertolet: Fixed it where suitable

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
```diff
@@ -7,100 +7,156 @@
...
+            "No project with name $projectName was found in organization $organizationName."
+        }
+        .map { project ->
+            lnkUserProjectService.getAllUsersByProject(project)
+        }
+        .map { users ->
+            users.map { it.requiredId() }.toSet()
+        }
+        .flatMap { projectUserIds ->
+            Mono.zip(
```
nulls, line 134: why do not call zip on original Mono instead of flatMap?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
```diff
@@ -7,100 +7,156 @@
...
+                Mono.just(projectUserIds),
+                lnkUserProjectService.getNonProjectUsersByName(prefix, projectUserIds).toMono()
+            )
+        }
+        .flatMap { (projectUserIds, exactMatchUsers) ->
```
nulls, line 139: why do not use `.map` with Pair as result?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
```diff
@@ -7,100 +7,156 @@
...
-        )
-        return (exactMatchUsers + prefixUsers).map { it.toUserInfo() }
-    }
+        .switchIfEmptyToNotFound {
+            "No project with name $projectName was found in organization $organizationName."
+        }
+        .filter { project ->
+            projectPermissionEvaluator.hasPermission(authentication, project, Permission.READ)
+        }
+        .switchIfEmptyToNotFound {
```
nulls, line 124: maybe forbidden instead of not found?

sanyavertolet: We decided that private projects should be invisible just as they don't exist

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
sanyavertolet: fixed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
sanyavertolet: removed almost all `flatMap`s

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserProjectController.kt
nulls: I suggest to use `.toMono` for nullable values

nulls (approved):

PR #1023, opened 2022-08-10
Fix pagination on ExecutionView

kgevorkyan:
### What's done:
* Fix pagination on ExecutionView

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
@@ -422,13 +422,18 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
                 }
             }
             getPageCount = { pageSize ->
-                val count: Int = post(
-                    url = "$apiUrl/test-executions?executionId=${props.executionId}&page=1&size=$pageSize&checkDebugInfo=true",
-                    headers = Headers().apply {
-                        set("Accept", "application/json")
-                        set("Content-Type", "application/json")
+                val status = filters.status?.let {
+                    "&status=${filters.status}"
+                } ?: ""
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
```diff
@@ -422,13 +422,18 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
...
-                    headers = Headers().apply {
-                        set("Accept", "application/json")
-                        set("Content-Type", "application/json")
+                val status = filters.status?.let {
+                    "&status=${filters.status}"
+                } ?: ""
+                val testSuite = filters.testSuite?.let {
+                    "&testSuite=${filters.testSuite}"
+                } ?: ""
+
```
petertrr, line 431: Maybe something like 
```kotlin
val filtersQueryString = buildString {
   filters.status?.let  { append(...) }
}
```
to make it easier to incorporate new filters should they be needed in the future?

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
```diff
@@ -439,15 +439,18 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
...
-                val status = filters.status?.let {
-                    "&status=${filters.status}"
-                } ?: ""
-                val testSuite = filters.testSuite?.let {
-                    "&testSuite=${filters.testSuite}"
-                } ?: ""
+                val filtersQueryString = buildString {
+                    filters.status?.let {
+                        append("&status=${filters.status}")
+                    } ?: append("")
```
nulls, line 445: not need to `append("")`

PR #1028, opened 2022-08-11
Trigger execution run by test suite ids

nulls:
### What's done:
- added a new endpoint to trigger execution run by test suites id without using preprocessor
- added selector for contest (for testing purpose)

It's part of #250

# Overview:
After TestSuitesSource is introduced, save-cloud has preloaded test suite ids. No need to load it during run.

As part of contest, we select test suite ids for contest.

## Changes:
- Added `com.saveourtool.save.backend.controllers.RunExecutionController` on backend. Now backend creates 'Execution' with all required fields and then sends a request to orchestrator
- Refactored current `com.saveourtool.save.backend.controllers.CloneRepositoryController` to use the new controller (create request by provided values and call controller)
- Refactored rerun: now save-cloud creates a copy of Execution and run it


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -1,57 +1,40 @@
...
-    private val projectService: ProjectService,
-    private val executionService: ExecutionService,
+    private val testSuitesService: TestSuitesService,
+    private val testSuitesSourceService: TestSuitesSourceService,
+    private val testSuitesSourceSnapshotStorage: TestSuitesSourceSnapshotStorage,
     private val fileStorage: FileStorage,
-    private val configProperties: ConfigProperties,
-    jackson2WebClientCustomizer: WebClientCustomizer,
+    private val gitService: GitService,
+    private val runExecutionController: RunExecutionController,
```
sanyavertolet, line 36: Never seen using controller inside another controller. Is it a good practice?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -95,72 +85,63 @@ class CloneRepositoryController(
...
+                    }
+                    .flatMapIterable { testSuitesService.getBySourceAndVersion(testSuitesSource, it) }
+                    .filter(testSuitesFilter)
+                    .map { it.requiredId() }
+            }
+            .let {
+                Flux.concat(it)
+            }
+            .collectList()
+        return fileStorage.convertToLatestFileInfo(projectCoordinates, shortFiles)
```
sanyavertolet, line 126: Is it a good idea to write reactive methods like this? Isn't reactivity broken here?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/RunExecutionController.kt
```diff
@@ -0,0 +1,193 @@
...
+                    asyncTrigger(execution)
+                }
+        }
+
+    /**
+     * @param executionId ID of [Execution] which needs to be copied for new execution
+     * @param authentication
+     * @return response with ID of new created [Execution]
+     */
+    @PostMapping("/reTrigger")
```
sanyavertolet, line 99: AFAIR, we want to have kebab-case in urls instead of camelCase

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -503,6 +504,29 @@ class TestSuitesSourceController(
...
+    /**
+     * @param organizationName
+     * @param name
+     * @param authentication
+     * @return response from preprocessor
+     */
+    @PostMapping("/api/$v1/test-suites-sources/{organizationName}/{name}/fetch")
+    fun triggerFetch(
+        @PathVariable organizationName: String,
+        @PathVariable name: String,
```
sanyavertolet, line 516: might be more readable to rename `name` to `sourceName` because we also have `{organizationName}/{projectName}` and it might be confusing

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ProjectService.kt
```diff
@@ -124,7 +124,7 @@ class ProjectService(
      * @param userName
      * @return optional of user
      */
-    fun findUserByName(userName: String): Optional<User> = userRepository.findByName(userName)
+    fun findUserByName(userName: String): User? = userRepository.findByName(userName)
```
sanyavertolet, line 127: Why saying no to `Optional`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
```diff
@@ -23,46 +25,48 @@ class UserDetailsService(
 ) : ReactiveUserDetailsService {
     private val logger = LoggerFactory.getLogger(javaClass)
 
-    override fun findByUsername(username: String): Mono<UserDetails> = Mono.fromCallable {
-        userRepository.findByName(username)
-    }.getIdentitySourceAwareUserDetails(username)
+    override fun findByUsername(username: String): Mono<UserDetails> =
+            { userRepository.findByName(username) }
```
sanyavertolet, line 29: Why is it wrapped into brackets?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -62,10 +62,35 @@ external interface ProjectExecutionRouteProps : PropsWithChildren {
     var currentUserInfo: UserInfo?
 }
 
+/**
+ * State with role
+ */
+external interface StateWithRole : State {
```
sanyavertolet, line 68: Why is it defined only here? I mean, it can be used all over the project

save-preprocessor/src/test/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectTest.kt
```diff
@@ -44,11 +44,12 @@ import java.time.Duration
 import java.time.LocalDateTime
 import java.util.concurrent.TimeUnit
 
-@WebFluxTest(controllers = [DownloadProjectController::class])
+// @WebFluxTest(controllers = [DownloadProjectController::class])
```
sanyavertolet, line 47: Will remove it later?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/RunExecutionController.kt
nulls: will change to re-trigger

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
nulls: ok, will rename

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ProjectService.kt
nulls: `?` is more kotlin way for working with nullable variable

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
nulls: it's extension for `Mono.fromCallable` for Kotlin -- this return `Mono<T>` instead of `Mono<T?>`

save-preprocessor/src/test/kotlin/com/saveourtool/save/preprocessor/controllers/DownloadProjectTest.kt
nulls: yeap

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
nulls: I tried to split a huge external inteface to several with fields are related to a single functionality.
And yes, it can be used in another places -- will extract to common file

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
nulls: I didn't get where is a problem. Can change it to extension function for `Flux<ShortFileInfo>` to make it more 'reactive' style

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
nulls: I don't see a technical problem here, but it's not good for architecture (better avoid BL in controllers).
But it was added for transition period to leave old endpoints -- will be removed after implementing selection of IDs of TestSuite instead of current logic

sanyavertolet: Oh ok

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ProjectService.kt
sanyavertolet: Always thought `Optional`s is something from async world, but if there is no difference, will use `?` More often, thanks

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
nulls: added `uploadedMillis` to `ShortFileInfo` and remove `convertToLatestFileInfo`

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ProjectService.kt
nulls: I changed only touch method in current PR

nulls:
Current limitations:
- GIT run accepts only branch for now. Logic was moved from preprocessor to backend and it can't detect a default branch or latest version in git repository

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -503,6 +504,37 @@ class TestSuitesSourceController(
         authentication: Authentication,
     ): Mono<List<String>> = testSuitesSourceService.getOrganizationsWithPublicTestSuiteSources().toMono()
 
+    /**
```
petertrr, line 507: Does diktat still add these KDocs? We've added several OpenApi annotations to `ignoreAnnotated` in `diktat.yml`, but probably haven't covered them all..

save-backend/src/test/kotlin/com/saveourtool/save/backend/controllers/RunExecutionControllerTest.kt
```diff
@@ -0,0 +1,197 @@
...
+        fun properties(registry: DynamicPropertyRegistry) {
+            mockServerOrchestrator = createMockWebServer()
+            mockServerOrchestrator.start()
+            registry.add("backend.orchestratorUrl") { "http://localhost:${mockServerOrchestrator.port}" }
+        }
+    }
+
+    @TestConfiguration
+    class AdditionalConfiguration {
+        @Bean fun meterRegistry() = SimpleMeterRegistry()
```
petertrr, line 195: For tests it can be initialized with a `CompositeMeterRegistry`, which will wrap an empty list of registries hence acting effectively no-op

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
```diff
@@ -204,6 +228,20 @@ private fun prepareTestSuitesSourcesTable(
...
+        column(id = "fetch", header = "Fetch new version", { this }) { cellProps ->
+            Fragment.create {
+                td {
+                    button {
+                        type = ButtonType.button
+                        className = ClassName("btn btn-sm btn-primary")
+                        onClick = {
+                            fetchHandler(cellProps.value)
+                        }
+                        +"fetch"
```
petertrr, line 240: Fetch will start in the backgroud, right? Maybe we should display some kind of confirmation that request is accepted and the version will be updated later?

petertrr (approved):

save-backend/src/test/kotlin/com/saveourtool/save/backend/controllers/RunExecutionControllerTest.kt
nulls: thanks, will change

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
nulls: actually, I added it manually. but `ignoreAnnotated` explains why I don't see errors on another methods 😄 
will remove kdoc

PR #1029, opened 2022-08-11
Removing of additional files should require confirmation

### LINKED ISSUE #986: Removing of additional files should require confirmation
Now it's kind of easy to misclick the button and accidentally remove the file

Capture

Cheshiriks:
* added confirmation for removing file
delete

closes #986


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -734,30 +741,40 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
         }
     }
 
-    private fun postFileDelete(file: FileInfo) {
+    private fun fileDelete() {
         scope.launch {
             val headers = Headers().also {
                 it.set("Accept", "application/json")
                 it.set("Content-Type", "application/json")
```
sanyavertolet, line 748: `jsonHeaders`

sanyavertolet (commented):
I strongly believe that modals that we draw ourselves now look much worse that `window.alert()` does. Either try to make it a bit nicer with `bootstrap`, or try using `window.alert()`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
Cheshiriks: fixed

sanyavertolet (approved):
Looks good, but we need to continue discussion about confirmation modal windows with team

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -734,30 +741,35 @@ class ProjectView : AbstractView<ProjectExecutionRouteProps, ProjectViewState>(f
...
-            val headers = Headers().also {
-                it.set("Accept", "application/json")
-                it.set("Content-Type", "application/json")
-            }
-
             val response = delete(
-                "$apiUrl/files/${props.owner}/${props.name}/${file.uploadedMillis}",
-                headers,
-                Json.encodeToString(file),
+                "$apiUrl/files/${props.owner}/${props.name}/${state.file?.uploadedMillis}",
```
petertrr, line 747: Does it make sense to have URL like `/saveourtool/diktat/null` here? I think you can save `state.file` to a local varialbe with nullability check, because if `state.file == null` this request won't make sense

Cheshiriks:
> Looks good, but we need to continue discussion about confirmation modal windows with team

I agree. But I think it would be better to change it everywhere later.

petertrr:
> I strongly believe that modals that we draw ourselves now look much worse that `window.alert()` does. Either try to make it a bit nicer with `bootstrap`, or try using `window.alert()`

I think we definitely should be improving our styles rather than using `alert` and `confirm`. Those things are too browser-dependent

petertrr (approved):

Issue #728, opened 2022-05-05
Bug with containers creation in Orchestrator on WSL


kgevorkyan:
After https://github.com/analysis-dev/save-cloud/pull/707 there is a problems with `getHostIp()` method, since `host.docker.internal` on WSL are already exists in `/etc/hosts` for some reasons, therefore, in process of containers creation. instead of  `host-gateway` value we get  `host.docker.internal`, which is wrong

petertrr:
IIUC, `host.docker.internal` on WSL distros points to the Windows hosts, so that processes from inside Docker on WSL distro can access processes running on Windows host and not on WSL VM. This can probably be worked around by passing `HOST_IP=host-gateway` when running orchestrator with `bootRun` inside WSL.

I believe if orchestrator runs on Windows and communicates with Docker Daemon via HTTP API (see #632), this very feature makes possible communication between agents running inside Docker for Windows and orchestrator. It's weird though, that there is a `hosts` entry even in usual distros.

nulls:
We stopped build containers on orchestrator level

PR #1030, opened 2022-08-11
Build base images on CI instead of orchestrator

### LINKED ISSUE #1025: Build base images outside of save-orchestrator's lifecycle
Base images don't change often, so we can build them once a week on Github actions. Also, there are few of them, so we can publish them to ghcr.io.

After that orchestrator won't need Docker for normal operation, and in Kubernetes mode we will be able to use containerd or other OCI engines.

petertrr:
* Added a new Dockerfile and workflow `build-base-images.yaml`
* Removed logic from code in orchestrator (`DockerClient` is used only in Docker mode) and backend (periodic job for building base images)
* Add Java 17 and Python 3.10 to the list of SDKs
* What's removed from orchestrator:
  * Build base images - now image from ghcr.io should be used
  * Images cleanup - no images are built; pulled images should be cleaned up during periodic `docker prune`
  * Docker client is no more used in `kubernetes` profile

Closes #1025


.github/workflows/build-base-images.yaml
```diff
@@ -0,0 +1,73 @@
...
+          - base_image_name: opendjk
+            base_image_tag: 10
+          - base_image_name: opendjk
+            base_image_tag: 11
+          - base_image_name: opendjk
+            base_image_tag: 12
+          - base_image_name: opendjk
+            base_image_tag: 13
+          - base_image_name: opendjk
+            base_image_tag: 14
```
petertrr, line 30: Only half of those images can be used directly, others probably have different tag format. Anyway, we may want to support only several LTS versions
https://github.com/saveourtool/save-cloud/pkgs/container/save-base/versions
https://github.com/saveourtool/save-cloud/actions/runs/2839233453

sanyavertolet (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManager.kt
```diff
@@ -1,112 +0,0 @@
...
-        "LongParameterList",
-        "TOO_MANY_PARAMETERS"
-    )
-    internal fun buildImage(baseImage: String = Sdk.Default.toString(),
-                            imageName: String,
-                            runCmd: String = "RUN /bin/bash",
-    ): String {
-        val tmpDir = createTempDirectory().toFile()
-        val dockerFile = createDockerFile(tmpDir, baseImage, runCmd)
-        val hostIp = getHostIp()
```
kgevorkyan, line 52: https://github.com/saveourtool/save-cloud/issues/728 could be closed?

save-deploy/base-images/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM $BASE_IMAGE_NAME:$BASE_IMAGE_TAG
+
+RUN apt-get update && env DEBIAN_FRONTEND="noninteractive" apt-get install -y libcurl4-openssl-dev tzdata
+RUN ln -fs /usr/share/zoneinfo/UTC /etc/localtime
+RUN rm -rf /var/lib/apt/lists/*
+
+# Fixme: https://github.com/saveourtool/save-cloud/issues/352
+RUN if [ "$BASE_IMAGE_NAME" = "python" ]; then \
+    curl -s "https://get.sdkman.io" | bash && \
+    bash -c 'source "$HOME/.sdkman/bin/sdkman-init.sh" && sdk install java 8.0.302-open' && \
```
kgevorkyan, line 12: Why java 8?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManager.kt
petertrr: No, host IP is still required for containers and this mapping is still being added when creating containers

save-deploy/base-images/Dockerfile
kgevorkyan: We used this version previously too, ok

kgevorkyan (approved):

save-orchestrator/src/test/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManagerTest.kt
```diff
@@ -30,27 +30,24 @@ import kotlin.io.path.createTempFile
...
     private lateinit var testImageId: String
 
     @BeforeEach
     fun setUp() {
-        dockerContainerManager = DockerContainerManager(configProperties, CompositeMeterRegistry(), dockerClient)
-        dockerClient.pullImageCmd("ubuntu")
-            .withTag("latest")
+        dockerClient.pullImageCmd("ghcr.io/saveourtool/save-base")
+            .withRegistry("https://ghcr.io")
+            .withTag("openjdk-11")
```
nulls, line 43: old value

PR #1033, opened 2022-08-11
Make `orchestrator.docker` properties optional

### LINKED ISSUE #1031: Use `alpine` variant of `eclipse-temurin` for base images with JDK
Docker images of `eclispe-temurin` have two variants: Ubuntu-based (e.g. `17`) and Alpine-based (e.g. `17-alpine`). Switching to Alpine can spare us some time on downloads, because its about 100MB smaller.

petertrr:
* Make `orchestrator.docker` properties optional and not used in Kubernetes
* Attempt to enable lock file maintenance in `renovate.json`

Can be considered another small follow-up of #1030

### DIFF
--- a/.github/workflows/build-base-images.yaml
+++ b/.github/workflows/build-base-images.yaml
@@ -15,11 +15,11 @@ jobs:
         # This should replicate list of SDKs from com.saveourtool.save.domain.Sdk
         include:
           - base_image_name: eclipse-temurin
-            base_image_tag: 8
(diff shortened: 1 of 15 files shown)


petertrr:
I've reverted changes with alpine-based images, because we are currently building executables of save-agent and save-cli linked with glibc on Ubuntu. So they are unable to start on alpine.

kgevorkyan (approved):

PR #1043, opened 2022-08-15
ExecutionView: Incorrect navigation to DebugInfo view

kgevorkyan:
### What's done:
* Fix incorrect navigation to DebugInfo view
* Fix flacky pagination in ExecutionView


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestStatusComponent.kt
```diff
@@ -66,7 +66,10 @@ fun <D : Any> testStatusComponent(testResultDebugInfo: TestResultDebugInfo, tabl
             colSpan = 2
             +"Reason ("
             a {
-                href = "${window.location}/details/$testSuiteName/$pluginName/$testFilePath"
+                // Trim location until the value of number of execution, since it could contain filters at the end, like `?status=PASSED`,
+                // when user got to this page by clicking corresponding table column on history view
+                val baseLocation = window.location.toString().dropLastWhile { !it.isDigit() }
```
petertrr, line 71: I think we can safely drop substring after `?`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
```diff
@@ -126,7 +126,15 @@ fun <D : Any, P : TableProps<D>> tableComponent(
...
-    useEffect(tableInstance.state.pageSize) {
+    // list of entities, updates of which will cause update of the data retrieving effect
+    val dependencies: Array<dynamic> = if (useServerPaging) {
+        arrayOf(tableInstance.state.pageIndex, tableInstance.state.pageSize, pageCount)
+    } else {
+        // when all data is already available, we don't need to repeat `getData` calls
+        emptyArray()
+    } + getAdditionalDependencies(props)
+
+    useEffect(*dependencies) {
```
petertrr, line 137: Did you figure out exactly which values from dependencies were required? `pageCount`? Or others too?

kgevorkyan: All of them 😐

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestStatusComponent.kt
kgevorkyan: I thought about it, but may be it is more safe way, independent of any changes in our API in future

petertrr: I think `?`, which is a dedicated delimiter in the general URL format, is more independent. Relying on digits in the link is in fact bound to changes of our API

kgevorkyan: Ok, changed

petertrr (approved):

PR #1045, opened 2022-08-15
TestSuiteSourceCreationComponent

sanyavertolet:
### What's done:
- [x] Added TestSuiteSourceCreationComponent
- [x] Added endpoint to save new TestSuitesSource
- [x] Improved SelectForm
- [x] Improved ValidationRegularExpressions

### How it looks:
https://user-images.githubusercontent.com/35039155/184871526-fe0aa342-4537-4da2-8bb6-7836e4cd2171.mov


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -352,6 +349,36 @@ class TestSuitesSourceController(
...
+    ): Mono<StringResponse> = getOrganization(organizationName)
+        .zipWhen { organization ->
+            gitService.findByOrganizationAndUrl(organization, testSuiteRequest.gitDto.url)
+                .toMono()
+                .switchIfEmptyToResponseException(HttpStatus.CONFLICT) {
+                    "There is no git credential with url ${testSuiteRequest.gitDto.url} in $organizationName"
+                }
+        }
+        .map { (organization, git) ->
+            testSuitesSourceService.getOrCreate(organization, git, testSuiteRequest.testRootPath, testSuiteRequest.branch)
```
nulls, line 378: can we create a separate method to create testSuitesSource by DTO entity?
current method `getOrCreate` generates `name` automatically as url to github. User can input something specific

sanyavertolet: Sure, will do it

sanyavertolet: Done

sanyavertolet:
The point is that now only entity in database is created, no preprocessing is run.
@nulls need your help here

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -352,6 +350,35 @@ class TestSuitesSourceController(
...
+    @RequiresAuthorizationSourceHeader
+    @PreAuthorize("permitAll()")
+    @Operation(
+        method = "POST",
+        summary = "Get or create a new test suite source by provided values.",
+        description = "Get or create a new test suite source by provided values.",
+    )
+    @ApiResponse(responseCode = "200", description = "Successfully get or create test suites source with requested values.")
+    @ApiResponse(responseCode = "404", description = "Organization was not found by provided name.")
+    @ApiResponse(responseCode = "404", description = "Git credentials were not found by provided url.")
```
petertrr, line 363: Should these be merged? Or does swagger merge them?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    }
+}
+
+@Suppress("TOO_LONG_FUNCTION", "LongMethod")
+private fun testSuiteSourceCreationComponent() = FC<TestSuiteSourceCreationProps> { props ->
+    val (testSuiteSource, setTestSuiteSource) = useState(TestSuitesSourceDto.empty.copy(organizationName = props.organizationName))
+    val (saveStatus, setSaveStatus) = useState<SourceSaveStatus?>(null)
+    val onSubmitButtonPressed = useRequest(dependencies = arrayOf(testSuiteSource)) {
```
petertrr, line 106: Does deferred request need dependencies?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreationView.kt
```diff
@@ -76,6 +76,8 @@ external interface ProjectSaveViewState : State {
 @JsExport
 @OptIn(ExperimentalJsExport::class)
 class CreationView : AbstractView<Props, ProjectSaveViewState>(true) {
+    @Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
+    private val selectFormRequired = selectFormRequired<String>()
```
petertrr, line 80:
```suggestion
    private val organzationSelectForm= selectFormRequired<String>()
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
```diff
@@ -110,7 +101,17 @@ private fun organizationTestsMenu() = FC<OrganizationTestsMenuProps> { props ->
             fetchTestSuitesSourcesSnapshotKeys()
         }
     }
-
+    showTestSuiteSourceCreationModal(
+        isTestSuiteSourceCreationModalOpen,
+        props.organizationName,
+        {
+            setIsTestSuitesSourceCreationModalOpen(false)
+            setIsSourceCreated { !it }
```
nulls, line 109: do we need `isTestSuiteSourceCreated` as additional dependency if we call `fetchTestSuitesSources` which updates `content`?

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
sanyavertolet: Oh you are right, they are not merged. Will fix it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
sanyavertolet: As far as there were `dependencies = arrayOf(props.organizationName)`, no request was performed. I decided to add extra dependency - `isTestSuiteSourceCreated`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
sanyavertolet: Oh, from the first glance it seemed to work as planned - sending request only if dependencies are changed, but not I am not so sure. Seems there is no need in it.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
sanyavertolet: @nulls used it in the same way, seems like it needs to be changed.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
petertrr: Now that I look at `useRequest` I don't understand why we need dependencies here:
https://github.com/saveourtool/save-cloud/blob/b665d1d8e54b3ec2ed5607d2ee38f8241db61f89/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt#L333

It makes sense to only use `isSending` when request is deferred and needs to only be activated by a trigger, e.g. a button click. And to use `dependencies` only in the effect, which is returned if `isDeferred  == false`:
https://github.com/saveourtool/save-cloud/blob/b665d1d8e54b3ec2ed5607d2ee38f8241db61f89/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt#L354-L358

Seems like now change of one of dependencies causes many invocations of the first effect, which are effectively negated by `isSending` being false

sanyavertolet: It also seems a bit weird that there actually are lots of `useEffect` calls (on line 333 and so on). Is there any other way to escape this cause it seems it can cause bit problems with efficiency.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
sanyavertolet: Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
nulls: Yep, but now there is a method `fetch` in testSuitesSourceService

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
nulls: But when we update `isTestSuiteSourceCreated`, we trigger `fetchTestSuitesSources` which updates `contest` which updates the table. Or I miss something?

sanyavertolet: Yep, this is what happens, afaiu.

nulls: so, it means that `isTestSuiteSourceCreated` is not needed as dependency in table because it will be re-rendered after `contest` is updated

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
petertrr: `useEffect` is used for callable request too, because it allows us to control the CoroutineScope and cancel it when component unmounts. But indeed it is not necessary to wrap one effect into another when a request is not deferred.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
sanyavertolet: You are right, will remove it.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
sanyavertolet: The fun thing is that if there is only `props.organizationName` in dependencies and I invoke `fetchTestSuitesSources()`, table in not updated with a new created source. If I add a trigger-flag into dependencies of useRequest, then I do not even need to invoke `fetchTestSuitesSources()`, which is fun. Will remove invocation and just leave flag

petertrr: There is no need in dependencies if `isDeferred` is true, because `useRequest` contains a trigger flag inside

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
sanyavertolet: Fun thing is that nothing happens if I call `fetchTestSuitesSources()` after source is created

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationTestsMenu.kt
sanyavertolet: Done

nulls (approved):

PR #1050, opened 2022-08-16
No TestSuite is selected in Git mode after deleting the project and creating the same one

kgevorkyan:
### What's done:
* Clean up corresponding test suite source, if data in DB is absent


save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/TestSuitesSourceSnapshotStorage.kt
```diff
@@ -64,6 +64,23 @@ class TestSuitesSourceSnapshotStorage(
...
+     */
+    fun removeKey(
+        organizationName: String,
+        testSuitesSourceName: String,
+        version: String,
+    ): Mono<Boolean> = findKey(organizationName, testSuitesSourceName, version)
+        .flatMap { key ->
+            delete(key).map { isRemoved ->
+                isRemoved
+            }
```
petertrr, line 81:
```suggestion
            delete(key)
```
Is this `.map` needed?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
```diff
@@ -112,7 +114,21 @@ class CloneRepositoryController(
                         keys.maxByOrNull(TestSuitesSourceSnapshotKey::creationTimeInMills)?.version
                             ?: throw IllegalStateException("Failed to detect latest version for $testSuitesSource")
                     }
-                    .flatMapIterable { testSuitesService.getBySourceAndVersion(testSuitesSource, it) }
+                    .flatMap { version ->
+                        val testSuiteList = testSuitesService.getBySourceAndVersion(testSuitesSource, version)
+                        // In some cases of contradictions, when DB is empty, but storage is not, cleanup data from storage
+                        if (testSuiteList.isEmpty()) {
+                            log.error("Test suites for test suite source ${testSuitesSource.name} not found in DB! Require fetch for this test suite source")
+                            testSuitesSourceSnapshotStorage.removeKey(testSuitesSource.organization.name, testSuitesSource.name, version)
```
petertrr, line 122: Should it be considered a quick temporary fix until we improve operations with fetch on the frontend? We will still get internal server error, but at least only for the first invocation, am I reading this correctly?

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/TestSuitesSourceSnapshotStorage.kt
kgevorkyan: Thanks, removed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CloneRepositoryController.kt
kgevorkyan: Basically, this error could appear is quite rare cases - when we manually drop db and forgot to do the same with storage.

Yes, we will still get exception, but only once.
However, I'm not sure, that decision, that we checking for existence of data in two different places is a good decision in general case. 
So, because of our current architecture, this error will be at least more clear

petertrr: > I'm not sure, that decision, that we checking for existence of data in two different places is a good decision in general case.

That's true, but that's just a consequence of the fact that for different types of data different types of storage are preferred.

> this error could appear is quite rare cases

And it sound like it never should appear on the real environment, where we don't usually drop either DB or storage. That's why I'm not sure if this check should exist in production code. Maybe, it's better to introduce a better cleanup task for local environment? The one that would not only call `liquibaseDropAll`, but also delete all data under `$HOME/.save-cloud`?

kgevorkyan: > That's true, but that's just a consequence of the fact that for different types of data different types of storage are preferred.
Probably we will need to talk about this with @nulls and do some refactoring in future

Anyway, left the warning message and created task for cleanup db & storage

petertrr (approved):

Issue #1054, opened 2022-08-16
kafka dev environment


icemachined:

akuleshov7:
❤️

petertrr:
Seems to be closed by #1055

PR #1055, opened 2022-08-16
kafka dev cluster configuration and controller #1054

icemachined:
### What's done:
* doker-compose kafka and zookeper services
* Created startKafka gradle task
* Test Kafka Configuration to send messages to configured request topic 
* Listen to configured result topic
* everything works only if both dev and kafka profiles are set


buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -138,24 +163,51 @@ fun Project.createStackDeployTask(profile: String) {
...
+    tasks.register<Exec>("startKafka") {
+        dependsOn("generateComposeFile")
+        doFirst {
+            logger.lifecycle("Running the following command: [docker-compose --file $buildDir/docker-compose.yaml up -d kafka]")
+        }
+        errorOutput = ByteArrayOutputStream()
+        commandLine("docker-compose", "--file", "$buildDir/docker-compose.yaml", "up", "-d", "kafka")
+        doLast {
+            logger.lifecycle("Waiting $KAFKA_STARTUP_DELAY_MILLIS millis for kafka to start")
+            Thread.sleep(KAFKA_STARTUP_DELAY_MILLIS)  // wait for kafka to start, can be manually increased when needed
```
akuleshov7, line 181: this is for local start?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/kafka/KafkaMsg.kt
```diff
@@ -0,0 +1,11 @@
+package com.saveourtool.save.kafka
+
+/**
+ * kafka message
+ */
+interface KafkaMsg {
+    /**
+     * @property messageId
+     */
+    val messageId: String?
```
akuleshov7, line 10: Id string? why not int?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
```diff
@@ -0,0 +1,164 @@
...
+import org.springframework.kafka.support.serializer.JsonSerializer
+import java.util.*
+
+/**
+ * Kafka producer and consumer configuration
+ *
+ * @property kafkaProperties
+ */
+@Configuration
+@Profile("dev & kafka")
```
akuleshov7, line 34: what is profile "kafka" for?

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
icemachined: yes, similar to mysql

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/kafka/KafkaMsg.kt
icemachined: currently it is not used, this is just the most general case

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
icemachined: kafka is for kafka specific settings like bootstrap servers, topics etc, it can be quite a lot so it's better to keep it in separate file

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
```diff
@@ -0,0 +1,164 @@
...
+    lateinit var consumerGroup: String
+
+    /**
+     * @return kafka producer properties
+     */
+    @Bean
+    fun producerConfig(): Map<String, Any?> {
+        val producerProps = HashMap(kafkaProperties.buildProducerProperties())
+        producerProps[ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG] = StringSerializer::class.java as Object
+        producerProps[ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG] = JsonSerializer::class.java as Object
```
petertrr, line 68: Currently we are using `kotlinx.serialization` on both agent and orchestrator to ensure that they exchange data in the same format (e.g. Jackson and kx.serialization have different field for type information in case of polymorphic serialization, may serialize DateTime differently etc.). As I understarnd, `org.springframework.kafka.support.serializer.JsonSerializer` uses Jackson. Can we - and will it be worth it - use kx.serialization here?

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -138,24 +163,51 @@ fun Project.createStackDeployTask(profile: String) {
         commandLine("docker-compose", "--file", "$buildDir/docker-compose.yaml", "up", "-d", "mysql")
         errorOutput = ByteArrayOutputStream()
         doLast {
-            if (!errorOutput.toString().contains(" is up-to-date")) {
```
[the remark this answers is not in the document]
icemachined: because error output doesn't contain this string, it only report that container started, even in the logs there is no such string,
i tried to get log output but failed. seems like this is separate task

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
icemachined: yes but it will be separate task, on save-agent we should use only kx.serialization in fact. but it would be cool it it will be compartible with spring json serializer because it is widely used in the community, it is very flexible serializer.

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
```diff
@@ -0,0 +1,164 @@
...
+     * @return kafka template
+     */
+    @Bean
+    fun kafkaTemplate(): KafkaTemplate<Any, Any> = KafkaTemplate(producerFactory())
+
+    /**
+     * @return kafka consumer properties
+     */
+    @Bean
+    fun consumerConfig(): Map<String, Any?> {
```
nulls, line 89: maybe move this property to property file and use `@ConfigurationProperties(prefix = "kafka.consumer")`?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
```diff
@@ -0,0 +1,164 @@
...
+        )
+        return factory
+    }
+
+    /**
+     * @return kafka agent tasks topic bean
+     */
+    @Bean
+    fun saveAgentTasks(): NewTopic {
+        log.info("Create topic: {} of {} partitions with {} replicas.", requestTopicName)
```
nulls, line 133: it's java style for logging. kotlin way (as I got) is
```suggestion
log.info { "Create topic: $requestTopicName of $missed partitions with $missed replicas." }
```

akuleshov7 (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
icemachined: fixed

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/config/KafkaConfiguration.kt
icemachined: in properties it is not convenient to set full class names, also not good to expose properties which is not supposed to change without changing business logic. also you already can set spring kafka properties in application-kafka.properties file which will be loaded in kafkaProperties

gradle/libs.versions.toml
```diff
@@ -12,6 +12,8 @@ spring-boot = "2.7.2"
 spring-security = "5.7.2"
 spring-cloud = "3.1.3"
 spring-cloud-kubernetes = "2.1.3"
+spring-kafka = "2.8.7"
```
nulls, line 15: we use `spring-boot-dependencies` and it contains version for kafka and spring-kafka

icemachined: fixed

gradle/libs.versions.toml
```diff
@@ -92,9 +93,14 @@ spring-cloud-starter-gateway = { module = "org.springframework.cloud:spring-clou
 spring-cloud-starter-kubernetes-client-config = { module = "org.springframework.cloud:spring-cloud-starter-kubernetes-client-config", version.ref = "spring-cloud-kubernetes" }
 spring-boot-starter-oauth2-client = { module = "org.springframework.boot:spring-boot-starter-oauth2-client" }
 spring-context-indexer = { module = "org.springframework:spring-context-indexer", version.ref = "spring" }
+spring-kafka = {module = "org.springframework.kafka:spring-kafka"}
+spring-kafka-test = {module = "org.springframework.kafka:spring-kafka-test"}
 spring-web = { module = "org.springframework:spring-web", version.ref = "spring" }
 jackson-module-kotlin = { module = "com.fasterxml.jackson.module:jackson-module-kotlin" }
 
+kafka-clients = {module = "org.apache.kafka:kafka-clients", version.ref = "kafka-client"}
```
nulls, line 101: these both versions also presented in [spring-boot-dependencies](https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/2.7.3/spring-boot-dependencies-2.7.3.pom)

icemachined: fixed

PR #1059, opened 2022-08-18
Minor fixes on frontend

### LINKED ISSUE #1057: `Create test project` button is unavailable before changing the organization
`Create test project` button is unavailable, even if all necessary fields are filled, until the organization will be changed.

Capture

sanyavertolet:
This PR closes #1057

### What's done:
 - [x] Fixed a bug with select form in `CreationView` - props were not set correctly so that the form didn't work fine
 - [x] Fixed a bug with `500` when `eligible-projects` endpoint is called - `eligible-projects` endpoint used to fetch all the projects (even deleted) as well as we used to allow Organization to be nullable implicitly - now only created projects are fetched by the endpoint  and organization is defiantly non-nullable
 - [x] Improved styles of `ContestEnroller`
 - [x] Added new style for react modals in order to provide better compatibility with bootstrap modals
 
### How it looks:
2022-08-18 в 12 51 47
2022-08-18 в 12 52 18

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
@@ -114,7 +114,7 @@ class LnkContestProjectController(
         @PathVariable contestName: String,
         authentication: Authentication,
     ): Mono<List<String>> = Mono.fromCallable {
-        lnkUserProjectService.getAllProjectsByUserId((authentication.details as AuthenticationDetails).id).filter { it.public }
+        lnkUserProjectService.getCreatedProjectsByUserId((authentication.details as AuthenticationDetails).id).filter { it.public }
     }
         .map { userProjects ->
             userProjects to lnkContestProjectService.getProjectsFromListAndContest(contestName, userProjects).map { it.project }
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
@@ -114,7 +114,7 @@ class LnkContestProjectController(
         @PathVariable contestName: String,
         authentication: Authentication,
     ): Mono<List<String>> = Mono.fromCallable {
-        lnkUserProjectService.getCreatedProjectsByUserId((authentication.details as AuthenticationDetails).id).filter { it.public }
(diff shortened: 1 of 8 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/LnkUserProjectService.kt
```diff
@@ -117,7 +118,7 @@ class LnkUserProjectService(
      * @param userId
      * @return list of [Project]s that are connected to user with [userId]
      */
-    fun getAllProjectsByUserId(userId: Long): List<Project> = lnkUserProjectRepository.findByUserId(userId)
+    fun getCreatedProjectsByUserId(userId: Long): List<Project> = lnkUserProjectRepository.findByUserIdAndProjectStatus(userId, ProjectStatus.CREATED)
```
petertrr, line 121: Maybe `getExistingProjectsByUserId` or `getNonDeletedProjectsByUserId`? Because it kinda sounds like 'get projects created by this user'

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
```diff
@@ -114,7 +114,7 @@ class LnkContestProjectController(
         @PathVariable contestName: String,
         authentication: Authentication,
     ): Mono<List<String>> = Mono.fromCallable {
-        lnkUserProjectService.getAllProjectsByUserId((authentication.details as AuthenticationDetails).id).filter { it.public }
```
kgevorkyan, line 117: As I remember, this request gives us the empty list, why `getCreatedProjectsByUserId` should help us, if the list of all projects are empty?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/LnkUserProjectService.kt
sanyavertolet: Right, will fix it

petertrr:
> Fixed a bug with select form in CreationView

@sanyavertolet please provide more details in the description on what exactly has been fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkContestProjectController.kt
sanyavertolet: The problem is that in prod database we have `DELETED` project with no organization while we somehow allow organization to be empty in `Project.kt` without setting organization type to be nullable

petertrr (approved):

PR #1065, opened 2022-08-19
Add random value to the container name in order to avoid conflicts

kgevorkyan:
### What's done:
* Add random value to the container name in order to avoid conflicts


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -254,4 +256,4 @@ class DockerAgentRunner(
 /**
  * @param id
  */
-private fun containerName(id: String) = "save-execution-$id"
+private fun containerName(id: String) = "save-execution-$id-${abs(Random.nextInt())}"
```
petertrr, line 259: Maybe format it to a fixed number of digits to make container names look more consistent with each other?

kgevorkyan: Add interval of possible values for random

nulls (approved):

PR #1070, opened 2022-08-20
Phase 1: New contests view template

akuleshov7:
### What's done:
- new contests view

image


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/contests/CountDownTimer.kt
```diff
@@ -0,0 +1,30 @@
...
+
+external interface CountDownProps : Props {
+    var contests: Array<ContestDto>
+}
+
+/**
+ * @return
+ */
+fun countDownTimer() = FC<CountDownProps> {props ->
+    console.log(Date().toLocaleString())
```
akuleshov7, line 28: temporary in phase 1

PR #1080, opened 2022-08-22
Supported overrideExecCmd and batchSize in save-cloud

### LINKED ISSUE #1052: Support `overrideExecCmd` and `batchSize` as arguments to save-cli
It should be supported after https://github.com/saveourtool/save-cli/issues/430

nulls:
### What's done:
* modified execFlags in batch size

It closes #1052

### DIFF
--- a/save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
+++ b/save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
@@ -13,11 +13,9 @@ import com.saveourtool.save.orchestrator.fillAgentPropertiesFromConfiguration
 import com.saveourtool.save.orchestrator.runner.AgentRunner
 import com.saveourtool.save.orchestrator.runner.AgentRunnerException
 import com.saveourtool.save.orchestrator.runner.EXECUTION_DIR
-import com.saveourtool.save.orchestrator.runner.TEST_SUITES_DIR_NAME
 import com.saveourtool.save.orchestrator.utils.LoggingContextImpl
 import com.saveourtool.save.orchestrator.utils.changeOwnerRecursively
 import com.saveourtool.save.orchestrator.utils.tryMarkAsExecutable
-import com.saveourtool.save.utils.DATABASE_DELIMITER
 import com.saveourtool.save.utils.orConflict
 
 import org.apache.commons.io.file.PathUtils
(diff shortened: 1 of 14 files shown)


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -209,9 +208,6 @@ class DockerService(
         "LongMethod",
     )
     private fun prepareImageAndVolumeForExecution(resourcesForExecution: Path, execution: Execution): RunConfiguration<PersistentVolumeId> {
-        // collect test suite names, which were selected by user
-        val saveCliExecFlags = " --include-suites \"${execution.getTestSuiteNames().joinToString(DATABASE_DELIMITER)}\" $TEST_SUITES_DIR_NAME"
```
petertrr, line 213: Is `--include-suites` flag not necessary anymore?

petertrr (commented):
Also, we have fixed save-core version at 0.3.2 now, and it includes save-cli too. To integrate these new features, we'll need to switch to `0.4.0-SNAPSHOT` again. Then, according to logic in `VersioningConfiguration`, we'll bundle new save-cli into new builds of save-cloud

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
nulls: yes, we found that save-cli gets command from backend with batch tests to be run. So `include-suites` is not required

petertrr: I think this flag might have been useful for test sources with wildcard mode enabled. However, we can't preprocess such suites, because tests get discovered only by the tool itself. I think we should add it to validation of test suite sources, if we are going to add one.

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
nulls: we have test suites ids on Execution which is used to detect tests are related to current execution (`test suite ids` -> `test suite` -> `tests`). Orchestrator provide `test suite names` using `test suite ids` from `execution`.
wildcard mode should detect tests in similar way for cases when `test suite names` is provided and `specific tests` and when only `specific tests` are provided

Or I missed something?

petertrr: Yes, wildcard mode should work fine in save-cli, however I don't think we can store tests during preprocessing of such suites. We'll get results after execution only.

PR #1089, opened 2022-08-23
Temporary hot fix to return the logic for "adding new tool" to SAVE

akuleshov7:
### What's done:
- introduced a wrapper to fix a hot error


sanyavertolet (approved):
Still don't understand why there is no way to enable custom tooltip and popover in both class `Component` and `FC` in Kotlin/JS. Seems weird

Later we will need to combine `inputTextFormOptional` and `inputTextFormRequired` into `inputTextForm`

petertrr:
> Still don't understand why there is no way to enable custom tooltip and popover in both class Component and FC in Kotlin/JS. 

Because the mechanism to run piece of raw JS code exactly once is different: in one case it's `componentDidMount`, in the other it's `useEffect` with empty dependencies

nulls (approved):

PR #1090, opened 2022-08-23
Keep execution history after removing of test executions

kgevorkyan:
### What's done:
* Keep execution after we remove test executions and mark them obsolete for history
* Added test suite source name field to ExecutionView


db/v-2/tables/execution.xml
```diff
@@ -44,4 +44,10 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add-testSuiteSourceName" author="kgevorkyan" context="dev or prod">
+        <addColumn tableName="execution">
+            <column name="test_suite_source_name" type="varchar(255)" defaultValue="NULL"/>
```
sanyavertolet, line 49: Why do we want it to be null?

kgevorkyan: Вackward compatibility for records, that didn't have it before

sanyavertolet: Maybe add one-time migration to fill all the required data?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestExecutionView.kt
```diff
@@ -72,7 +72,7 @@ class ContestExecutionView : AbstractView<ContestExecutionViewProps, State>(fals
         columns = columns<ExecutionDto> {
             column("result", "", { status }) { cellProps ->
                 val result = when (cellProps.row.original.status) {
-                    ExecutionStatus.ERROR -> ResultColorAndIcon("text-danger", faExclamationTriangle)
+                    ExecutionStatus.ERROR, ExecutionStatus.OBSOLETE -> ResultColorAndIcon("text-danger", faExclamationTriangle)
```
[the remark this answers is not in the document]
kgevorkyan: Changed

db/v-2/tables/execution.xml
kgevorkyan: Ok, changed

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Execution.kt
```diff
@@ -171,8 +173,19 @@ class Execution(
...
+            testSuiteSourceName = "",
         )
 
+        /**
+         * Parse and get testSuiteIds as List<Long>
+         *
+         * @param testSuiteIds
+         * @return list of TestSuite IDs
+         */
+        fun parseAndGetTestSuiteIds(testSuiteIds: String?): List<Long>? = testSuiteIds
```
sanyavertolet, line 185: Maybe `testSuiteIdsAsList`?

sanyavertolet (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Execution.kt
kgevorkyan: Since one of this methods is a member function, it more appropriate to use current name, take look:

```kotlin
val execution = Execution(....)
val testSuiteids = execution.parseAndGetTestSuiteIds()
```
**VS**
```kotlin
val execution = Execution(....)
val testSuiteids = execution.testSuiteIdsAsList()
```

Will leave this name as is

PR #1091, opened 2022-08-23
Fixes for `TestSuiteSelector` and `TestSuiteSourceCreationComponent`

### LINKED ISSUE #1079: Fix small selector issues
1) after adding a new test source with a selector window - we don't have a reactive window and we don't close the modal window
2) fix the search in the selector: right now we have the same selector text for suite IDs even if we have searched for the version of tests on previous step

sanyavertolet:
This PR closes #1079

### What's done:
 * Added (returned back) testSuiteSelector for general tests - all the avaliable ones
 * Fixed an issue with TestSuiteSelector in browser mode - now input form is cleared when an entity from avaliable list is chosen
 * Fixed an issue with unclosing modal - now when TestSuiteSource is successfully created, a new request for a fetch is sent

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
@@ -103,6 +103,15 @@ fun ChildrenBuilder.showTestSuiteSourceCreationModal(
 private fun testSuiteSourceCreationComponent() = FC<TestSuiteSourceCreationProps> { props ->
     val (testSuiteSource, setTestSuiteSource) = useState(TestSuitesSourceDto.empty.copy(organizationName = props.organizationName))
     val (saveStatus, setSaveStatus) = useState<SourceSaveStatus?>(null)
+    val fetchTestSuiteSource = useRequest {
+        post(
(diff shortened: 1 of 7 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
```diff
@@ -112,6 +121,7 @@ private fun testSuiteSourceCreationComponent() = FC<TestSuiteSourceCreationProps
             responseHandler = ::responseHandlerWithValidation,
         )
         if (response.ok) {
+            fetchTestSuiteSource()
```
petertrr, line 124: Do we have an issue for better support of this fetch operation? Because it may b quite slow, and now we neither display status of fetching in progress nor indicate that it has started.

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuiteSourceCreationComponent.kt
sanyavertolet: I guess for now we don't. I think it might be nice to use something like web sockets for this kind of tasks.

PR #1100, opened 2022-08-25
Split `useRequest` into `useRequest` and `useDeferredRequest`

petertrr:
As discussed in one of the previous PRs, shared logic for `useEffect`-like and callback-like requests doesn't make sense: `isSending` is not needed for `useEffect`-like requests, and because it's a flag that controls request sending, passing dependencies into deferred `useRequest`s doesn't have any impact.

This PR splits the hook into two new ones: the one only with dependencies and the one that uses `isSending` under the hood. Judging by the old code, dependencies passed together with `isDeferred = true` shouldn't have any effect, but for some reason we pass them in plenty of places.


sanyavertolet (approved):
Now it seems much more clear that it used to be

PR #1102, opened 2022-08-25
Hotfix: Set correct path to save-cli in `deploy_images.yml`

petertrr:
Caused by changes in ded475278ee66d06787741c523777ce25baa84b9, where I've changed configuration for `Download` tasks for save-agent and save-cli. It now works better and doesn't rely on manual copying, like it used to be in this workflow.

Example run with these changes: https://github.com/saveourtool/save-cloud/runs/8017837622?check_suite_focus=true


akuleshov7 (approved):

akuleshov7:
this should fix buildAndDeploy also?

Issue #1035, opened 2022-08-12
Properly support results of the contests


akuleshov7:
1. All contest run results should be properly stored in a database and linked to the project. This should be elegantly be done in the same manner as execution history is done - the best scenario is when we reuse all the related logic to existing run:
- You can have a normal execution run;
- Have a table for contest results with links to these execution - something like this: **Score, ExecutionId, Contest** ;
- As executions are linked to the project already - there will not be any problems to show/calculate total score for the project on the frontend. Also it will be easy to find the champion in contests and show results dashboard on the contests page.

3. We need to fill in two tabs on contests View: 
- Participants (a good example is [here](https://www.hackerrank.com/leaderboard));
- Results only for **your** projects. That is a tough task, because you we need to show results for **all projects** where user is added. For example - if user added to both diktat and ktlint and both of them were participating in contests - we should show them.

sanyavertolet:
The frontend part is done with #1081 and #1108. Backend part will be done as #1115

PR #1108, opened 2022-08-26
ContestSummaryMenu improvements

sanyavertolet:
This PR is connected with #1035 

### What's done:
 * Slightly improved `ContestSummaryMenu`
 * Fixed styles for execution labels for non-contest executions.
### How it looks:
screenshot

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/contests/ContestSummaryMenu.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/contests/ContestSummaryMenu.kt
@@ -3,7 +3,6 @@
 package com.saveourtool.save.frontend.components.basic.contests
 
 import com.saveourtool.save.entities.ContestResult
-import com.saveourtool.save.frontend.components.basic.scoreCard
 import com.saveourtool.save.frontend.utils.*
 
 import csstype.*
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/contests/ContestSummaryMenu.kt
```diff
@@ -30,6 +31,72 @@ external interface ContestSummaryMenuProps : Props {
...
+    ul {
+        className = ClassName("col-10 mb-2 list-group")
+        displayResult(
+            "Top position",
+            "Project",
+            "Organization",
+            "Score",
+            null
+        )
+        results.filter {
```
petertrr, line 44: It's too implicit that results here are already sorted (also - are they being sorted twice, both on backend and on frontend?). Maybe rename `results` -> `resultsSorted`?

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/contests/ContestSummaryMenu.kt
sanyavertolet: Oh ok, will rename it

sanyavertolet: Done

PR #1109, opened 2022-08-26
`TestSuiteSelector` version label

### LINKED ISSUE #1104: TestSuiteSelectorSearchMode should not display all versions
Now TestSuiteSelectorSearchMode shows all available versions, which is complicated. Probably need to think of something like displaying only latest version.

sanyavertolet:
This PR closes #1104

### What's done:
 * Fixed tooltips (finally)
 * Added latest_version column to test-suites-source.xml
 * Added version label to TestSuitesDisplayer.kt
 * Improved `TestSuiteSelectorSearchMode` to be able to show only latest version

### How it looks:
https://user-images.githubusercontent.com/35039155/186946548-4c6d3391-b5c6-4d1e-a792-d8d0815c2b86.mov


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
```diff
@@ -41,14 +42,35 @@ class TestSuitesService(
...
+            "Do not save test suites from different commits at the same time."
+        }
+
+        val testSuiteSourceVersion = testSuitesDto.first().version
+        val testSuiteSource = testSuitesDto.first()
+            .let { dto ->
+                testSuitesSourceService.getByName(dto.source.organizationName, dto.source.name)
+            }
+            .apply {
+                latestVersion = testSuiteSourceVersion
```
petertrr, line 71: What's this for? Load data and override its version?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuitesSourceDto.kt
```diff
@@ -24,6 +25,7 @@ data class TestSuitesSourceDto(
     val gitDto: GitDto,
     val branch: String,
     val testRootPath: String,
+    val latestVersion: String = "",
```
petertrr, line 28: If it is possible to have an empty value here, maybe better make it nullable? Or if empty value is not possible, it's better to remove the default value

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/testsuiteselector/TestSuiteSelectorSearchMode.kt
```diff
@@ -61,6 +63,30 @@ private fun ChildrenBuilder.buildInput(
...
+    testSuites: List<TestSuiteDto>,
+    selectedTestSuites: List<TestSuiteDto>,
+    isOnlyLatestVersion: Boolean,
+    onTestSuiteClick: (TestSuiteDto) -> Unit,
+) {
+    val testSuitesToBeShown = testSuites.filter {
+        !isOnlyLatestVersion || it.version == it.source.latestVersion
+    }
+
+    // small hack to hide version if only latest is shown
```
petertrr, line 76: Do we need to hide it though?

sanyavertolet: Well, I think so. I try not to show too many labels if possible

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuitesSourceDto.kt
sanyavertolet: Empty value is possible as far as we have created test suite source, but not fetched yet. I see no difference whether to check nullability or emptiness

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
sanyavertolet: Yes, we still need to load `testSuiteSource` in order to save `testSuite`s. And we need to update it. Decided to update as soon as I get it from db

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuitesSourceDto.kt
petertrr: For me nullable field says more explicitly that the value can be missing. To handle empty strings as an empty value you have to always keep it in mind, while nullability leads to some enforcement during compilation.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesController.kt
```diff
@@ -54,7 +54,10 @@ class TestSuitesController(
     @Tag(name = "internal")
     @ApiResponse(responseCode = "200", description = "Successfully saved test suites.")
     fun saveTestSuite(@RequestBody testSuiteDtos: List<TestSuiteDto>): Mono<List<TestSuite>> =
-            Mono.just(testSuitesService.saveTestSuite(testSuiteDtos))
+            Mono.just(testSuiteDtos)
+                .filter { it.isNotEmpty() }
+                .map { testSuitesService.saveTestSuite(testSuiteDtos) }
```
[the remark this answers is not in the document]
sanyavertolet: Oh sure

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/testsuiteselector/TestSuiteSelectorSearchMode.kt
petertrr: I think it's confusing that at first you see versions and then after you tick 'Show only latest', the version disappears. I think it's better to keep seeing the exact version you are about to select

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuitesSourceDto.kt
sanyavertolet: Makes sense, will fix it

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
petertrr: Oh, so we are setting version from test suites, ok, makes sense. But how do you know we are saving test suiites with the latest version and not fetching from an old commit? Maybe we should keep creation timestamp in test suite source too?

sanyavertolet: For now we are fetching only HEAD of a given branch, so I am not sure that we need to add a new column for that

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/testsuiteselector/TestSuiteSelectorSearchMode.kt
sanyavertolet: Ok, we might want to display version regardless the checkbox, will fix it

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
```diff
@@ -92,6 +92,17 @@ class TestSuitesSourceService(
...
         }
 
+    /**
+     * Raw update
+     *
+     * @param entity [TestSuitesSource] to be updated
+     * @return updated [TestSuitesSource]
+     */
+    fun update(entity: TestSuitesSource): TestSuitesSource {
+        entity.requiredId()
```
nulls, line 102: what is it? if you added it to check that this entity is for update -- need to update error message and add a comment at least

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuitesDisplayer.kt
```diff
@@ -62,6 +66,15 @@ fun ChildrenBuilder.showAvaliableTestSuites(
                         +(testSuite.tags?.joinToString(", ") ?: "")
                     }
 
+                    if (displayMode.shouldDisplayVersion()) {
+                        small {
+                            asDynamic()["data-toggle"] = "tooltip"
+                            asDynamic()["data-placement"] = "bottom"
+                            title = "Hash of commit with current test suite"
+                            +testSuite.version.take(GIT_HASH_PREFIX_LENGTH)
```
nulls, line 74: maybe fix/enhance it on preprocessor side instead of hacking a displayed version?

db/v-2/tables/test-suites-source.xml
```diff
@@ -31,4 +31,12 @@
         <addUniqueConstraint tableName="test_suites_source" columnNames="organization_id,name" constraintName="test_suites_source_name"/>
         <addUniqueConstraint tableName="test_suites_source" columnNames="git_id,test_root_path,branch" constraintName="test_suites_source_git_location"/>
     </changeSet>
+
+    <changeSet id="add-latest-version" author="sanyavertolet">
```
nulls, line 35: as I got, we introduce a new field just to support filtering `Show only latest version`. 
Is this a specific field required? Can we have an extra endpoint for it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
nulls: Preprocessor has ability to fetch tests  with a specific version, so potentially we can fetch with old version. For example: tag name vs commit version detected automatically

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuitesDisplayer.kt
sanyavertolet: I guess it is not that important
Moreover, collision happens, so let's at least be sure it works on backend

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
sanyavertolet: I like a message that is inside `requiredId()`
Will add comment

db/v-2/tables/test-suites-source.xml
sanyavertolet: It is going to be either null, or prefix of commit hash (just like GitHub does)

There will be an extra endpoint, but not sure it should be in this PR.

This field is filled with the first fetch (which mostly will accrue right after source is created)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/testsuiteselector/TestSuiteSelectorSearchMode.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
sanyavertolet: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
sanyavertolet: If someone fetched some older version lately, does he expect it to be called *latest*?

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
sanyavertolet: Renamed to `latestFetchedVersion`

PR #1116, opened 2022-08-29
Editing TestSuitesSources

### LINKED ISSUE #1105: Support editing of snapshots of TestSuitesSources
* [x] #1169 
* [ ] #1116
* [x] #1171 
* [x] #1210
* [x] #1180

nulls:
### What's done:
* Update TestSuitesSource entity
* Update version name of snapshot of TestSuitesSource
* Fetch a custom version of TestSuitesSource

It closes #1105


nulls:
As agreed, we will remove `branchName` from `TestSuitesSource`.
Now version of `TestSuites` and `TestSuitesSourceSnapshot` is `tagName` in general. 
But also we support fetching a single version from a branch -- `branchName` will be version name for such snapshots. It will overwrite the existed version

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -432,6 +499,11 @@ class TestSuitesSourceController(
                     "TestSuitesSource not found by name $name for organization $organizationName"
                 }
 
+    private fun getTestSuitesSource(id: Long): Mono<TestSuitesSource> =
+            blockingToMono {
+                testSuitesSourceService.findById(id)
+            }
```
sanyavertolet, line 505: Maybe process situation when source is not found by id here?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Git.kt
```diff
@@ -21,11 +21,11 @@ class Git(
...
-) : BaseEntity() {
+) : BaseEntityWithDto<GitDto>() {
     /**
      * @return git dto
      */
-    fun toDto() = GitDto(
+    override fun toDto() = GitDto(
         url = url,
         username = username,
         password = password,
```
sanyavertolet, line 31: Do we always send all the credentials inside `GitDto`?

nulls: yes

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
nulls: yes, sure -- will add

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Git.kt
sanyavertolet: Do we have an issue for that?

nulls: Probably, but not in this PR )

nulls: it's required more changed, also need to revise it additionally: we need to get somehow credential from backend to preprocessor. also access has only authorized users ~and only users who has access to the test suites source.~
We need to implement role managements for test suites source. It's not a scope of this task, but can add checking access to `Organization`

sanyavertolet: Let's at least add **new** issue and move on :)

nulls: actually, currently it can break public tests. we need to introduce `PermissionEvaluator` for `TestSuitesSource` and then we can rework it

nulls: > Let's at least add **new** issue and move on :)

created: https://github.com/saveourtool/save-cloud/issues/1219

PR #1122, opened 2022-08-30
Support additional setup sh

### LINKED ISSUE #1120: Support `setup.sh` script to install local environment
If this file is added to additional files - then we use it as setup for our local env.

nulls:
### What's done:
* support setup.sh from additional resources
* support save-overrides.toml

It closes #1120

### DIFF
--- a/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
+++ b/save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
@@ -6,6 +6,7 @@ import com.saveourtool.save.agent.utils.*
 import com.saveourtool.save.agent.utils.readFile
 import com.saveourtool.save.agent.utils.requiredEnv
 import com.saveourtool.save.agent.utils.sendDataToBackend
+import com.saveourtool.save.core.config.resolveSaveOverridesTomlConfig
 import com.saveourtool.save.core.files.getWorkingDirectory
 import com.saveourtool.save.core.logging.describe
 import com.saveourtool.save.core.plugin.Plugin
@@ -13,6 +14,7 @@ import com.saveourtool.save.core.result.CountWarnings
 import com.saveourtool.save.core.utils.ExecutionResult
 import com.saveourtool.save.core.utils.ProcessBuilder
 import com.saveourtool.save.core.utils.runIf
+import com.saveourtool.save.domain.FileKey
 import com.saveourtool.save.domain.TestResultDebugInfo
 import com.saveourtool.save.plugins.fix.FixPlugin
 import com.saveourtool.save.reporter.Report
@@ -41,6 +43,7 @@ import kotlinx.serialization.json.Json
 import kotlinx.serialization.modules.SerializersModule
 import kotlinx.serialization.modules.polymorphic
 import kotlinx.serialization.modules.subclass
+import okio.Path
 
 /**
  * A main class for SAVE Agent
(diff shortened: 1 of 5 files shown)


save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -121,6 +141,57 @@ class SaveAgent(private val config: AgentConfiguration,
...
+    // a temporary workaround for python integration
+    private fun executeAdditionallySetup(targetDirectory: Path, additionalFiles: List<FileKey>) = runCatching {
+        additionalFiles
+            .singleOrNull { it.name == "setup.sh" }
+            ?.let { fileKey ->
+                val targetFile = targetDirectory / fileKey.name
+                logDebugCustom("Additionally setup of evaluated tool by $targetFile")
+                val setupResult = ProcessBuilder(true, FileSystem.SYSTEM)
+                    .exec(
+                        targetFile.toString(),
```
petertrr, line 153: Shouldn't it be `./$targetFile`? Or maybe we should call shell ourselves in case their script doesn't specify any interpreter?

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -121,6 +141,57 @@ class SaveAgent(private val config: AgentConfiguration,
...
+                if (setupResult.code != 0) {
+                    throw IllegalStateException("${fileKey.name} is failed with error: ${setupResult.stderr}")
+                }
+            }
+    }
+
+    // prepare save-overrides.toml based on config.save.*
+    private fun prepareSaveOverridesToml(targetDirectory: Path) = runCatching {
+        with(config.save) {
+            val generalConfig = buildMap {
```
petertrr, line 167: Waiting for ktoml to support serialization? 🤔

nulls: actually, I wanted to construct `WarnPluginConfig` directly and then write it to a toml file. But it requires `regex` library for linuxx64. (similar as for 'coroutines')

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
nulls: I see that `save-cli` is called by `./script-name` -- will fix it
but I don't think that we should call a shell. I think it's responsibility of script's provider

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
petertrr: Isn't regex support completely provided by Kotlin standard library?

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -121,6 +141,57 @@ class SaveAgent(private val config: AgentConfiguration,
...
+                val setupResult = ProcessBuilder(true, FileSystem.SYSTEM)
+                    .exec(
+                        "./$targetFile",
+                        "",
+                        null,
+                        SETUP_SH_TIMEOUT
+                    )
+                if (setupResult.code != 0) {
+                    throw IllegalStateException("${fileKey.name} is failed with error: ${setupResult.stderr}")
+                }
```
petertrr, line 160: Maybe also `logTrace` stdout if setup is successful for easier debugging?

petertrr (approved):

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
nulls: actually, we log info `logInfoCustom("Additionally setup has completed for execution $executionId")` as for rest actions

petertrr: Yes, but I mean for cases when, for example, for `pip install` from `setup.sh` we would like to see exactly which versions have been installed or something like this. IIRC, our `ProcessBuilder` captures stdout completely and doesn't replay it automatically.

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
nulls: looks like the issue in IDE:
image

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
nulls: got it, will add logging of stdout

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
petertrr: I think there is no dependency on `warn-plugin` in save-agent

nulls: problem with `regex` there.

The following change fixes it
image

@petertrr @akuleshov7 is it okay to apply this?

petertrr: It looks really strange. If it helps simplifying code, then maybe, but otherwise I'd rather not.

nulls: will try with using Toml's serialization then

nulls: KToml doesn't support serialization. Will create a separate task to add more targets for save-agent to simplify debugging

PR #1124, opened 2022-08-31
Improve tolerance to backend inavailability in orchestrator

### LINKED ISSUE #1075: Orchestrator shouldn't stop agents if backend is unavailable
Now if agents keep heartbeating, but backend is unavailable and `agentService.updateAgentStatusesWithDto` fails with network error, orchestrator will treat agents as abnormal and terminate them after some time. This is incorrect; if agents keep heartbeating in `IDLE` state, execution can still be finished once backend is back online

petertrr:
* Skip heartbeat processing but update `HeartbeatInspector`'s registry if agent state can't be posted to backend instead of failing and potentially treating all agents as crashed
* Make `areAgentsHaveStarted` track agents per execution

Closes #1075


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/agent/Heartbeat.kt
```diff
@@ -13,9 +13,13 @@ import kotlinx.serialization.Serializable
  * Progress of tests execution
  *
  * @property percentCompletion percentage of completed jobs, integer 0..100
+ * @property executionId ID of an execution under which the progress is being tracked
  */
 @Serializable
-data class ExecutionProgress(val percentCompletion: Int) {
+data class ExecutionProgress(
+    val percentCompletion: Int,
+    val executionId: Long,
```
akuleshov7, line 21: how do you use executionId here?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -58,21 +57,20 @@ class HeartbeatController(private val agentService: AgentService,
...
-            .onErrorResume(WebClientResponseException::class.java) {
-                logger.warn("Couldn't update agent statuses for agent ${heartbeat.agentId}, will skip this heartbeat: ${it.message}")
-                Mono.empty()
+        val executionId = heartbeat.executionProgress.executionId
+        logger.info("Got heartbeat state: ${heartbeat.state.name} from ${heartbeat.agentId} under execution id=$executionId")
+        return {
+            dockerService.markAgentForExecutionAsStarted(executionId)
+            heartBeatInspector.updateAgentHeartbeatTimeStamps(heartbeat)
+        }
+            .toMono()
```
akuleshov7, line 66: why not Mono.just({}) here to make lazy evaluations

petertrr: It's an extension which wraps the lambda into `Mono.fromSupplier`, so it's essentially the same

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -39,6 +42,8 @@ class DockerService(
     private val agentRunner: AgentRunner,
     private val agentService: AgentService,
 ) {
+    private val areAgentsHaveStarted: ConcurrentMap<Long, AtomicBoolean> = ConcurrentHashMap()
```
akuleshov7, line 45: `haveAgentsStarted`? or diktat issue?

akuleshov7, line 45: `AtomicReference<Map<Long, AtomicBoolean>>`?
@nulls what do you suggest?

petertrr: `ConcurrentHashMap` allows concurrent access to different keys, while `AtomicReference<Map<*, *>>` will restrict access to a single thread only, I think

petertrr: Yep, but this variable has been here before, I just moved it

save-cloud-charts/save-cloud/templates/orchestrator-configmap.yaml
```diff
@@ -14,6 +14,7 @@ data:
     management.endpoints.web.exposure.include=*
     orchestrator.agent-settings.orchestrator-url=http://{{ .Values.orchestrator.name }}
     orchestrator.agent-settings.backend-url=http://{{ .Values.backend.name }}
+    orchestrator.agent-settings.debug=true
```
nulls, line 17: do we want support `debug` by default?

akuleshov7 (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
nulls: agree about `ConcurrentHashMap`

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -100,20 +105,22 @@ class DockerService(
                 val duration = AtomicLong(0)
                 Flux.interval(configProperties.agentsStartCheckIntervalMillis.milliseconds.toJavaDuration())
                     .takeWhile {
-                        duration.get() < configProperties.agentsStartTimeoutMillis && !areAgentsHaveStarted.get()
+                        val isAnyAgentStarted = areAgentsHaveStarted.computeIfAbsent(executionId) { AtomicBoolean(false) }.get()
```
nulls, line 108: do we need compute here? maybe the same as onComplete:
`areAgentsHaveStarted[executionId]?.get() == true`?

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -126,8 +127,9 @@ class SaveAgent(private val config: AgentConfiguration,
         logInfoCustom("Scheduling heartbeats")
         while (true) {
             val response = runCatching {
+                val executionId = requiredEnv(AgentEnvName.EXECUTION_ID).toLong()
```
nulls, line 130: maybe move it to `AgentConfiguration`?

PR #1126, opened 2022-08-31
Invalid validation errors on frontend

### LINKED ISSUE #1066: Invalid validation error in Create organization view
If we try to create an organization with an already existing name, we will get an error about an invalid character in the name. But we expected to get only an error about an already existing name.
Сни12121

sanyavertolet:
This PR closes #1066

### What's done:
 * Fixed a bug with wrong error message on frontend when CONFLICT response received
 * Improved forms - now required forms are also wrapped into FC and have tooltip


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputTextFormRequired.kt
```diff
@@ -0,0 +1,149 @@
...
+                    +(form.errorMessage ?: "Input a valid ${form.str}, please.")
+                }
+            }
+        }
+    }
+}
+
+private fun inputTextFormRequiredWrapper() = FC<InputTextFormRequiredProps> { props ->
+    useTooltip()
+    inputTextFormRequired(
```
petertrr, line 139: Why do you need a separate function? Maybe create elements right here, because otherwise you have to pass too many parameters

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputTextFormRequired.kt
sanyavertolet: @akuleshov7 once mentioned that this is a temporary solution

PR #1133, opened 2022-09-01
Rework modal windows

### LINKED ISSUE #803: Rework modal windows

sanyavertolet:
This PR closes #803

### What's done:
 * Joined `runConfirmationWindow` and `runErrorWindow` into `modalBuilder` and `displayModal` (`modalBuilder` wrapped into `ReactModal`)
 * Updated `ReactModal` styles


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/modal/ModalBuilder.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Universal function to create modals with bootstrap styles inside react modals.
+ *
+ * @param isOpen modal openness indicator - should be in state
+ * @param title title of the modal that will be shown in top-left corner
+ * @param message main text that will be shown in the center of modal
+ * @param modalStyle [Styles] that will be applied to react modal
+ * @param onCloseButtonPressed callback that will be applied to `X` button in the top-right corner
+ * @param buttonBuilder lambda that generates several buttons, must contain either [button] or [buttonBuilder]
```
petertrr, line 26: > must contain either [button] or [buttonBuilder]

But it's not checked in any way :) I see the intention, however

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/modal/ModalBuilder.kt
sanyavertolet: For now I have no idea how to implement it to be both simple and safe. Could pass array of buttons, but not sure it would work nice. Moreover, I do not like that I should explicitly pass `onCloseButtonPressed`, but again I have no better idea

PR #1139, opened 2022-09-02
[Gradle] Use custom save-cli distribution only with snapshot

petertrr:
Respect values of `-PsaveCliPath` and `-PsaveCliVersion` only if save-core version in `libs.versions.toml` ends with `-SNAPSHOT`


buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/VersioningConfiguration.kt
```diff
@@ -117,5 +117,9 @@ fun Project.readSaveCliVersion(): String {
 fun Project.getSaveCliPath(): String {
     val saveCliVersion = readSaveCliVersion()
     val saveCliPath = findProperty("saveCliPath") as String? ?: "https://github.com/saveourtool/save-cli/releases/download/v$saveCliVersion"
```
nulls, line 119: duplicate?

nulls (approved):

PR #1144, opened 2022-09-02
Use `docker compose` instead of `docker-compose`

0x6675636b796f75676974687562:
There's no separate `docker-compose` executable on Linux anymore -- just
the Docker plug-in.


akuleshov7 (approved):
lgtm

PR #1145, opened 2022-09-02
Don't include "save-agent" on Windows because of the missing `libcurl` dependency

0x6675636b796f75676974687562:


akuleshov7 (approved):
lgtm

petertrr (changes requested):
We already have a conditional dependency on `save-agent` in `save-backend`, and also compilation and syntax hightlightning in IDE work on Windows too using cross-compilation. What exactly will we solve by excluding the project completely?

0x6675636b796f75676974687562:
Converted to draft as discussed with @petertrr.

PR #1160, opened 2022-09-05
Calculate score for execution

kgevorkyan:
### What's done:
* Calculate score for execution


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/execution/ExecutionDto.kt
```diff
@@ -42,6 +42,33 @@ data class ExecutionDto(
...
+    /**
+     * @return recall rate
+     */
+    fun getRecallRate() = calculateRate(matchedChecks, matchedChecks + unmatchedChecks) ?: 0
+
+    /**
+     * @return F-measure
+     */
+    // FixMe: do we need some other algorithm? for now it's F-measure
+    fun calculateScore() = (2 * getPrecisionRate() * getRecallRate()) / (getPrecisionRate() + getRecallRate()).toDouble()
```
petertrr, line 70: I think that to support better extensibility in the future, we should from the beginning isolate this logic a bit more, maybe even move completely to JVM sources and only request already computed values from the frontend. Then we may introduce a enum of score types, for example. Also `calculateScore` should accept at least `TestingType` as an argument, because I would imagine that score measures would differ for contests and for ordinary executions.
cc @akuleshov7

sanyavertolet, line 70: Seems like in some cases there could be null division. Probably we would like to check denominator to be not null

kgevorkyan: Good remark, done

kgevorkyan: Moved into separate file, added enum for score types

petertrr (approved):

sanyavertolet (approved):

Issue #1061, opened 2022-08-18
Support latest changes in save-api


nulls:
Subj

akuleshov7:
why backlog?

akuleshov7:
Let's wait till contests will be finished and then we will update the client

petertrr:
Also after we update the client library, we can remove `CloneRepositoryController` and use only new `trigger` endpoints

nulls:
> Also after we update the client library, we can remove `CloneRepositoryController` and use only new `trigger` endpoints

it's already in my PR about editing test suite sources

PR #1169, opened 2022-09-06
cleanup unused methods and endpoints

nulls:
First part of changes from #1105: cleanup of `CloneRepositoryController.kt` and support changes in `save-cli`


nulls:
Save CLI part is refactored without testing, it will be covered in scope #1061

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/SelectForm.kt
```diff
@@ -109,6 +109,7 @@ fun <D : Any> selectFormRequired() = FC<SelectFormRequiredProps<D>> { props ->
                 className = ClassName("form-control")
                 id = props.formType.name
                 required = true
+                // TODO: why we need an extra option
```
petertrr, line 112: I believe it's an empty option which is selected by default

save-api/src/main/kotlin/com/saveourtool/save/api/utils/RequestUtils.kt
```diff
@@ -101,60 +84,17 @@ suspend fun HttpClient.uploadAdditionalFile(
...
-                    json.encodeToString(shortFileInfo),
-                    Headers.build {
-                        append(HttpHeaders.ContentType, ContentType.Application.Json)
-                    }
-                )
-            }
-        }))
-    }
+suspend fun HttpClient.submitExecution(runExecutionRequest: RunExecutionRequest): HttpResponse = this.post {
+    url("${Backend.url}/api/$v1/run/trigger")
```
petertrr, line 94: I believe this should contain `testingType` as a request param?

nulls: yes

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/SelectForm.kt
nulls: probably, but there is `value = props.selectedValue`. maybe we need to add an empty value only if `props.selectedValue` is not presented in `elements`. Need to discuss with @sanyavertolet

save-api/src/main/kotlin/com/saveourtool/save/api/utils/RequestUtils.kt
nulls: returned back

save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClient.kt
```diff
@@ -81,126 +77,50 @@ class SaveCloudClient(
...
+    ): RunExecutionRequest? {
+        val runExecutionRequest = RunExecutionRequest(
+            projectCoordinates = ProjectCoordinates(
+                organizationName = evaluatedToolProperties.organizationName,
+                projectName = evaluatedToolProperties.projectName,
+            ),
+            testSuiteIds = evaluatedToolProperties.testSuites
+                .split(DATABASE_DELIMITER)
+                .map { it.toLong() },
+            files = additionalFiles?.map { it.toStorageKey() }.orEmpty(),
```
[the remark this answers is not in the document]
nulls: it's used in backend. we put files to `RunExecutionRequest` and then it's put to `Execution`

petertrr (approved):

Issue #1115, opened 2022-08-29
Rating: add rating to organisations and tools


akuleshov7:
We need to have the following addition on the backend:

1) We would like to have proper calculations of **contest** rating
2) To do this we need to have new fields in `LnkContestProject.kt: bestExecution: Execution` and `bestScore: Int` 
3) In Organization and Project entities we should have `totalScore` (or something like this - please check that these fields are not yet added). It should be taken as sum of the list of BEST EXECUTIONS 
4) When contest execution finishes - we update all these fields and rating

kgevorkyan:
First phase:

* Pass type of execution to the backend
* Fill lnk_contest_execution
* Move `score` from lnk_contest_execution into the Execution entity

petertrr:
> Move score from `lnk_contest_execution` into the Execution entity

I think the assumption there was that we'll only need scores in contests. Are you planning to change this?

kgevorkyan:
> I think the assumption there was that we'll only need scores in contests. Are you planning to change this?

We decide with @akuleshov7 to move it anyway to Execution

petertrr:
Ok, to clarify reasons - score will be used not only for contests, even if it's calculation may depend on the exact type of execution.

petertrr:
>  It should be taken as sum of the list of BEST EXECUTIONS

Why do we need to sum all executions? For me this metric is not descriptive, because it mostly represents not the quality of the tool but rather how active they are using our platform. I think historical diagram of best score or variance of best scores over time may be a better option.

petertrr:
The only thing that concerns me is that score in different contests may be calculated differently. Now we support only F-measure (i.e. combined precision and recall for warn-plugin-based tests), but we may have different types in the future. Then we'll need to change rules of aggregation and recalculated already stored values.

Right now, to calculate `contestRating` (this is the name of the field we already have) we can do something like
`contestRating = lnkContestProject.filter { it.project == this }.sumBy { bestScore }`

And to calculate organization rating,
`organizationRating = projects.findAllByOrganization(this).sumBy { it.contestRating} `

I think at least the latter can be calculated per request, because it doesn't involve a lot of data. `Project.contestRating` can be cahced as a column, however.

petertrr:
For other plugins calculation of rating requires https://github.com/saveourtool/save-cli/issues/449

petertrr:
https://github.com/saveourtool/save-cloud/pull/1162#discussion_r963028426

[...]

PR #1170, opened 2022-09-06
Don't recalculate counters in Execution if value is `NOT_APPLICABLE`

petertrr:
* Store aggregated metrics in Execution excluding `NOT_APPLICABLE` metrics from individual test executions (until we can calculate scores for other types of plugins too)
* Change logic of dispaly on frontend: don't calculate metrics only if all test executions under an execution are `NOT_APPLICABLE`; otherwise use filtered data from Execution

Related to #1115

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
@@ -167,6 +167,7 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
         "LongMethod",
         "MAGIC_NUMBER",
         "MagicNumber",
+        "PARAMETER_NAME_IN_OUTER_LAMBDA",
     )
     @Transactional
     fun saveTestResult(testExecutionsDtos: List<TestExecutionDto>): List<TestExecutionDto> {
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
@@ -214,10 +214,10 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
                     it.unexpected = testExecDto.unexpected
 
                     with(counters) {
-                        unmatchedChecks += unmatchedChecks.increaseIfApplicable(testExecDto.unmatched)
-                        matchedChecks += matchedChecks.increaseIfApplicable(testExecDto.matched)
-                        expectedChecks += expectedChecks.increaseIfApplicable(testExecDto.expected)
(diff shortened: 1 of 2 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
```diff
@@ -352,6 +356,10 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
         }
     }
 
+    private fun Long.increaseIfApplicable(delta: Long?) = takeUnless { CountWarnings.isNotApplicable(it.toInt()) }
```
akuleshov7, line 359: Nariman also love takeUnless, but I do not think it is more readable as straight-forward `if`

akuleshov7 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
```diff
@@ -212,10 +213,12 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
                     it.expected = testExecDto.expected
                     it.unexpected = testExecDto.unexpected
 
-                    counters.unmatchedChecks += testExecDto.unmatched ?: 0L
-                    counters.matchedChecks += testExecDto.matched ?: 0L
-                    counters.expectedChecks += testExecDto.expected ?: 0L
-                    counters.unexpectedChecks += testExecDto.unexpected ?: 0L
+                    with(counters) {
+                        unmatchedChecks += unmatchedChecks.increaseIfApplicable(testExecDto.unmatched)
```
nulls, line 217: maybe `=` instead of `+=`? also `testExecDto.unmatched` can be not applicable. need to check it instead of `unmatched`

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
petertrr: Yeah, but diktat will turn a straightforward `if` into a messy `?.let { } ?: `

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
petertrr: Yeah, sure, thanks

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
nulls, line 359:
```suggestion
    private fun Int?.orZeroIfNotApplicable() = this?.takeUnless { CountWarnings.isNotApplicable(it) } ?: 0
```

PR #1171, opened 2022-09-06
Fetch by tags, branch and commits

nulls:
Introducing fetching test suites by tags, branch and commit.
Now the button `fetch` on frontend fetches all tags + re-fetch default branch

It's part of #1105


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
```diff
@@ -156,8 +158,31 @@ class TestSuitesSourceService(
...
+                .toBodilessEntity()
+        }
+        .collectList()
+        .flatMap {
+            preprocessorWebClient.post()
+                .uri("/git/default-branch-name")
+                .bodyValue(testSuitesSource.gitDto)
+                .retrieve()
+                .bodyToMono<String>()
+                .flatMap { branchName ->
```
petertrr, line 180: This `flatMap` can be moved out of the current lambda to avoid nesting

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/TestSuitesPreprocessorController.kt
```diff
@@ -18,6 +19,9 @@ import org.springframework.web.bind.annotation.RestController
 import reactor.core.publisher.Mono
 import kotlin.io.path.div
 
+typealias TestSuiteList = List<TestSuite>
+typealias CloneAndProcessDirectory = GitPreprocessorService.(GitDto, String, GitRepositoryProcessor<TestSuiteList>) -> Mono<TestSuiteList>
```
petertrr, line 23:
```suggestion
typealias CloneAndProcessDirectoryAction = GitPreprocessorService.(GitDto, String, GitRepositoryProcessor<TestSuiteList>) -> Mono<TestSuiteList>
```
Maybe something like this?

petertrr (approved):

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/controllers/TestSuitesPreprocessorController.kt
nulls: agree. changed

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
nulls: thanks, moved

PR #1175, opened 2022-09-07
Bugfix/correct filters global rating view

Arrgentum:
- Make filters with partial match search
- Add a value deletion button
- Change the default value of the filter
- Do not change the parameter values in the url if filters is null

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/contests/ContestGlobalRatingView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/contests/ContestGlobalRatingView.kt
@@ -148,8 +148,8 @@ class ContestGlobalRatingView : AbstractView<ContestGlobalRatingProps, ContestGl
                                 organizationFilters = filter
                             }
                             getOrganization(filter)
-                            getProject(ProjectFilters(null))
-                            window.location.href = filterValue?.let {"${window.location.href.substringBefore("?")}?organizationName=$filterValue"} ?: window.location.href.substringBefore("?")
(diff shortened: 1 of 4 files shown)


Cheshiriks (approved):
lgtm

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/contests/ContestGlobalRatingView.kt
```diff
@@ -148,8 +148,8 @@ class ContestGlobalRatingView : AbstractView<ContestGlobalRatingProps, ContestGl
                                 organizationFilters = filter
                             }
                             getOrganization(filter)
-                            getProject(ProjectFilters(null))
-                            window.location.href = "${window.location.href.substringBefore("?")}?organizationName=$filterValue"
+                            window.location.href = filterValue?.let { "${window.location.href.substringBefore("?")}?organizationName=$filterValue" }
+                                ?: window.location.href.substringBefore("?")
```
petertrr, line 152:
```suggestion
                            window.location.href = buildString {
                                append(window.location.href.substringBefore("?"))
                                filterValue?.let { append("?organizationName=$filterValue")
                            }
```

petertrr (approved):

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/contests/ContestGlobalRatingView.kt
Arrgentum: Done

PR #1178, opened 2022-09-07
Fix Diktat/Spotless warnings in test code

0x6675636b796f75676974687562:


petertrr (approved):

nulls (approved):

save-agent/src/linuxX64Test/kotlin/com/saveourtool/save/agent/SaveAgentTest.kt
```diff
@@ -22,13 +22,6 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.PolymorphicSerializer
 
 class SaveAgentTest {
-    init {
```
petertrr, line 25: Order of initialization is essential for this test to work, unfortunately. That's why CI is failing.
Because of poor API of `kotlin-test`, there is no other way to call a `BeforeAll` kind of logic.

akuleshov7 (approved):

.gitignore
```diff
@@ -10,3 +10,6 @@ save-cloud-charts/save-cloud/charts
 *.log
 *.jar
 /save-backend/files/testSuites/CQFN.org/Standard+test+suites/*/*.zip
+
+# JPA Buddy
```
akuleshov7, line 14: you use JPA Buddy?

0x6675636b796f75676974687562: Well, yes &mdash; I decided to give it a try.

PR #1182, opened 2022-09-07
Create `LnkContestExecution` whenever a new execution is submitted under a contest

petertrr:
* Create `LnkContestExecution` whenever a new execution is submitted under a contest
  * Move `testingType` from request params to `RunExecutionRequest`
  * Pass `contestName` if `testingType` is `CONTEST_MODE`
* Check if the project is enrolled into a contest when the execution request is submitted
* Support specifying contest name in save-api


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/LnkContestExecutionService.kt
```diff
@@ -49,4 +54,20 @@ class LnkContestExecutionService(
...
+    fun createLink(execution: Execution, contestName: String) = blockingToMono {
+        contestRepository.findByName(contestName)
+    }
+        .filter { it.isPresent }
+        .map {
+            lnkContestExecutionRepository.save(
+                LnkContestExecution(execution = execution, contest = it.get())
+            )
+        }
+        .switchIfEmptyToNotFound()
```
sanyavertolet, line 72: Are you sure it should be here but not before `map`?

sanyavertolet (approved):

nulls (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/RunExecutionRequest.kt
```diff
@@ -23,4 +26,13 @@ data class RunExecutionRequest(
     val sdk: Sdk,
     val execCmd: String? = null,
     val batchSizeForAnalyzer: String? = null,
-)
+
+    val testingType: TestingType,
+    val contestName: String? = null,
```
nulls, line 31: a suggestion: try to use ID instead of name to simplify searching entity on backend side

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/LnkContestExecutionService.kt
petertrr: Doesn't really matter in this case, but I've changed for better readability

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/RunExecutionRequest.kt
petertrr: I think that this would make API less clear, because e.g. in save-api-cli we would need to make two requests: at first get `Contest` and only then submit an execution request. Also, `contest` table will likely be not very complicated for exact method of searching to matter.

Issue #973, opened 2022-07-27
In organization settings make it clearer why the git url from git credentials list is not available for modification


kgevorkyan:
![1](https://user-images.githubusercontent.com/26010098/181219615-73477278-d351-444f-9d8f-278e84262fbf.PNG)

petertrr:
@nulls I believe it's your implementation, what are the reasons for this code?
https://github.com/saveourtool/save-cloud/blob/3a7cd4eab11e032d3870e9623402160602c6fde2/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/GitWindow.kt#L91

nulls:
git url is unique key for git dto -- modification of it means remove one and create a new one.
there is another button:  `Add new`. It allows to set git url and credentials

petertrr:
> git url is unique key for git dto

Yeah, that makes sense. I think we should make it more clear for the user, though. For example, a tooltip on the frontend, that would describe that this field is immutable, or we can just make it look differently - not like a disabled `input` but like a textfield of something.

PR #1186, opened 2022-09-07
Git validation on creation

### LINKED ISSUE #380: Prohibit the creation of the project in case of invalid credentials were provided
Now we only have a 'validate git connection' button, but this check should be (?) used to prevent the creation of a project.
There should be no validation, if the user did not provide us the credentials and url

nulls:
Git validation on creation

It closes #380 and #973


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
```diff
@@ -0,0 +1,10 @@
+/**
+ * This file contains util methods for kotlin
+ */
+
+package com.saveourtool.save.utils
+
+/**
+ * @return current value cast to [R] or null if casting is not possible
+ */
+inline fun <T : Any, reified R : Any> T.castOrNull(): R? = takeIf { it is R }?.let { it as R }
```
petertrr, line 10: Does it do the same as `as?` operator?

petertrr (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
nulls: yeap, read documentation, but missed that: https://kotlinlang.org/docs/typecasts.html#safe-nullable-cast-operator

PR #1187, opened 2022-09-08
Display global rating on ContestGlobalRatingView

petertrr:
* Change returned type in `OrganizationController` from `Organization` to `OrganizationDto`
* Add rating in `OrganizationDto`
* Display rating on frontend

Part of #1115


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -85,8 +85,20 @@ internal class OrganizationController(
         description = "Get non-deleted organizations.",
     )
     @ApiResponse(responseCode = "200", description = "Successfully fetched non-deleted projects.")
```
kgevorkyan, line 87:
```suggestion
    @ApiResponse(responseCode = "200", description = "Successfully fetched non-deleted organizations.")
```

kgevorkyan (approved):

sanyavertolet (approved):

PR #1194, opened 2022-09-09
## Whats added

Arrgentum:
- Edit error in regex url address classification


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/ProjectMenuBar.kt
```diff
@@ -16,7 +16,7 @@ enum class ProjectMenuBar {
 
     companion object : TabMenuBar<ProjectMenuBar> {
         // The string is the postfix of a [regexForUrlClassification] for parsing the url
-        private val postfixInRegex = values().map { it.name.lowercase() }.joinToString { "|" }
+        private val postfixInRegex = values().joinToString("|") { it.name.lowercase() }
```
petertrr, line 19: 👀

petertrr (approved):

Issue #805, opened 2022-06-02
Support testing of redirects logic


petertrr:
On frontend, we often use redirects by modifying `window.location` directly. This approach is hard to cover by unit tests.
Another possible approach is to use `Link`/`NavLink` components from `react-router` and wrap components under test in `MemoryRouter`.
We should investigate differences between these two options and find a way to perform unit tests to check if redirects are actually called when we expect them to be.

akuleshov7:
Will never be done. Instead of weeks spent on this task much easier to test it manually

PR #1216, opened 2022-09-13
Introduce `NavigateFunctionContext`, fix incorrect link after execution submission

### LINKED ISSUE #1185: Problems with submitting execution for a contests
* `3. Enroll for a contest` - this card is confusing, because if the project is already enrolled through the contests view, the contest doesn't appear in the list and user might think that the project isn't enrolled, while it already is
* ~For projects that are not enrolled to any contest we don't show a `run as contest` tab on `ProjectView`, which, combined with the former, makes even less sense~ Turns out, I was wrong: we don't show `CONTEST_MODE` button only for private projects. Should we diplay diabled button with explanation tooltip?
* Right after submission user is redirected to `project/{organizationName}/{projectName}/run/history` which doesn't exists. This may be a known problem, though. This is caused by #1088, where URL of a run submission tab has changed, but redirect logic is still the same

petertrr:
* Introduce `NavigateFunctionContext`
* Fix incorrect link after execution submission
* Fix display of score on `ContestSubmissionsMenu.kt`
* Add comment about contests eligible for enrollment
* Set fixed version of `autoprefixer` to avoid warnings during build

Closes #1185

As noted for example in #805, we should eventually start using navigation support from react-router instead of manually changing `window.location`, doing page reloads if it doesn't work etc. This includes switching from `<a>` to `<Link>` and from assigning `window.location` to using `<Navigate>` or `useNavigate`

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RouterUtils.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RouterUtils.kt
@@ -2,11 +2,38 @@
  * Utilities to work with react-router
  */
 
+@file:Suppress("FILE_NAME_MATCH_CLASS")
+
 package com.saveourtool.save.frontend.utils
 
 import history.Location
 import react.*
 import react.router.*
+
+/**
+ * Interface that provides [NavigateFunction] to use [navigation support](https://reactrouter.com/en/v6.3.0/api#navigation)
+ * from react-router v6
+ */
+interface NavigateFunctionContext {
+    /**
+     * Function that performs navigation using react-router library
+     */
+    val navigate: NavigateFunction
+}
+
+/**
+ * @param handler DOM builder that can consume [NavigateFunctionContext]
+ */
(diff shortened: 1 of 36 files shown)


sanyavertolet (approved):
lgtm

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RouterUtils.kt
```diff
@@ -2,13 +2,38 @@
...
+     * Function that performs navigation using react-router library
+     */
+    val navigate: NavigateFunction
+}
+
+/**
+ * @param handler DOM builder that can consume [NavigateFunctionContext]
+ */
+fun <T : Props> ChildrenBuilder.withNavigate(handler: ChildrenBuilder.(NavigateFunctionContext) -> Unit) {
+    val wrapper: FC<T> = FC {
```
nulls, line 28: I don't see that `props` is being used here. Maybe use `react.VFC`

nulls (approved):

PR #1221, opened 2022-09-14
Bug on ContestView with invalid URL

### LINKED ISSUE #1165: Bug on ContestView with invalid  URL
When entering an invalid contest name in the url, we stay on the original page instead of going to 404
Contest

Cheshiriks:
* fixed bug on ContestView 
Closes #1165

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestView.kt
@@ -97,7 +97,7 @@ class ContestView : AbstractView<ContestViewProps, ContestViewState>(false) {
         } else if (props.location != prevProps.location) {
             urlAnalysis(ContestMenuBar, Role.NONE, false)
         } else if (props.currentContestName != prevProps.currentContestName) {
-            getContest()
+            fetchContest()
         }
     }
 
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestView.kt
```diff
@@ -88,20 +96,35 @@ class ContestView : AbstractView<ContestViewProps, ContestViewState>(false) {
...
     }
 
     override fun componentDidMount() {
         super.componentDidMount()
         urlAnalysis(ContestMenuBar, Role.NONE, false)
         getIsFeaturedAndSetState()
+        getContest()
+    }
+
+    private fun getContest() {
```
nulls, line 111: suggestion to rename `getContest` to `fetchContest` for example, because it's not a getter

Cheshiriks: Renamed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestView.kt
```diff
@@ -69,6 +71,11 @@ external interface ContestViewState : State, HasSelectedMenu<ContestMenuBar> {
      * Flag that shows if current contest is featured or not
      */
     var isFeatured: Boolean
+
+    /**
+     * Contest
```
petertrr, line 76: Please indicate with a comment somewhere that this field is unused but acts as a marker of contest existence

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ContestView.kt
Cheshiriks: This field is now used to get the name of the contest in the view header

**+state.contest.name**

PR #1224, opened 2022-09-14
Support new mappings in Execution deletion logic

kgevorkyan:
### What's done:
* Don't remove executions which connected with contests


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
```diff
@@ -112,7 +114,9 @@ class ExecutionController(private val executionService: ExecutionService,
         }
         .flatMap { organization ->
             projectService.findWithPermissionByNameAndOrganization(authentication, name, organization.name, Permission.READ).map {
-                executionService.getExecutionDtoByNameAndOrganization(name, organization).reversed()
+                executionService.getExecutionByNameAndOrganization(name, organization).map {
+                    it.toDto().copy(contestName = lnkContestExecutionService.findContestByExecution(it)?.name)
```
petertrr, line 118: Maybe we can check `testingType` of execution here and perform query only if `testingType == CONTEST_MODE`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
```diff
@@ -137,15 +141,15 @@ class ExecutionController(private val executionService: ExecutionService,
...
+     * Delete all, except participating in contests, executions, by project name and organization
      *
      * @param name name of project
      * @param organizationName organization of project
      * @param authentication
      * @return ResponseEntity
      * @throws NoSuchElementException
      */
-    @PostMapping(path = ["/api/$v1/execution/deleteAll"])
+    @PostMapping(path = ["/api/$v1/execution/deleteAllExceptContest"])
```
petertrr, line 152:
```suggestion
    @PostMapping(path = ["/api/$v1/execution/delete-all-except-contest"])
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
```diff
@@ -162,10 +166,13 @@ class ExecutionController(private val executionService: ExecutionService,
         )
             .mapNotNull { it.id!! }
             .map { id ->
-                testExecutionService.deleteTestExecutionWithProjectId(id)
-                agentStatusService.deleteAgentStatusWithProjectId(id)
-                agentService.deleteAgentWithProjectId(id)
-                executionService.deleteExecutionByProjectNameAndProjectOrganization(name, organization)
+                val executionsNotInContests = executionService.getExecutionNotParticipatingInContestByNameAndOrganization(name, organization).map {
+                    it.id!!
```
petertrr, line 170:
```suggestion
                    it.requiredId()
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/HistoryView.kt
```diff
@@ -166,6 +166,16 @@ class HistoryView : AbstractView<HistoryProps, HistoryViewState>(false) {
                         }
                     }
                 }
+                column("contestName", "Participating in contest", { contestName ?: "N/A" }) { cellProps ->
+                    Fragment.create {
+                        td {
+                            a {
+                                href = getHrefToExecution(cellProps.row.original.id, cellProps.row.original.status, null)
```
petertrr, line 173: Seems confusing that clicking on contest name leads to execution view. Maybe, add a symbol like [this](https://fontawesome.com/icons/arrow-up-right-from-square?s=solid&f=classic) and change link's `href` to contest view for corresponding contest?

kgevorkyan: Initially I did this like for all other columns, however yes, in this case it's better to redirect to the contest itself
Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
kgevorkyan: Fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
kgevorkyan: Good, fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
kgevorkyan: Done

petertrr (approved):

PR #1226, opened 2022-09-14
Moving snapshots of test suites sources

nulls:
### What's done:
* fixed test suites source name
* move only if name is changed


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -359,12 +359,20 @@ class TestSuitesSourceController(
         .flatMap { (originalName, updatedEntity) ->
             when (val saveStatus = testSuitesSourceService.update(updatedEntity)) {
                 EntitySaveStatus.EXIST, EntitySaveStatus.CONFLICT -> Mono.just(saveStatus.toResponseEntity())
-                EntitySaveStatus.UPDATED -> testSuitesSourceSnapshotStorage.list(updatedEntity.organization.name, originalName)
-                    .map { it.copy(testSuitesSourceName = originalName) to it }
```
nulls, line 363: here was the issue: `list()` returns keys with the old name

petertrr (approved):

Issue #1167, opened 2022-09-05
Sandbox support: we need to support easy-checks for users 


akuleshov7:
1. We need to create a new VIEW, where users will be able to check their configs
2. It should be like diktat demo 
3. Has 2 input windows (config, test) and 1 output window with debug info

![image](https://user-images.githubusercontent.com/58667063/190160382-a75cbd5c-f5c5-4983-a416-edc4815e8b1f.png)

petertrr:
We also were talking about improved support in save-cli:
* Better (clear and readable) step-by-step logging: which test files have matched, why, which options have been used to extract warnings
* This information could be additionally written into a separate structured file (JSON?) that could be easily parsed for displaying in save-cloud-frontend

petertrr:
Another idea - we can base the new view on the existing one with public tests under contests. Then before submitting execution for a contest the user will be able to test their tool on a public test.

akuleshov7:
![image](https://user-images.githubusercontent.com/58667063/189941864-0dc17393-60af-4087-b49c-1a4466970159.png)

akuleshov7:
The idea is changed a little bit:

![image](https://user-images.githubusercontent.com/58667063/190159928-0943716f-f8e5-495f-9226-5b9672ccd892.png)

nulls:
- [ ] https://github.com/saveourtool/save-cloud/issues/1228

PR #1229, opened 2022-09-15
Sandbox on frontend strikes back

sanyavertolet:
This PR is connected with #1167

### What's new:
 * Reworked design - now there is only one switchable editor
 * Added light `selectorBuilder`

### How it looks now:
https://user-images.githubusercontent.com/35039155/190392271-a8b663d5-def9-4e7e-ad9d-3547d701eaca.mov


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
```diff
@@ -282,6 +270,9 @@ class SandboxView : AbstractView<SandboxViewProps, SandboxViewState>(false) {
             |    println("saveourtool loves ${ '$' }{bestLanguage.name}")
             |}
         """.trimMargin()
+        private val setupShExample = """
+            |sudo rm -rf /
```
nulls, line 274: please change this stub\example

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
sanyavertolet: :)

sanyavertolet: Done

PR #1236, opened 2022-09-19
Generalization of `orchestrator` and `save-agent`

### LINKED ISSUE #1239: Generalization of orchestrator and save-agent
Subj

nulls:
### What's done:
* Extracted all communications in `orchestrator` with `backend` to a dedicated interface
* Extracted all endpoints to `AgentConfiguration`
* Added a stub implementation for `sandbox` mode on Orchestrator -- currently it doesn't work

It closes #1239


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/sandbox/SandboxController.kt
```diff
@@ -0,0 +1,84 @@
...
+
+typealias ByteBufferFluxResponse = ResponseEntity<Flux<ByteBuffer>>
+typealias StringResponse = ResponseEntity<String>
+
+/**
+ * Sandbox implementation of endpoints which are required for save-agent
+ */
+@Profile(SANDBOX_PROFILE)
+@RestController
+@RequestMapping("/sandbox")
```
petertrr, line 25: The resulting path for all endpoints will be like `/sandbox/internal`, shouldn't it be the other way around?

petertrr (approved):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/sandbox/SandboxController.kt
nulls: it's for now only "/sandbox", but we need to have open api to communicate with `frontend` from `/sandbox` box.
but I would like to extract it to a dedicated rest controller.
Will add `suffix` to current controller

PR #1238, opened 2022-09-19
[WIP] authenticated microservices

petertrr:
* [WIP] App-to-app authentication using `ServiceAccount` tokens
* Switch `spring-cloud-kubernetes` implementation from `kubernetes-client` to `fabric8-client` in save-backend
* Improvements for Gradle build (extracted into #1237)

### DIFF
--- a/save-cloud-charts/save-cloud/templates/backend-deployment.yaml
+++ b/save-cloud-charts/save-cloud/templates/backend-deployment.yaml
@@ -18,6 +18,7 @@ spec:
       annotations:
         {{- include "pod.common.annotations" (dict "service" .Values.backend ) | nindent 8 }}
     spec:
+      serviceAccountName: microservice-sa
       restartPolicy: Always
       {{- include "cnb.securityContext" . | nindent 6 }}
       containers:
@@ -35,6 +36,8 @@ spec:
               mountPath: /home/cnb/files
             - name: database-secret
               mountPath: {{ .Values.backend.dbPasswordFile }}
+            - name: service-account-projected-token
+              mountPath: /var/run/secrets/tokens
           {{- include "spring-boot.management" .Values.backend | nindent 10 }}
           resources:
             limits:
@@ -109,3 +112,9 @@ spec:
             secretName: db-secrets
         - name: migrations-data
           emptyDir: {}
+        - name: service-account-projected-token
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: sa-token
+                  expirationSeconds: 7200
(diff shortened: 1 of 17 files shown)


save-cloud-charts/save-cloud/templates/backend-deployment.yaml
```diff
@@ -18,6 +18,7 @@ spec:
       annotations:
         {{- include "pod.common.annotations" (dict "service" .Values.backend ) | nindent 8 }}
     spec:
+      serviceAccountName: microservice-sa
```
sanyavertolet, line 21: As far as I can see, all the microservices that are able to send requests should have `microservice-sa` (e.g. `demo-cpg` does not send anything so service account for it is not required(?))

Moreover, it seems that orchestrator and demo role bindings should also reference `microservice` cluster role

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/SavePreprocessor.kt
```diff
@@ -1,17 +1,28 @@
...
  */
-@SpringBootApplication
+@SpringBootApplication(exclude = [
+    ReactiveSecurityAutoConfiguration::class,
+    ReactiveUserDetailsServiceAutoConfiguration::class,
+    ReactiveManagementWebSecurityAutoConfiguration::class,
+])
 @EnableWebFlux
 @EnableConfigurationProperties(ConfigProperties::class)
+@Import(SecurityWebClientCustomizers::class)
```
sanyavertolet, line 25: It seems that all the microservices should also be configured with `SecurityWebClientCustomizers`. What if `ktor` client is used? Should we implement some kind of plugin that adds required header?

save-cloud-charts/save-cloud/templates/backend-deployment.yaml
petertrr: The idea was, that communication between some microservices can still be secure without security, because we were planning to account only for not trusted executed tools and their network activity could be restricted with a `NetworkPolicy` on agent pods. This way using SA token for authentication is only required for microservices which receive requests from agents. But then, microservices that send requests to these services also need to authenticate, so it may be easier to unify everything.

sanyavertolet: As far as I can see, many things have been done: `KubernetesAuthenticaitonUtils` provide almost everything for `X-Service-Account-Token` check. So the question is: what hasn't been done yet? (regarding this PR) Seems that ktor needs to be able to add custom headers. And it seems that some `Bean`s are required (e.g. `KubernetesClient` bean in `KubernetesAuthenticationUtils`). Am I right?

petertrr: I suppose it was already more or less working. This PR affects only services, not agents, and I'm not sure if we have ktor on server-side. But yes, the part that validates token against k8s api server requires `KubernetesClient`. The part that adds it as a custom header requires the token to be mounted (added to YAML spec as `projected volume`)

PR #1242, opened 2022-09-19
Synchronize Agent ID and container name

kgevorkyan:
### What's done:
* Add container name to agent entity
* Display container name on Frontend


db/v-2/tables/agent.xml
```diff
@@ -0,0 +1,17 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="add-addContainerName" author="kgevorkyan" context="dev or prod">
+        <sql>
+            ALTER TABLE agent ADD COLUMN container_name text;
+            UPDATE agent SET container_name = 'unknown' WHERE container_name IS NULL;
+            ALTER TABLE agent MODIFY container_name text NOT NULL;
+        </sql>
```
petertrr, line 13: `<addColumn>` with default value?

save-agent/src/linuxX64Main/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -319,6 +319,7 @@ class SaveAgent(private val config: AgentConfiguration,
                         tr.resources.test.toString(),
                         pluginExecution.plugin,
                         config.id,
+                        config.name,
```
petertrr, line 322: Do we need agent to know its name? Or orchestrator can manage all data based on ID?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -81,7 +83,8 @@ class AgentsController(
                 .flatMap { agentIds ->
                     agentService.saveAgentsWithInitialStatuses(
                         agentIds.map { id ->
-                            Agent(id, execution)
+                            val containerName = dockerClient.inspectContainerCmd(id).exec().name
```
petertrr, line 86: Let's hide this logic in `DockerAgentRunner` via `DockerService`, because `DockerClient` is missing in Kuberenetes and orchestrator will simply fail to start

petertrr (changes requested):
Please support this logic in `kubernetes` profile too; there name and ID can simply be equal and set in `KubernetesManager` into agent's env

db/v-2/tables/agent.xml
kgevorkyan: Changed

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
kgevorkyan: But in this case we will need to change signature of methods, and return containerName from `createContainerFromImage()`, for example, that's what you mean?

kgevorkyan:
> Please support this logic in kubernetes profile too; there name and ID can simply be equal and set in KubernetesManager into agent's env

Do you mean `agentIdEnv`? But I don't see any suitable fields in `EnvVar()`

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -213,6 +212,7 @@ class DockerAgentRunner(
         val envFile = createTempDirectory("orchestrator").resolve(".env").apply {
             writeText("""
                 ${AgentEnvName.AGENT_ID.name}=$containerId
+                ${AgentEnvName.AGENT_NAME.name}=$containerName
```
nulls, line 215: `containerName` is available before running the docker agent, it can be moved to `.withEnv()`

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/docker/DockerAgentRunner.kt
```diff
@@ -168,6 +167,13 @@ class DockerAgentRunner(
         val baseImageTag = configuration.imageTag
         val runCmd = configuration.runCmd
         val envFileTargetPath = "$SAVE_AGENT_USER_HOME/.env"
+        val envVariables = configuration.env.map { (key, value) ->
```
nulls, line 170:
```suggestion
        val envVariables = configuration.env.map { (key, value) ->
                    "$key=$value"
                } + "${AgentEnvName.AGENT_NAME.name}=$containerName"
```

kgevorkyan: Thanks, changed

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Agent.kt
```diff
@@ -9,11 +9,14 @@ import javax.persistence.ManyToOne
  * @property containerId id of the container, inside which the agent is running
  * @property execution id of the execution, which the agent is serving
  * @property version
+ * @property containerName
```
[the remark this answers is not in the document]
kgevorkyan: Fixed

nulls:
> Please support this logic in `kubernetes` profile too; there name and ID can simply be equal and set in `KubernetesManager` into agent's env

We use `Pod::metadata.name`, but there is `Pod::metadata.uid`. Looks like we already put docker's analog of docker's name and kubernetes api uses `Pod::metadata.name`.
We can:
* store `Pod::metadata.name` as `container_id` and `container_name` 
* **or** remove `container_id` at el and store `docker's name` only and get container by:
```kotlin
private fun getContainerIdByName(containerName: String): String = dockerClient
    .listContainersCmd()
    .withNameFilter(setOf(containerName))
    .exec()
    .firstOrNull()
    ?.id
    ?: throw NotFoundException("There is no container with name $containerName")
```

@petertrr what do you think about it?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
petertrr: Yes, something like this

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
```diff
@@ -74,7 +76,8 @@ class AgentsController(
                 .flatMap { containerIds ->
                     agentService.saveAgentsWithInitialStatuses(
                         containerIds.map { containerId ->
-                            AgentDto(containerId, request.executionId)
+                            val containerName = dockerClient.inspectContainerCmd(containerId).exec().name
```
nulls, line 79: need to move it to `com.saveourtool.save.orchestrator.runner.AgentRunner` (in `dockerClient`) and implement this only for `DockerAgentRunner` and return `containerId` for `KubernetersManager`

nulls (changes requested):

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
nulls: We need to put `AgentEnvName.AGENT_NAME.name` with the same value as `AgentEnvName.AGENT_ID.name`.
I think we can add to `KubernetersManager`
```kotlin
private val agentIdEnv = EnvVar().apply {
    name = AgentEnvName.AGENT_ID.name
    valueFrom = EnvVarSource().apply {
        fieldRef = ObjectFieldSelector().apply {
            fieldPath = "metadata.name"
        }
    }
}
private val agentNameEnv = EnvVar().apply {
    name = AgentEnvName.AGENT_NAME.name
    valueFrom = agentIdEnv.valueFrom
}
```

nulls (approved):

save-sandbox/src/test/kotlin/com/saveourtool/save/orchestrator/controller/agents/AgentsControllerTest.kt
```diff
@@ -62,6 +63,10 @@ class AgentsControllerTest {
         )
         whenever(dockerService.createContainers(any(), any()))
             .thenReturn(listOf("test-agent-id-1", "test-agent-id-2"))
+
+        whenever(agentRunner.getContainerIdentifier(any())).thenReturn("save-test-agent-id-1")
```
nulls, line 67: twice `whenever(agentRunner.getContainerIdentifier(any()))`, does it work?

save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/controller/AgentsController.kt
kgevorkyan: Thank you!

save-sandbox/src/test/kotlin/com/saveourtool/save/orchestrator/controller/agents/AgentsControllerTest.kt
kgevorkyan: It did somehow
Changed to one invocation

petertrr (approved):

PR #1246, opened 2022-09-19
Add storages for sandbox mode

### LINKED ISSUE #1240: [Sandbox] Add storage for files and test suites sources on `orchestrator`
Need to introduce storage(s) for file and test suites source on `orchestrator` for `sandbox` mode (a special profile in `orchestrator`).
We should re-use implementation from `backend`
We should support only a single version of each files and a single version of test suites source.

Project doesn't exist, so `organization name` should be `username`, project name -- some stub value

nulls:
It closes #1240


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/DockerService.kt
```diff
@@ -170,7 +170,7 @@ class DockerService(
             runCmd = listOf(
                 "sh", "-c",
                 "set -o xtrace" +
-                        " && curl -vvv -X POST \$${AgentEnvName.GET_AGENT_LINK.name} --output $SAVE_AGENT_EXECUTABLE_NAME" +
+                        " && curl -vvv -X POST \$${AgentEnvName.GET_AGENT_LINK.name}?agentId=\$${AgentEnvName.AGENT_ID.name} --output $SAVE_AGENT_EXECUTABLE_NAME" +
```
nulls, line 173: unfortunately,  `AgentEnvName.AGENT_ID` is not available here.
potentially, we can create a `.run` as executable file which will be run. but this file should be runnable
it requires to mark copied file to docker as executable.

nulls: reverted

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxInternalController.kt
```diff
@@ -20,15 +39,21 @@ typealias StringResponse = ResponseEntity<String>
...
     /**
      * @param agentVersion
      * @return Mono with empty body
      */
     @PostMapping("/saveAgentVersion")
     fun saveAdditionalData(
         @RequestBody agentVersion: AgentVersion
-    ): Mono<Unit> = Mono.empty()
+    ): Mono<Unit> = {
+        // Do nothing for now
```
akuleshov7, line 55: Todo function may be here?

akuleshov7 (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxInternalController.kt
nulls: nope, it's normal implementation. no need to save agent version for sandbox mode

save-sandbox/src/main/resources/application.properties
```diff
@@ -15,8 +16,8 @@ sandbox.docker.runtime=runsc
 sandbox.agents-count=3
 sandbox.agents-heart-beat-timeout-millis=60000
 sandbox.heart-beat-inspector-interval=10
-sandbox.agent-settings.backend-url=http://host.docker.internal:5800
-sandbox.agent-settings.orchestrator-url=http://host.docker.internal:5100
+sandbox.agent-settings.backend-url=http://host.docker.internal:${server.port}
```
petertrr, line 19: This should be backend's port, but will be resolved as sandbox's, won't it?

petertrr: Oh, I see, both should be equal, unlike in regular orchestrator/agent

PR #1247, opened 2022-09-19
[WIP] Authenticated agent

### LINKED ISSUE #1049: Restrict network access of the executed tool
(https://gvisor.dev/docs/architecture_guide/security/)
> A sandbox is not a substitute for a secure architecture.

save-cloud services should either be able to authorize against each other (including agent), or the tested tool should have restricted network access (i.e. being able to communicate only with targets outside of cluster IP range)

Plan:
* [ ] Add a `NetworkPolicy` to allow access from save-agent pods only to main ports of backend, orchestrator and sandbox (#1247)
* [x] Move `/actuator` and `/swagger-ui` to a separate management port (#1342)
* [ ] Make requests between services authorized using `ServiceAccount` tokens (#1238)
* [ ] Make requests from save-agent use ServiceAccount token; hide the token from the tested tool (also #1247)

petertrr:
(partially TODO):
* Run tested tool as a dedicated user `save-executor`
* Authorize requests from save-agent using ServiceAccount token
* Create a NetworkPolicy that would restrict network access of save-agent pods only to orchestrator, kube-dns and the Internet

Closes #1049

### DIFF
--- a/save-deploy/base-images/Dockerfile
+++ b/save-deploy/base-images/Dockerfile
@@ -14,9 +14,13 @@ RUN if [ "$BASE_IMAGE_NAME" = "python" ]; then \
     fi
 
 RUN groupadd --gid 1100 save-agent && \
+    groupadd --gid 1200 save-executor && \
     useradd --uid 1100 --gid 1100 --create-home --shell /bin/sh save-agent && \
+    useradd --uid 1200 --gid 1100 --create-home --shell /bin/sh save-agent && \
+    usermod -aG save-executor save-agent && \
     # `WORKDIR` directive creates the directory as `root` user unless the directory already exists
     mkdir /home/save-agent/save-execution && \
-    chown -R 1100:1100 /home/save-agent/save-execution
+    chown -R 1100:1100 /home/save-agent/save-execution && \
+    echo 'save-agent ALL = (save-executor) ALL' >> /etc/sudoers
 USER save-agent
 WORKDIR /home/save-agent/save-execution
--- a/save-cloud-charts/save-cloud/templates/save-agent-network-policy.yaml
+++ b/save-cloud-charts/save-cloud/templates/save-agent-network-policy.yaml
@@ -1,0 +1,31 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  namespace: save-agent-network-policy
+spec:
+  podSelector:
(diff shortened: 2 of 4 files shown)


nulls:
@sanyavertolet, a good enhancement for save-demo-agent too

sanyavertolet:
> @sanyavertolet, a good enhancement for save-demo-agent too

Totally agree, will take a look

save-deploy/base-images/Dockerfile
```diff
@@ -14,9 +14,13 @@ RUN if [ "$BASE_IMAGE_NAME" = "python" ]; then \
     fi
 
 RUN groupadd --gid 1100 save-agent && \
+    groupadd --gid 1200 save-executor && \
     useradd --uid 1100 --gid 1100 --create-home --shell /bin/sh save-agent && \
+    useradd --uid 1200 --gid 1100 --create-home --shell /bin/sh save-agent && \
+    usermod -aG save-executor save-agent && \
```
sanyavertolet, line 20: I cannot understand why do we create a new user **and** remove the key from the pod?
As far as I can see, there are two ways: 
1. Leave the key in `/var/run/secrets/kubernetes.io/serviceaccount/token` and forbid the child process of save-agent accessing it (probably create another user as you do here);
2. Forget about Unix permission system and delegate token management to orchestrating pod.

To my mind second way brings less problems as in case of orchestrator sending token to agent might cause problems with save-demo-agent (it has ktor server running and so we might need to implement auth for ktor server also). What was your idea?

Sorry for bothering and thanks in advance

petertrr: The problem with the second approach is that you'd need to reliably detect requests from not-yet-authorized agents and distinguish them from similar requests from a tested tool. One idea (I believe, that's why I've added `Agent.isAuthenticated`) was to initialize agents with `isAuthenticated = false`, then these agents would send request for a token (which they would do strictly before starting a tested tool), orchestrator would update `isAuthenticated = true` and won't return token for any new requests from this agent. However, there is no guarantee that already running tool will be able to mimic as a newly created agent and get the token.
That's why the first approach seemed more viable: The agent would read token from a file, that has restricted access for a user, that runs a tested tool.

sanyavertolet: Thanks! Now I got the idea

PR #1249, opened 2022-09-19
Enhance the client API and add integration tests

0x6675636b796f75676974687562:


akuleshov7:
The work is still in progress? Any updates?

save-api/src/main/kotlin/com/saveourtool/save/api/github/Project.kt
```diff
@@ -0,0 +1,33 @@
+package com.saveourtool.save.api.github
+
+import java.net.URL
+
+data class Project(
```
kgevorkyan, line 5: May be rename to something like `GitHubProject` so as not to overlap over existing `Project` enitity?

save-api/src/main/kotlin/com/saveourtool/save/api/impl/DefaultSaveCloudClient.kt
```diff
@@ -0,0 +1,64 @@
...
+     * @see HttpClient.get
+     */
+    private suspend fun HttpClient.getAndAssert(url: URL, binaryContent: Boolean = false): HttpResponse =
+        get {
+            url(url)
+            accept(Application.Json)
+        }.httpResponseCheck(binaryContent)
+
+    private companion object {
+        private const val DEFAULT_API_VERSION = "v1"
```
kgevorkyan, line 40: Can we use value from `ApiVersions`?

save-api/src/main/kotlin/com/saveourtool/save/api/github/Project.kt
0x6675636b796f75676974687562: Already done that in the newest version.

save-api/src/main/kotlin/com/saveourtool/save/api/impl/DefaultSaveCloudClient.kt
0x6675636b796f75676974687562: Thanks for the heads-up, fixed.

save-api/src/test/kotlin/com/saveourtool/save/api/SaveBackendTest.kt
```diff
@@ -0,0 +1,410 @@
...
+                .getOrHandle(SaveCloudError::fail)
+                .id
+            LOGGER.debug("Waiting for execution (id = $executionId) to complete...")
+
+            var execution: ExecutionDto
+            val nanos = measureNanoTime {
+                do {
+                    execution = getExecutionById(executionId)
+                        .getOrHandle(SaveCloudError::fail)
+                    delay(100L)
```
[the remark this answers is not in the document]
kgevorkyan: And may be increase delay? Or 100L is enough

kgevorkyan:
Could you please add more docs? And update README.md?

kgevorkyan (approved):

save-api/src/test/kotlin/com/saveourtool/save/api/SaveBackendTest.kt
```diff
@@ -0,0 +1,410 @@
...
+                } while (execution.status in arrayOf(PENDING, RUNNING))
+            }
+            LOGGER.debug("The execution (id = $executionId) has completed in ${nanos / 1000L / 1e3} ms.")
```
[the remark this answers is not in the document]
0x6675636b796f75676974687562: This is a shorthand for converting a nanosecond value to milliseconds with a 3-digit fractional precision.

In other words, a value of 123456789L will be converted to 123.456 (ms).

I can achieve the same with `printf()` or `String.format()`, but it'll become less readable.

save-api/src/test/kotlin/com/saveourtool/save/api/SaveBackendTest.kt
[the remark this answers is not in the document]
0x6675636b796f75676974687562: Extracted. 

`100L` is small enough I believe, so I'd rather leave the value as-is.

0x6675636b796f75676974687562:
> Could you please add more docs? And update README.md?

@kgevorkyan, definitely will do. There's still an open task of adding support for OAuth 2.0 on the client (otherwise, we can't run the test against [saveourtool.com](https://saveourtool.com)).

akuleshov7 (approved):

akuleshov7:
btw - what is "api/github"? I missed that part

0x6675636b796f75676974687562:
> btw - what is "api/github" directory in our project? I missed that part
> Github client, etc.

@akuleshov7, the API allows to upload files so they can be used by a test suite during an execution. Our usual scenario is running `java -jar ktlint ...`, so both `ktlint` and `diktat.jar` are necessary.

As agreed with @petertrr, we made it possible to download files in an unattended manner (and then upload them to the back-end), by declaring

```kotlin
        private val gitHubProjects = listOf(
            GitHubProject("saveourtool" / "diktat"),
            GitHubProject("pinterest" / "ktlint", tag = "0.46.1")
        )
```

You can see usage example [here](https://github.com/0x6675636b796f75676974687562/save-backend-tests/blob/master/src/test/kotlin/com/saveourtool/save/api/SaveBackendTest.kt#L285)

As a bonus, we test one extra REST call (`/api/v1/files/{organizationName}/{projectName}/upload`, see `DownloadFilesController#upload`).

akuleshov7:
> > btw - what is "api/github" directory in our project? I missed that part
> > Github client, etc.
> 
> @akuleshov7, the API allows to upload files so they can be used by a test suite during an execution. Our usual scenario is running `java -jar ktlint ...`, so both `ktlint` and `diktat.jar` are necessary.
> 
> As agreed with @petertrr, we made it possible to download files in an unattended manner (and then upload them to the back-end), by declaring
> 
> ```kotlin
>         private val gitHubProjects = listOf(
>             GitHubProject("saveourtool" / "diktat"),
>             GitHubProject("pinterest" / "ktlint", tag = "0.46.1")
>         )
> ```
> 
> You can see usage example [here](https://github.com/0x6675636b796f75676974687562/save-backend-tests/blob/master/src/test/kotlin/com/saveourtool/save/api/SaveBackendTest.kt#L285)
> 
> As a bonus, we test one extra REST call (`/api/v1/files/{organizationName}/{projectName}/upload`, see `DownloadFilesController#upload`).

I got it - you will use it for downloading of files from GitHub, but do we need this code in save?

0x6675636b796f75676974687562:
> I got it - you will use it for downloading of files from GitHub, but do we need this code in save?

No problem, will move it to [save-backend-tests](https://github.com/0x6675636b796f75676974687562/save-backend-tests).

PR #1257, opened 2022-09-23
Abstract from using Execution in orchestrator

### LINKED ISSUE #1244: [Sandbox] Abstract from `Execution` in `orchestrator`
`Execution` contains information to run agents in `orchestrator` and also stores information about run.
`Sandbox` doesn't need to have `Execution`, we can abstract `orchestrator` from using `Execution` directly: `executionId` should be enough for storing the result.
Need to introduce a new class: `ExecutionRequest`, it should contains only static info which required to run agents on `orchestrator`

nulls:
* renamed `RunExecutionRequest` to `CreateExecutionRequest`
* introduced `RunExecutionRequest` instead of `Execution` for `orchestrator`

It closes #1244


Cheshiriks (approved):
lgtm

PR #1266, opened 2022-09-27
Introduce `save-orchestrator-common`

### LINKED ISSUE #1259: [Sandbox] Extract common part from `save-orchestrator` and `save-sandbox`
Need to introduce `save-orchestrator-common` and reuse it in `save-orchestrator` and `save-sandbox`

nulls:
It closes #1259


save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/HeartBeatInspector.kt
```diff
@@ -25,7 +25,7 @@ typealias AgentStateWithTimeStamp = Pair<String, Instant>
  * TODO: can be used to store data about existing agents on orchestrator startup ([#11](https://github.com/saveourtool/save-cloud/issues/11))
  */
 @Component
-@PropertySource("classpath:application.properties")
+@PropertySource("classpath:META-INF/save-orchestrator-common/application.properties")
```
nulls, line 28: looks like it's not required

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/HeartBeatInspector.kt
```diff
@@ -25,7 +24,6 @@ typealias AgentStateWithTimeStamp = Pair<String, Instant>
  * TODO: can be used to store data about existing agents on orchestrator startup ([#11](https://github.com/saveourtool/save-cloud/issues/11))
  */
 @Component
-@PropertySource("classpath:application.properties")
```
kgevorkyan, line 28: Worked without it?
How crone task `run()` in this class will determine necessary property?

nulls: This cron task is part of spring boot application. All properties will be loaded by Spring Configuration

petertrr (approved):

PR #1270, opened 2022-09-29
Refactoring `save-orchestrator-common`

nulls:
`save-agent` needs to know only url to where need to post heartbeats and result of run

PR #1280, opened 2022-10-03
AgentRepository in Sandbox

### LINKED ISSUE #1241: [Sandbox] Add AgentRepository in `orchestrator`
Need to introduce `AgentRepository` for `sandbox` mode in `orchestrator`.
It should use a new database, also need to add endpoints to check content there

nulls:
Added implementation of `AgentRepository` in Sandbox

It closes #1241


Cheshiriks:
Maybe then the BaseEntity class can also be transferred to the _Spring_ directory?

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/service/SandboxAgentRepository.kt
```diff
@@ -1,56 +1,174 @@
...
-    override fun getNextTestBatch(agentId: String): Mono<TestBatch> = Mono.empty()
+    override fun getNextTestBatch(containerId: String): Mono<TestBatch> = getExecutionAsMonoByContainerId(containerId)
+        .filter { !it.initialized }
+        .map { execution -> sandboxExecutionRepository.save(execution.apply { initialized = true }) }
+        .flatMap { execution ->
+            sandboxStorage.list(execution.userId, SandboxStorageKeyType.TEST)
+                .map { it.fileName }
+                .map { fileName ->
+                    TestDto(
+                        filePath = fileName,
```
nulls, line 75: actually `TestBatch` needs only `filePath`

nulls:
> Maybe then the BaseEntity class can also be transferred to the _Spring_ directory?

Moved

save-sandbox/src/main/resources/application-dev.properties
```diff
@@ -1,3 +1,6 @@
 spring.datasource.url=jdbc:mysql://localhost:3306/save_sandbox
 spring.datasource.username=root
 spring.datasource.password=123
+sandbox.url=http://host.docker.internal:${server.port}
+orchestrator.agent-settings.backend-url=http://host.docker.internal:${server.port}/sandbox
```
petertrr, line 5: `ochestrator.`? not `sandbox.`?

petertrr (approved):

save-sandbox/src/main/resources/application-dev.properties
nulls: yes, `ConfigProperties` are in common module. I didn't find how to customize prefix for it. 
So, common properties for orchestrator and sandbox have prefix orchestrator.
In future it can be resolved by introducing an interface for ConfigProperties

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/service/SandboxAgentRepository.kt
nulls: will be managed in https://github.com/saveourtool/save-cloud/pull/1270

PR #1283, opened 2022-10-04
Fixed database issues in sandbox

### LINKED ISSUE #1241: [Sandbox] Add AgentRepository in `orchestrator`

nulls:
Closes #1241

### DIFF
--- a/save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxAgentStatus.kt
+++ b/save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxAgentStatus.kt
@@ -28,7 +28,7 @@ class SandboxAgentStatus(
     @Enumerated(EnumType.STRING)
     var state: AgentState,
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "agent_id")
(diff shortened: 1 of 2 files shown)


save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxAgentStatus.kt
```diff
@@ -28,7 +28,7 @@ class SandboxAgentStatus(
     @Enumerated(EnumType.STRING)
     var state: AgentState,
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.EAGER)
```
akuleshov7, line 31: why eager here? I thought that LAZY will also ok here

akuleshov7 (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxAgentStatus.kt
nulls: We already got similar problem: there is an issue with hibernate proxy object, which loaded lazy. Eager is the simpler solution

Issue #1274, opened 2022-09-29
[Sandbox] Improve FileUploader


sanyavertolet:
Make `FileUploader` more independent and available for use in different places

PR #1289, opened 2022-10-05
Endpoints to operate with files for sandbox

nulls:
It's related to #1274

### DIFF
--- a/save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
+++ b/save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
@@ -7,9 +7,9 @@ import com.saveourtool.save.sandbox.entity.SandboxExecution
 import com.saveourtool.save.sandbox.repository.SandboxExecutionRepository
 import com.saveourtool.save.sandbox.repository.SandboxUserRepository
 import com.saveourtool.save.sandbox.service.BodilessResponseEntity
+import com.saveourtool.save.sandbox.storage.SandboxStorage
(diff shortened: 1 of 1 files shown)


save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
```diff
@@ -35,11 +35,22 @@ import javax.transaction.Transactional
...
-    val storage: Storage<SandboxStorageKey>,
+    val storage: SandboxStorage,
     val sandboxExecutionRepository: SandboxExecutionRepository,
     val sandboxUserRepository: SandboxUserRepository,
     val agentsController: AgentsController,
 ) {
+    /**
+     * @param userName
+     * @return list of available files for provided [userName]
+     */
```
sanyavertolet, line 46: Shouldn't we use SwaggerDoc here?

sanyavertolet (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
nulls: Probably we can add it for current controller

sanyavertolet: Will you do it?🥺

nulls: doing it right now

sanyavertolet (approved):

PR #1290, opened 2022-10-06
Provide save.toml as list of test files

nulls:


save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/service/SandboxAgentRepository.kt
```diff
@@ -70,6 +70,7 @@ class SandboxAgentRepository(
         .flatMap { execution ->
             sandboxStorage.list(execution.userId, SandboxStorageKeyType.TEST)
                 .map { it.fileName }
+                .filter { it.endsWith("save.toml") }
```
sanyavertolet, line 73: why `endsWith`?

sanyavertolet (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/service/SandboxAgentRepository.kt
nulls: save-cli supports list of test files or save.toml as argument to control which tests should be run.
previously I marked test files to detect only test files. Now we upload all files as TEST and select tests by `save.toml`

p.s. `endsWith` supports hierarchy of tests (which can be added in future)

sanyavertolet: Oh ok

PR #1294, opened 2022-10-06
Document external OAuth provider (GitHub, Dex) usage with local deployments

0x6675636b796f75676974687562:


Cheshiriks (approved):

save-deploy/README.md
```diff
@@ -78,6 +78,110 @@ Usually not the whole stack is required for development. Application logic is pe
...
+ 
+   * `hash`: the `bcrypt` hash of the password string:
+     ```bash
+     echo 'password' | htpasswd -BinC 16 'user' | cut -d: -f2
+     ```
+     The `htpasswd` utility is a part of `apache2-utils` package. The maximum
+     cost supported by `htpasswd` is **17**. _Dex_, on the other hand, only
+     allows values up to **16**.
+   * `username`: this is the name of the user from _Dex_ perspective only. Since
+     _Dex_ (unlike _GitHub_), provides no means to query user details (i.e.
```
petertrr, line 138: According to [this PR id Dex](https://github.com/dexidp/dex/pull/1473/files) it seems that there should be a `/userinfo` endpoint

petertrr (approved):

save-deploy/README.md
0x6675636b796f75676974687562: Thanks, will look into it later. 

I'll definitely document it if I make any progress.

PR #1301, opened 2022-10-07
Add typeButton in Buttons

Arrgentum:
## What added:
- Add field ```type = ButtonType.button``` in Buttons {} to change the default value (```submit```)

## Helps
@petertrr @sanyavertolet @Cheshiriks Please look at the changes made. Previously, these buttons did not have this field and had a default value (``submit``) (A button for sending form data to the server). It is she who breaks the logic by substituting ```"?"``` in the URL before ```"#"```


petertrr:
I think it's an okay change, because we don't rely on default button behavior anyway. Just please make sure that you've tried to click all these buttons and there have been no weird errors.

Arrgentum:
> I think it's an okay change, because we don't rely on default button behavior anyway. Just please make sure that you've tried to click all these buttons and there have been no weird errors.

I have tested everything, everything works correctly

Cheshiriks (approved):

petertrr (approved):
> I have tested everything, everything works correctly

OK then, lgtm

PR #1311, opened 2022-10-12
Avoid java.nio.file.AccessDeniedException thrown on Windows

0x6675636b796f75676974687562:
The test succeeds on Windows, because the temporary directory is now JUnit-managed.


kgevorkyan (approved):
Strange, I didn't saw any problems with this test before

sanyavertolet (approved):

PR #1313, opened 2022-10-12
Use the `agentDistro/` directory rather than its children as a CLASSPATH entry

0x6675636b796f75676974687562:
If `save-agent.kexe` gets downloaded *before* Gradle (re-)configures the project, then `agentDistro/save-agent.kexe` rather than `agentDistro/` will become a CLASSSPATH entry, resulting in an HTTP 404 when accessing `/internal/files/download-save-agent` in local deployments.

### DIFF
--- a/save-backend/build.gradle.kts
+++ b/save-backend/build.gradle.kts
@@ -107,7 +107,7 @@ dependencies {
                 "save-agent, please test them on Linux " +
                 "or put the file with name like `save-agent-*-distribution.jar` built on Linux into libs subfolder."
         )
-        runtimeOnly(fileTree("$buildDir/agentDistro").apply {
+        runtimeOnly(files("$buildDir/agentDistro").apply {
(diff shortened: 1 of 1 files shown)


akuleshov7 (approved):

save-backend/build.gradle.kts
```diff
@@ -107,7 +107,7 @@ dependencies {
                 "save-agent, please test them on Linux " +
                 "or put the file with name like `save-agent-*-distribution.jar` built on Linux into libs subfolder."
         )
-        runtimeOnly(fileTree("$buildDir/agentDistro").apply {
```
akuleshov7, line 110: Who has broken it?

0x6675636b796f75676974687562: Was broken from the very beginning.

petertrr (approved):

PR #1310, opened 2022-10-12
Support authentication in Sandbox

kgevorkyan:
### What's done:
* Support authentication in Sandbox

This PR could be consider as solution for https://github.com/saveourtool/save-cloud/issues/1281, or could be closed, if we plan to use auth in Sandox only within the https://github.com/saveourtool/save-cloud/issues/1299

PR #1314, opened 2022-10-12
Line markers for codeEditorComponent in SandboxView

### LINKED ISSUE #1300: [Sandbox] Highlight uncommitted difference
In sandbox mode we can edit the text - you need to highlight lines that were changed and not committed.

Also probably we can check if we can change line endings (Window/Linux) with existing mechanisms in code editor.

sanyavertolet:
This PR closes #1300

### What's done:
* Added few new states to tell saved file from edited
* Added `AceMarker` interface
* Implemented quite dummy diff for two strings
* Made `AceEditor` mark unmatched lines
* Updated `save-frontend.scss` to add class for `AceMarker`
* Started to use `kotlin-multiplatform-diff`
* Split tab save and load - now it happens separately
* Added confirm window if there are some uncommitted changes

### How it looks like:
https://user-images.githubusercontent.com/35039155/195398054-73e2a3f3-85cd-4706-95e2-392bee8130e5.mov

#### **upd**:
https://user-images.githubusercontent.com/35039155/195636822-840d51af-f7e9-4a5f-b892-36f2469ea239.mov

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
@@ -4,22 +4,23 @@ package com.saveourtool.save.frontend.components.basic.codeeditor
 
 import com.saveourtool.save.frontend.components.basic.cardComponent
 import com.saveourtool.save.frontend.externals.fontawesome.*
+import com.saveourtool.save.frontend.externals.reactace.AceMarker
 import com.saveourtool.save.frontend.externals.reactace.AceModes
 import com.saveourtool.save.frontend.externals.reactace.AceThemes
 import com.saveourtool.save.frontend.externals.reactace.aceBuilder
-import com.saveourtool.save.frontend.utils.buttonBuilder
-import com.saveourtool.save.frontend.utils.selectorBuilder
-import com.saveourtool.save.frontend.utils.useOnceAction
-import com.saveourtool.save.frontend.utils.withUnusedArg
+import com.saveourtool.save.frontend.utils.*
+
 import csstype.ClassName
 import react.ChildrenBuilder
 import react.FC
 import react.Props
+import react.useState
 import react.dom.html.ButtonType
 import react.dom.html.ReactHTML.button
 import react.dom.html.ReactHTML.div
 import react.dom.html.ReactHTML.h6
-import react.useState
+
+import kotlinx.js.jso
 
 /**
  * CodeEditor component
@@ -41,6 +42,11 @@ external interface CodeEditorComponentProps : Props {
      * Currently inputted text
      */
     var editorText: String
+
+    /**
+     * Text saved on backend
+     */
+    var savedEditorText: String
 
     /**
      * Currently selected [FileType]
@@ -221,6 +227,38 @@ private fun codeEditorComponent() = FC<CodeEditorComponentProps> { props ->
             }
         }
 
-        aceBuilder(props.editorText, selectedMode, selectedTheme, props.selectedFile == null, props.onEditorTextUpdate)
+        aceBuilder(
+            props.editorText,
+            selectedMode,
+            selectedTheme,
+            getAceMarkers(props.savedEditorText, props.editorText),
+            props.selectedFile == null,
+            props.onEditorTextUpdate,
+        )
     }
+}
+
+private fun getAceMarkers(oldFile: String, newFile: String): Array<AceMarker> = getChangedLinesIndices(oldFile, newFile)
+    .map { getAceMarker(it) }
+    .toTypedArray()
+
+private fun getAceMarker(lineIndex: Int): AceMarker = jso {
+    startRow = lineIndex
+    endRow = lineIndex + 1
+    startCol = 0
+    endCol = 0
+    className = "unsaved-marker"
+    type = "fullLine"
+}
+
+private fun getChangedLinesIndices(oldFile: String, newFile: String): Set<Int> {
+    val oldLines = oldFile.split("\n")
+    val newLines = newFile.split("\n")
+    val changedIndices: MutableSet<Int> = mutableSetOf()
+    oldLines.zip(newLines).forEachIndexed { index, (oldLine, newLine) ->
+        if (oldLine != newLine) {
+            changedIndices.add(index)
+        }
+    }
+    return changedIndices.toSet()
 }
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
@@ -244,11 +244,12 @@ private fun getAceMarkers(oldFile: String, newFile: String): Array<AceMarker> = 
 
 private fun getAceMarker(lineIndex: Int): AceMarker = jso {
     startRow = lineIndex
-    endRow = lineIndex + 1
+    endRow = lineIndex
     startCol = 0
-    endCol = 0
+    endCol = 1
     className = "unsaved-marker"
     type = "fullLine"
+    inFront = false
 }
 
 private fun getChangedLinesIndices(oldFile: String, newFile: String): Set<Int> {
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
@@ -252,13 +252,23 @@ private fun getAceMarker(lineIndex: Int): AceMarker = jso {
 }
 
 private fun getChangedLinesIndices(oldFile: String, newFile: String): Set<Int> {
-    val oldLines = oldFile.split("\n")
-    val newLines = newFile.split("\n")
-    val changedIndices: MutableSet<Int> = mutableSetOf()
-    oldLines.zip(newLines).forEachIndexed { index, (oldLine, newLine) ->
-        if (oldLine != newLine) {
-            changedIndices.add(index)
+    val oldLines = oldFile.split("\n").map { "$it\n" }
+    val newLines = newFile.split("\n").map { "$it\n" }
+
+    val extendedOldLines: List<String> = oldLines.extendWithEmptyStrings(maxOf(oldLines.size, newLines.size))
+    val extendedNewLines: List<String> = newLines.extendWithEmptyStrings(maxOf(oldLines.size, newLines.size))
+
+    return extendedOldLines.zip(extendedNewLines)
+        .mapIndexed { index, (oldLine, newLine) ->
+            Triple(index, oldLine, newLine)
         }
-    }
-    return changedIndices.toSet()
+        .filter { (_, oldLine, newLine) ->
+            oldLine != newLine
+        }
+        .map { (index, _, _) ->
+            index
+        }
+        .toSet()
 }
+
+private fun List<String>.extendWithEmptyStrings(requiredSize: Int) = plus(List(requiredSize - size) { "" })
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
@@ -2,32 +2,19 @@
 
 package com.saveourtool.save.frontend.components.basic.codeeditor
 
-import com.saveourtool.save.frontend.components.basic.cardComponent
 import com.saveourtool.save.frontend.externals.fontawesome.*
-import com.saveourtool.save.frontend.externals.reactace.AceMarker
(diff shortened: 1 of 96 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
```diff
@@ -221,6 +227,49 @@ private fun codeEditorComponent() = FC<CodeEditorComponentProps> { props ->
...
+    startRow = lineIndex
+    endRow = lineIndex
+    startCol = 0
+    endCol = 1
+    className = "unsaved-marker"
+    type = "fullLine"
+    inFront = false
+}
+
+private fun getChangedLinesIndices(oldFile: String, newFile: String): Set<Int> {
```
petertrr, line 255: Any plans on integrating kotlin-multiplatform-diff here? 👉👈

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
```diff
@@ -65,16 +65,31 @@ external interface SandboxViewState : State, HasSelectedMenu<ContestMenuBar> {
      */
     var codeText: String
 
+    /**
+     * Code from backend
+     */
+    var savedCodeText: String
```
petertrr, line 71: WDYT about extracting an editor window into a separate component that would store a draft (==unsaved) version of its text in its own state and update parent component's state on save? This way you could avoid duplicating all three fields in this component's state as well as always making 3 requests on save.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
```diff
@@ -208,6 +226,12 @@ class SandboxView : AbstractView<SandboxViewProps, SandboxViewState>(true) {
                     FileType.SETUP_SH -> state.setupShText
                     else -> "Press on any of the buttons above to start editing"
```
petertrr, line 227: Is this branch reachable? I think we can always initialize `selectedFile` with a certain tab

sanyavertolet: We can, but it might be not that obvious that there are three tabs. The reason FileType is nullable here is that I wanted to tell user that there is an ability to select the file he wants to delete. Moreover, it is `when` with no default case, I guess diktat would like to add `else`

petertrr: > to tell user that there is an ability to select the file

OK, makes sense

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
sanyavertolet: We have two components on the same level (children of the entire editor component): editor itself and it's toolbar. I am not sure how to pass event from one component to another horizontally.
But Idea is very nice as far as more logic that could be encapsulated is actually being encapsulated

sanyavertolet: Oh I got it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
sanyavertolet: Will do it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/reactace/ReactAce.kt
```diff
@@ -64,5 +65,18 @@ external interface AceEditorProps : Props {
     // keyboardHandler?: string;
```
Cheshiriks, line 65: Maybe delete unused parameters for now? And add them later as needed.

sanyavertolet: Well, I wanted to make it easier to support `react-ace` features as the one who needs it won't go to GitHub trying to find all the possible props.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/CodeEditorComponent.kt
sanyavertolet: Done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
sanyavertolet: Done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/FileType.kt
```diff
@@ -0,0 +1,81 @@
+package com.saveourtool.save.frontend.components.basic.codeeditor
+
+import com.saveourtool.save.frontend.externals.reactace.AceModes
+
+private val codeExample = """
```
nulls, line 5: good luck with conflict resolving

sanyavertolet: ☺️☺️☺️

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
```diff
@@ -196,30 +213,9 @@ class SandboxView : AbstractView<SandboxViewProps, SandboxViewState>(true) {
...
-                            FileType.CODE -> codeText = it
-                            FileType.SAVE_TOML -> configText = it
-                            FileType.SETUP_SH -> setupShText = it
-                            else -> {}
-                        }
-                    }
-                }
-                doUploadChanges = ::uploadChanges
-                doReloadChanges = ::reloadChanges
+                currentUserInfo = props.currentUserInfo ?: UserInfo("")
```
kgevorkyan, line 218: May be create some extension method for such users `UserInfo("")`? Somewhere we use `UserInfo("Unknown")`, somewhere `UserInfo("Undefined")` and so on. May be we can unify them into `UserInfo("Guest")` or something like this?

sanyavertolet: Well maybe, but talking about this PR it will be removed when #1310 will be merged

But adding something like 
```Kotlin
UserInfo?.orDefault(): UserInfo
```
might be nice

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/EditorToolbar.kt
```diff
@@ -0,0 +1,132 @@
...
+    onRunExecution: () -> Unit,
+    onFileTypeChange: (FileType) -> Unit,
+) {
+    toolbarCard {
+        div {
+            className = ClassName("input-group")
+            div {
+                className = ClassName("input-group-prepend")
+
+                button {
```
kgevorkyan, line 58: Could we extract these `button`'s  into some methods, receiving `faIcon` and `title` as arguments?

sanyavertolet: Opened #1320, tried to do it in this PR but I didn't like the implementation I had. Will try to come up with a better idea.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
petertrr: My original suggestion was to remove text pairs (regular / saved version) so that unsaved versions are encapsulated in a new component and regular versions in the parent component are updated on save

sanyavertolet: I am not sure if your variant would be reusable enough as far as we wanted to make diktat-demo be a part of saveourtool

sanyavertolet: I am not sure if your variant would be reusable enough as far as we wanted to make diktat-demo be a part of saveourtool

petertrr: I think it's a rather far perspective actually

sanyavertolet: Then the question of sending data rases. We might want to either send one of drafts or all of them. Also it is not that clear how to pass isModified flag from codeEditorComponent to toolbar. Seems like we might need even more states in that case.

petertrr: Now you have: 3 tabs, each has 2 text fields in state. What if we add more state variables for each tab? You'll need to add 3 more overall. What if we add more tabs?

What I suggest: 1 text field in state of parent, 1 field for isModified for each tab. Each instance of CodeEditorTab (or whatever you call it) receives text as props and a callback to update parent's isModified. Draft text and whatever state we may add in the future lives only inside CodeEditorTab. Toolbar can receive callbacks to access/modify parent's isModified too. I think this is more scalable and less coupled with the specific tabs we may have.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
```diff
@@ -0,0 +1,204 @@
...
+     */
+    var doRunExecution: () -> Unit
+
+    /**
+     * Action to reload debug info
+     */
+    var doResultReload: () -> Unit
+
+    /**
+     * todo: will be removed soon
```
petertrr, line 52: A link to a relevant Github issue would be helpful if we forget to remove it soon :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/FileType.kt
```diff
@@ -0,0 +1,81 @@
...
+            |
+            |data class BestLanguage(val name = "Kotlin")
+            |
+            |fun main {
+            |    val bestLanguage = BestLanguage()
+            |    println("saveourtool loves ${'$'}{bestLanguage.name}")
+            |}
+        """.trimMargin()
+
+private val configExample = """
```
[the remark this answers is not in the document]
nulls: it will be moved to backend

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
sanyavertolet: Well, it will be removed right now as #1310 is merged

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
sanyavertolet: Ok, got what you want. In this case I would like to add **another** one component, but probably in next PR as this one is getting bigger and bigger

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/FileType.kt
sanyavertolet: Finally done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/FileType.kt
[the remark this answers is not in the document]
sanyavertolet: Removed it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
```diff
@@ -0,0 +1,200 @@
...
+                getTypedOption(fileType, setSavedCodeText, setSavedConfigText, setSavedSetupShText)(content)
+            }
+        }
+    }
+
+    val uploadTexts = useDeferredRequest {
+        FileType.values().forEach { fileType ->
+            val (urlPart, content, fileName) = getTypedOption(
+                fileType,
+                Triple("test", draftCodeText, "test"),
```
nulls, line 130: suggestion rename `code` to `test` and use FileType.prettyName as fileName

nulls: in next PR

PR #1315, opened 2022-10-12
Use the correct request parameter name (orchestrator -> backend)

0x6675636b796f75676974687562:
When sending a request to `/internal/agents/get-next-test-batch` from the orchestrator, the request parameter used is now the one expected by the back-end.

### DIFF
--- a/save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/BackendAgentRepository.kt
+++ b/save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/BackendAgentRepository.kt
@@ -36,7 +36,7 @@ class BackendAgentRepository(
 
     override fun getNextTestBatch(containerId: String): Mono<TestBatch> = webClientBackend
         .get()
-        .uri("/agents/get-next-test-batch?agentId=$containerId")
+        .uri("/agents/get-next-test-batch?containerId=$containerId")
         .retrieve()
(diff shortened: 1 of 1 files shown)


save-orchestrator/src/main/kotlin/com/saveourtool/save/orchestrator/service/BackendAgentRepository.kt
```diff
@@ -36,7 +36,7 @@ class BackendAgentRepository(
 
     override fun getNextTestBatch(containerId: String): Mono<TestBatch> = webClientBackend
         .get()
-        .uri("/agents/get-next-test-batch?agentId=$containerId")
```
akuleshov7, line 39: Could you please check sandbox also

akuleshov7: Sandbox also uses this API

0x6675636b796f75676974687562: I will.

0x6675636b796f75676974687562: Resolving for now, so that the change propagates into `master` quickly.

PR #1318, opened 2022-10-13
Moved examples to backend, forced /r in text files + fixed DebugInfo

nulls:


save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
```diff
@@ -217,7 +218,7 @@ class SandboxController(
         .flatMap { key ->
             storage.overwrite(
                 key = key,
-                content = Flux.just(ByteBuffer.wrap(content.toByteArray()))
+                content = Flux.just(ByteBuffer.wrap(content.replace("\r\n", "\n").toByteArray()))
```
sanyavertolet, line 221: What about macOS and its `\r`?

sanyavertolet (approved):
lgtm, but pls think about `\r`

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
sanyavertolet: @Cheshiriks The point is that the file is sent from some PC to our sandbox, which is LF. We need to convert all possible eol symbols (CRLF, LF, CR) to LF

akuleshov7: > System.lineSeparator() ?

backend is ran on ubuntu - System.lineSeparator - is always /n :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/SandboxView.kt
```diff
@@ -347,8 +346,7 @@ class SandboxView : AbstractView<SandboxViewProps, SandboxViewState>(true) {
         if (response.ok) {
             response.text().await()
         } else {
-            postContentAsText(urlPart, fileName, defaultValue)
-            defaultValue
+            "No content for $fileName"
```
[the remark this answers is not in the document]
nulls: return error message from response

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/SandboxController.kt
nulls: changed to `\r\n?` -- it should handle Win and Mac

PR #1327, opened 2022-10-14
Single execution per user

### LINKED ISSUE #1297: [Sandbox] Only a single execution per user
Need to add validation that there is no running execution when user run a new execution

nulls:
Added validation that there is only a single running execution per user and cleanup of DebugInfo

It closes #1297

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
@@ -169,7 +169,6 @@ private fun sandboxCodeEditorComponent() = FC<SandboxCodeEditorComponentProps> {
                     if (window.confirm("Some changes are not saved. Save and run execution?")) {
                         uploadTexts()
                         props.doRunExecution()
-                        window.alert("Successfully saved and started execution.")
                     } else {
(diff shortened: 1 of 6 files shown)


sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
```diff
@@ -169,7 +169,6 @@ private fun sandboxCodeEditorComponent() = FC<SandboxCodeEditorComponentProps> {
                     if (window.confirm("Some changes are not saved. Save and run execution?")) {
                         uploadTexts()
                         props.doRunExecution()
-                        window.alert("Successfully saved and started execution.")
                     } else {
                         window.alert("Run canceled.")
                     }
```
sanyavertolet, line 174: This movement will cause double alert if there were no changes. Please fix it

sanyavertolet (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/codeeditor/SandboxCodeEditorComponent.kt
nulls: yes, already found -- fixing it now

PR #1342, opened 2022-10-20
[Helm] Set another port as a management port for all JVM services

petertrr:
Extracted from #1238; part of #1049. This PR moves `/actuator` as well as `/swagger-ui.html` to a new port, that would be protected by a `NetworkPolicy` under #1247.


akuleshov7:
Why do we change it?

akuleshov7 (approved):

PR #1343, opened 2022-10-20
Frontend: On execution view don't show executions, participating in contest

kgevorkyan:
### What's done:
* On execution view don't show executions, participating in contest

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
@@ -115,12 +115,10 @@ class ExecutionController(private val executionService: ExecutionService,
         }
         .flatMap { organization ->
             projectService.findWithPermissionByNameAndOrganization(authentication, name, organization.name, Permission.READ).map {
-                executionService.getExecutionByNameAndOrganization(name, organization).map {
-                    if (it.type == TestingType.CONTEST_MODE) {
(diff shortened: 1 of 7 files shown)


Cheshiriks (approved):

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
```diff
@@ -115,12 +115,10 @@ class ExecutionController(private val executionService: ExecutionService,
         }
         .flatMap { organization ->
             projectService.findWithPermissionByNameAndOrganization(authentication, name, organization.name, Permission.READ).map {
-                executionService.getExecutionByNameAndOrganization(name, organization).map {
-                    if (it.type == TestingType.CONTEST_MODE) {
-                        it.toDto().copy(contestName = lnkContestExecutionService.findContestByExecution(it)?.name)
-                    } else {
-                        it.toDto()
-                    }
+                executionService.getExecutionByNameAndOrganization(name, organization).filter {
```
Cheshiriks, line 118: I think instead of a filter, you can use a query to the database by Type, but this is not necessary.

kgevorkyan: Actually, we don't know anything about `type` on frontend at this moment, only about `organizationName` and `projectName`. Also it will require changing of API in this endpoint (which also used in our integration test).
So, I think, that filtering is not as bad as it might seem

Issue #1137, opened 2022-09-02
'Secret' Organization view with possibility to delete projects


akuleshov7:
Super users (admins of SAVE) should see the list of all organizations and have possibility to delete some dummy organizations.

PR #1349, opened 2022-10-20
Enable super-admins to delete projects within an organization

0x6675636b796f75676974687562:
This partially implements #1137.

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/DeleteButton.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/DeleteButton.kt
@@ -1,0 +1,79 @@
+@file:Suppress(
+    "FILE_NAME_MATCH_CLASS",
+)
+
+package com.saveourtool.save.frontend.utils
+
+import com.saveourtool.save.frontend.components.modal.ModalDialog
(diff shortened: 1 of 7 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/DeleteButton.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * The tooltip text of this button.
+     */
+    var tooltipText: String
+
+    /**
+     * The child elements of this button.
+     */
+    var children: (ChildrenBuilder) -> Unit
```
petertrr, line 68: Maybe rename this to not confuse with `children` from `PropsWithChildren`?

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/DeleteButton.kt
0x6675636b796f75676974687562: Thanks for the heads-up.

Addressed with 0dfe34012d09d2dd2305536c519c54722bfc4eb0.

Issue #1361, opened 2022-10-24
Implement TestSuitesSource deletion


sanyavertolet:
We need to support TestsSuitesSource deletion on frontend.

PR #1362, opened 2022-10-24
OrganizationTestsMenu improvements: Ph.1

### LINKED ISSUE #1097: Improve `TESTS` tab in `OrganizationView`
For now we have two tables, one is always displayed, another one is displayed only when any test suites source is selected. The problem is that it is not obvious that after a click the second table is rendered (you literally cannot see it because the first table is big enough). Suggestion to do something with expandable rows.

sanyavertolet:
This PR closes #1097 

### What's done:
 * Replaced react-tables with custom lists on test tab of `OrganizationView`

### How it looks:
https://user-images.githubusercontent.com/35039155/197551706-8b3c3648-5b46-4dae-af4c-449b7d21ed6b.mov

### TODO:
 * Add TestSuiteSource deletion (#1361)
 * Improve buttons on TEST tab
 * Add menu for permissions browsing


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/ChildrenBuilderUtils.kt
```diff
@@ -126,13 +126,16 @@ private fun ChildrenBuilder.buttonBuilder(
...
         val builtClasses = buildString {
-            append("btn btn-")
-            if (isOutline) {
-                append("outline-")
-            }
-            append(style)
-            if (isActive) {
-                append(" active")
+            append("btn")
+            style?.let { style ->
```
nulls, line 130: maybe use `it` or `style` directly without shadowing

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/TestSuitesSourcesDisplayer.kt
```diff
@@ -0,0 +1,172 @@
...
+                li {
+                    className = ClassName("list-group-item")
+                    div {
+                        className = ClassName("clearfix")
+                        div {
+                            className = ClassName("float-left")
+                            +testSuitesSourceSnapshotKey.version
+                        }
+                        buttonBuilder(faTimesCircle, style = null, classes = "float-right btn-sm") {
+                            deleteHandler(testSuitesSourceSnapshotKey)
```
nulls, line 161: do we have a confirmation before deleting?

sanyavertolet: For now we don't, but I can use a simple `window.confirm()` for that, guess it is enough.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/ChildrenBuilderUtils.kt
sanyavertolet: Yeah, sure

petertrr:
A note about UI (probably, for future PRs): from the video it seems not intuitive for me, how navigation between the list and the selected item is performed. I.e., it's not clear that there is an expanded option and that clicking its name once again opens the list again. Probably, we could add some sort of stacking, so that expanded card looked like it's on top of the previous card, and also it may be worth adding a dedicated back button near the title. This applies also for other similar views.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/TestSuitesSourcesDisplayer.kt
sanyavertolet: Done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/ChildrenBuilderUtils.kt
sanyavertolet: Done

nulls (approved):

PR #1367, opened 2022-10-25
Support the `onlyActive` parameter for the `/organizations/all` call to avoid client-side filtering

### LINKED ISSUE #1365: Refactor the `/organizations/all` REST API call and add the `onlyActive ` boolean parameter
Originally from https://github.com/saveourtool/save-cloud/pull/1358#discussion_r1003477824

0x6675636b796f75676974687562:
Closes #1365.

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -70,6 +70,12 @@ internal class OrganizationController(
 ) {
     private val webClientToPreprocessor = WebClient.create(config.preprocessorUrl)
 
+    /**
+     * @param includeDeleted whether deleted organizations should be included in
+     *   the response.
+     *   The default is `true`.
+     * @return the list of organizations.
+     */
     @GetMapping("/all")
     @PreAuthorize("permitAll()")
     @Operation(
@@ -77,12 +83,15 @@ internal class OrganizationController(
         summary = "Get all organizations",
         description = "Get organizations",
     )
-    @Parameters(
-        Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of an organization", required = true),
-    )
     @ApiResponse(responseCode = "200", description = "Successfully fetched all registered organizations")
-    fun getAllOrganizations() = Mono.fromCallable {
-        organizationService.findAll()
+    fun getAllOrganizations(
+        @RequestParam(required = false, defaultValue = "true") includeDeleted: Boolean
+    ): Mono<List<Organization>> = Mono.fromCallable {
+        when {
+            includeDeleted -> organizationService.findAll()
+
+            else -> organizationService.getFiltered(organizationFilters = OrganizationFilters.empty)
+        }
     }
 
     @PostMapping("/not-deleted")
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -70,19 +70,28 @@ internal class OrganizationController(
 ) {
     private val webClientToPreprocessor = WebClient.create(config.preprocessorUrl)
 
+    /**
+     * @param includeDeleted whether deleted organizations should be included in
```
nulls, line 74: @sanyavertolet forces to use swagger api annotation instead of kdoc for API endpoints

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationAdminView.kt
```diff
@@ -153,18 +152,13 @@ internal class OrganizationAdminView : AbstractView<Props, OrganizationAdminStat
      */
     private suspend fun getOrganizations(): MutableList<Organization> {
         val response = get(
-            url = "$apiUrl/organizations/all",
+            url = "$apiUrl/organizations/all?includeDeleted=${false}",
```
nulls, line 155: why `${false}`? do we need to calculate it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -70,19 +70,28 @@ internal class OrganizationController(
...
         description = "Get organizations",
     )
-    @Parameters(
-        Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of an organization", required = true),
-    )
     @ApiResponse(responseCode = "200", description = "Successfully fetched all registered organizations")
-    fun getAllOrganizations() = Mono.fromCallable {
-        organizationService.findAll()
+    fun getAllOrganizations(
+        @RequestParam(required = false, defaultValue = "true") includeDeleted: Boolean
```
nulls, line 88: right now all is `created` + `deleted`, but as I got from stand-up. we are going to add `BAN` status.
maybe rename this property to somethin like `onlyActive`?

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
0x6675636b796f75676974687562: ...and invert the flag meaning.

Thanks, will do.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationAdminView.kt
0x6675636b796f75676974687562: Mainly to avoid mistyping when copying or refactoring.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
0x6675636b796f75676974687562: Addressed with 1ad377dbb7b516ab572c629fb5f184d50091a605.

PR #1399, opened 2022-10-31
Fixed contest deletion

### LINKED ISSUE #1396: Contest deletion does not work properly
Right now multi selection when deleting contests is said to be implemented, but the truth is that only the latest selected contest is deleted. Tested both locally and on saveourtool. 

### Steps to reproduce:
1. Create two contests
2. Select both of them using checkboxes (on `OrganizationContestMenu`)
3. Click `Delete selected contests` button

### Expected behavior:
Both contests are marked as `DELETED`

### Actual behavior:
Only the last selected contest is marked as `DELETED`

### PS
If you select several contests and then unselect one of them (actually, any of them, not just the last one), none of contests will be in state

Cheshiriks:
Fixed contest deletion

Closed #1396

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
@@ -26,6 +26,7 @@ import react.dom.html.ReactHTML.input
 import react.dom.html.ReactHTML.td
 import react.table.columns
 
+import kotlinx.browser.window
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
```diff
@@ -216,7 +227,7 @@ private fun organizationContestsMenu() = FC<OrganizationContestsMenuProps> { pro
         isContestCreationModalOpen,
         {
             setIsContestCreationModalOpen(false)
-            refreshTable()
+            window.location.href = it
```
sanyavertolet, line 230: Far as I remember, you'd better use `Link` (or smith else from react-router) instead of `window.location`

nulls: `react.router.Navigate` -- this one

Cheshiriks: Link, as I understand it, is an analogue of a clickable link _a_. I just need to make the transition after receiving the response _OK_.

Cheshiriks: Fixed

sanyavertolet (approved):

PR #1404, opened 2022-11-01
Fixed dropdown

### LINKED ISSUE #1402: Clicking on the user profile button does not work
It is not possible to go to the user profile using the button

Cheshiriks:
Closed #1402

### DIFF
--- a/save-frontend/webpack.config.d/optimizations.js
+++ b/save-frontend/webpack.config.d/optimizations.js
@@ -1,3 +1,11 @@
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+config.plugins.push(
+    new HtmlWebpackPlugin({
+        template: 'index.html'
+    })
+);
+
 if (config.mode === "production") {
     config.optimization = {
         // todo: use https://webpack.js.org/guides/output-management/ instead of manually adding js files into html
--- a/save-frontend/webpack.config.d/optimizations.js
+++ b/save-frontend/webpack.config.d/optimizations.js
@@ -1,11 +1,3 @@
-const HtmlWebpackPlugin = require('html-webpack-plugin');
-
-config.plugins.push(
-    new HtmlWebpackPlugin({
-        template: 'index.html'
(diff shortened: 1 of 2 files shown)


save-frontend/webpack.config.d/optimizations.js
```diff
@@ -1,11 +1,3 @@
-const HtmlWebpackPlugin = require('html-webpack-plugin');
-
-config.plugins.push(
-    new HtmlWebpackPlugin({
-        template: 'index.html'
```
petertrr, line 5: Problems with cache are back in 3.. 2..

Why did you need to remove it? Please describe the problem with the dropdown more specifically. `HtmlWebpackPlugin` is needed to resolve an EJS expression in `index.html` (inject current timestamp into a `<meta>` tag) to make file content refresh on every build and avoid stale cache. Which problems does it code with the existing code?

petertrr (changes requested):

save-frontend/webpack.config.d/optimizations.js
Cheshiriks: The dropdown has the _ariaExpanded_ parameter, which changes to _true_ on click and shows a dropdown list. Now _ariaExpanded_ doesn't change on click.

petertrr: Any clues about why this is happening? `HtmlWebpackPlugin` is called only during build time and in theory only affects `index.html`, where there are no references to Bootstrap or to Popper.

akuleshov7: > Any clues about why this is happening? `HtmlWebpackPlugin` is called only during build time and in theory only affects `index.html`, where there are no references to Bootstrap or to Popper.

We have no idea :(

Cheshiriks: Changed the logic for displaying Dropdown. It works now. But I still don't understand why webpack broke it.

Issue #877, opened 2022-07-04
`TestExecutionService`: counters in Execution are not updated


petertrr:
In method `markTestExecutionsOfAgentsAsFailed` we are marking some test executions as `INTERNAL_ERROR`, but we don't adjust counters in `Execution`. E.g., if test execution is changed from `RUNNING` to `INTERNAL_ERROR`, then `Execution.running` should be decreased

PR #1406, opened 2022-11-01
counters in Execution are not updated #877

icemachined:


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
```diff
@@ -224,34 +220,31 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
-                execution.apply {
-                    log.debug {
-                        "Updating counters in execution id=$executionId: running=$runningTests-${counters.total()}, " +
-                                "passed=$passedTests+${counters.passed}, failed=$failedTests+${counters.failed}, skipped=$skippedTests+${counters.skipped}"
-                    }
-                    runningTests -= counters.total()
-                    passedTests += counters.passed
-                    failedTests += counters.failed
-                    skippedTests += counters.skipped
+        transactionTemplate.execute {
```
petertrr, line 223: `transactionTemplate` was used here to ensure transaction is closed inside `synchronized` block. It's probably not needed since the lock is handled on the DB itself.

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/ExecutionRepository.kt
```diff
@@ -45,8 +48,10 @@ interface ExecutionRepository : BaseEntityRepository<Execution> {
     fun findTopByProjectNameAndProjectOrganizationNameOrderByStartTimeDesc(name: String, organizationName: String): Optional<Execution>
 
     /**
-     * @param project to find execution
+     * @param id if the execution
      * @return execution
      */
-    fun findTopByProjectOrderByStartTimeDesc(project: Project): Execution?
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints(QueryHint(name = "javax.persistence.lock.timeout", value = "10000"))
```
petertrr, line 55: Can we make it configurable via properties? In case the DB is too slow

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
```diff
@@ -61,13 +61,14 @@ class ExecutionService(
...
+     * @param srcExecution [Execution]
      * @param newStatus [Execution.status]
      * @throws ResponseStatusException
      */
     @Transactional
-    fun updateExecutionStatus(execution: Execution, newStatus: ExecutionStatus) {
-        log.debug("Updating status to $newStatus on execution id = ${execution.requiredId()}")
+    fun updateExecutionStatus(srcExecution: Execution, newStatus: ExecutionStatus) {
+        log.debug("Updating status to $newStatus on execution id = ${srcExecution.requiredId()}")
+        val execution = executionRepository.findWithLockingById(srcExecution.requiredId()).orElse(null).orNotFound()
```
petertrr, line 71: I suggest to make a version of `orNotFound` with `Optional` as a receiver

nulls: or change `findWithLockingById` to return `Execution?`

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
```diff
@@ -224,34 +220,31 @@ class TestExecutionService(private val testExecutionRepository: TestExecutionRep
...
+            val execution = executionRepository.findWithLockingById(executionId).orElse(null).orNotFound()
```
nulls, line 224: what's about removing all these calculation on Execution and move it to Runtime and calculate them by request using `TestExecutionDto`?
something like:
```kotlin
val counters = Counters()
testExecutionRepository.findByExecutionId(executionId)
.forEach {
    when (it.status) {
        TestResultStatus.PASSED -> counters.passed++
        TestResultStatus.FAILED -> counters.failed++
        else -> counters.skipped++
    }

    with(counters) {
        unmatchedChecks += it.unmatched.orZeroIfNotApplicable()
        matchedChecks += it.matched.orZeroIfNotApplicable()
        expectedChecks += it.expected.orZeroIfNotApplicable()
        unexpectedChecks += it.unexpected.orZeroIfNotApplicable()
    }
}
```

and then return it when it's required?
Execution can be updated finally when status is changed to terminated state (or during archiving)

akuleshov7 (commented):
@icemachined ,  this PR was not updated for a while

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
nulls: changed to `Execution?`

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
nulls: removed

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestExecutionService.kt
nulls: raised: https://github.com/saveourtool/save-cloud/issues/1480

save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/ExecutionRepository.kt
nulls: looks like we can't -- it's `javax.persistence.QueryHint` and it can't take a Spring's parameter as input

petertrr (approved):

PR #1408, opened 2022-11-02
Exclude DataSourceAutoConfiguration from api-gateway

kgevorkyan:
### What's done:
* Exclude DataSourceAutoConfiguration from api-gateway

### DIFF
--- a/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/SaveGateway.kt
+++ b/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/SaveGateway.kt
@@ -4,13 +4,15 @@ import com.saveourtool.save.gateway.config.ConfigurationProperties
 
 import org.springframework.boot.SpringApplication
 import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+import org.springframework.boot.autoconfigure.jdbc.XADataSourceAutoConfiguration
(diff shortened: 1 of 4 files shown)


api-gateway/src/main/kotlin/com/saveourtool/save/gateway/SaveGateway.kt
```diff
@@ -4,13 +4,15 @@ import com.saveourtool.save.gateway.config.ConfigurationProperties
...
 import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+import org.springframework.boot.autoconfigure.jdbc.XADataSourceAutoConfiguration
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 
 /**
  * Entrypoint for api-gateway spring boot application
  */
 @EnableConfigurationProperties(ConfigurationProperties::class)
-@SpringBootApplication
```
akuleshov7, line 13: It worked?

akuleshov7: How did you find the solution?

akuleshov7 (approved):

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/SaveGateway.kt
kgevorkyan: Yeah, tested both - locally and at box

Issue #1415, opened 2022-11-03
Create save-demo


sanyavertolet:
We need to create new Spring service for different types of `Demo`s e.g. diktat-demo etc

PR #1416, opened 2022-11-03
[Sandbox] DemoService

sanyavertolet:
This PR is connected with #1415

### What's done:
 * Added AbstractDemoService interface
 * Added DiktatDemoController
 * Added DiktatDemoService
 * Added DiktatDemoMode enum
 * Added DiktatDemoTool enum

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
@@ -1,0 +1,17 @@
+package com.saveourtool.save.demo.diktat
+
+/**
+ * Enum that represents possible demo modes in diktat-demo
+ */
+enum class DiktatDemoMode {
+    /**
+     * Run demo in fix mode
+     */
+    FIX,
+
+    /**
+     * Run demo in warn-only mode
+     */
+    WARN,
+    ;
+}
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
@@ -1,17 +1,16 @@
 package com.saveourtool.save.demo.diktat
 
+import com.saveourtool.save.demo.DemoResult
 import kotlinx.serialization.Serializable
 
 /**
  * Class that represents the output of diktat demo
  *
- * @param warnings list of warnings that were found
- * @param outputText resulting file text
- * @property warnings
- * @property outputText
+ * @property warnings list of warnings that were found
+ * @property outputText resulting file text
  */
 @Serializable
 data class DiktatDemoResult(
     val warnings: List<String>,
     val outputText: String,
-)
+) : DemoResult
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
@@ -1,0 +1,17 @@
+package com.saveourtool.save.demo.diktat
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Class that represents the output of diktat demo
+ *
+ * @param warnings list of warnings that were found
+ * @param outputText resulting file text
+ * @property warnings
+ * @property outputText
+ */
+@Serializable
+data class DiktatDemoResult(
+    val warnings: List<String>,
+    val outputText: String,
+)
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoTool.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoTool.kt
@@ -1,0 +1,17 @@
+package com.saveourtool.save.demo.diktat
+
+/**
+ * Enum that represents possible tool names in diktat-demo
+ */
+enum class DiktatDemoTool {
+    /**
+     * Run diktat
+     */
+    DIKTAT,
+
+    /**
+     * Run ktlint
+     */
+    KTLINT,
+    ;
+}
--- a/save-sandbox/build.gradle.kts
+++ b/save-sandbox/build.gradle.kts
@@ -38,6 +38,8 @@ tasks.withType<Test> {
     }
 }
 
+val diktatVersion: String = libs.versions.diktat.get()
+
 dependencies {
     implementation(projects.saveOrchestratorCommon)
     implementation(libs.zip4j)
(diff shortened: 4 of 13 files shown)


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
```diff
@@ -0,0 +1,17 @@
+package com.saveourtool.save.demo.diktat
+
+/**
+ * Enum that represents possible demo modes in diktat-demo
+ */
+enum class DiktatDemoMode {
```
kgevorkyan, line 6: Can we create some common interface, may be? From which there will be inherited our particular cases, like diktat and so on?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Class that represents the output of diktat demo
+ *
+ * @param warnings list of warnings that were found
+ * @param outputText resulting file text
+ * @property warnings
+ * @property outputText
+ */
+@Serializable
+data class DiktatDemoResult(
```
kgevorkyan, line 14: Same question

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoTool.kt
```diff
@@ -0,0 +1,17 @@
+package com.saveourtool.save.demo.diktat
+
+/**
+ * Enum that represents possible tool names in diktat-demo
+ */
+enum class DiktatDemoTool {
```
kgevorkyan, line 6: May be some more general name? Why do we need `diktat` prefix?

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/DiktatDemoController.kt
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Controller for diktat-demo
+ */
+@ApiSwaggerSupport
+@Tags(
+    Tag(name = "diktat-demo"),
+)
+@RestController
+@RequestMapping("/diktat-demo/api")
```
kgevorkyan, line 23:
```suggestion
@RequestMapping("/api/diktat-demo")
```

sanyavertolet: Later it is planned to move `diktat-demo` and other demos to a new service so the endpoint would be `/demo/api/diktat`

save-sandbox/build.gradle.kts
```diff
@@ -50,4 +52,13 @@ dependencies {
     implementation(libs.spring.boot.starter.security)
     implementation(libs.spring.security.core)
     implementation(libs.save.common.jvm)
+
+    implementation(libs.ktlint.core)
+    implementation(libs.ktlint.rulesets.standard)
+    implementation("org.cqfn.diktat:diktat-common:$diktatVersion") {
```
petertrr, line 58: I don't really like the idea of putting diktat-demo service in the same monorepo as save-cloud. At least let's make it a separate module and not mix with sandbox. In the future we could have a demo service with pluggable modules for different modes.

My reason: we had a lot of problems with dependency versions in diktat-demo, because diktat/ktlint require a very specific version of kotlin-compiler-embeddable, which should be the same as version of Kotlin compiler and Kotlin dependencies. In this case this having all versions mixed together will force us either to update save-cloud less frequently or occasionally break diktat-demo.

I would imagine something like what spotless does: a common facade which loads different tools in isolated classloaders.

sanyavertolet: The next phase includes moving all demo staff to external spring service.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
sanyavertolet: For now it's hard for me to set the limits on Demo's inputs. Some tools might include run mode and so on, some might not. Not sure we should try to unify it right now. Probably should be done in phase 3

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
sanyavertolet: The output can be unified as far as all the tools surely have result, will add `DemoResult` interface

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoTool.kt
sanyavertolet: Cause this demo is responsible for diktat and ktlint. Our diktat-demo can run two tools.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
sanyavertolet: Will add `DemoAdditionalParams` as an argument of `runDemo` function

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoResult.kt
sanyavertolet: Done

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoMode.kt
sanyavertolet: Done

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/DiktatDemoController.kt
```diff
@@ -0,0 +1,45 @@
...
+) {
+    /**
+     * @param inputFiles pair of code file and config file
+     * @param tool one of [DiktatDemoTool]
+     * @param mode one of [DiktatDemoMode]
+     * @return [DiktatDemoResult]
+     */
+    @PostMapping("/run")
+    fun runCheckDemo(
+        @RequestBody inputFiles: Pair<String, String?>,
```
petertrr, line 36: I suggest to create a class instead of a pair

petertrr (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/controller/DiktatDemoController.kt
sanyavertolet: Done

kgevorkyan (approved):

PR #1421, opened 2022-11-07
save-demo

### LINKED ISSUE #1415: Create save-demo

sanyavertolet:
This PR closes #1415

### What's done:
* Added new spring service - `save-demo`
* Moved everything connected with `demo`s  into external demo service
* Disabled security for `SaveDemo`

### DIFF
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
@@ -1,0 +1,16 @@
+package com.saveourtool.save.demo
+
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+import org.springframework.boot.autoconfigure.jdbc.XADataSourceAutoConfiguration
+
+/**
+ * An entrypoint for spring boot for save-demo
+ */
+@SpringBootApplication(exclude = [DataSourceAutoConfiguration::class, XADataSourceAutoConfiguration::class])
+open class SaveDemo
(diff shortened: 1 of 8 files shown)


save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+import org.springframework.boot.autoconfigure.jdbc.XADataSourceAutoConfiguration
+
+/**
+ * An entrypoint for spring boot for save-demo
+ */
+@SpringBootApplication(exclude = [DataSourceAutoConfiguration::class, XADataSourceAutoConfiguration::class])
```
kgevorkyan, line 11: I think you could easily remove this, since you have no JPA dependency

kgevorkyan (approved):

save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
sanyavertolet: Seems like some dependencies (probably `libs.save.common.jvm`) still carry JPA dependencies so I still should mark those configurations as excluded

petertrr: It may be `projects.saveCloudCommon`, because `save.common` is part of `save-cli`. Anyway, it's strange that we use a dependency with autoconfiguration classes there

PR #1427, opened 2022-11-09
diktat-demo on frontend

### LINKED ISSUE #1423: Support diktat-demo on frontend
Right now we have a new service: `save-demo`, which has only one demo. We need to create a view for it

sanyavertolet:
This PR closes #1423

# For now it is only accessible through `#/demo/diktat`

### What's done:
 * Disabled fix mode for some time
 * Added view for `diktat/ktlint-demo` - which is `VFC`!
 * Updated dev-server.js

### How it looks:
screenshot

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoRequest.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoRequest.kt
@@ -1,0 +1,6 @@
+package com.saveourtool.save.demo
+
+/**
+ * Interface that should be implemented by all data classes that are meant to be some demo service run request.
+ */
+interface DemoRequest
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
@@ -9,6 +9,6 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 data class DiktatDemoRunRequest(
-    val codeLines: String,
+    val codeLines: List<String>,
     val params: DiktatDemoAdditionalParams,
 ) : DemoRequest
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
@@ -1,5 +1,6 @@
 package com.saveourtool.save.demo.diktat
 
+import com.saveourtool.save.demo.DemoRequest
 import kotlinx.serialization.Serializable
 
 /**
@@ -10,4 +11,4 @@ import kotlinx.serialization.Serializable
 data class DiktatDemoRunRequest(
     val codeLines: String,
     val params: DiktatDemoAdditionalParams,
-)
+) : DemoRequest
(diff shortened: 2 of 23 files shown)


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
```diff
@@ -10,4 +11,4 @@ import kotlinx.serialization.Serializable
 data class DiktatDemoRunRequest(
     val codeLines: String,
```
nulls, line 12: maybe pass as lines instead of a single string to avoid problem with EOL

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoComponent.kt
```diff
@@ -0,0 +1,258 @@
...
+import react.dom.html.ReactHTML.label
+import react.dom.html.ReactHTML.p
+import react.dom.html.ReactHTML.strong
+
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+val diktatDemoComponent = diktatDemoComponent()
+
+private val diktatDemoDefaultCode = """
```
nulls, line 38: suggestion to move an example on backend side -- it can be different for different demo tool

nulls (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
sanyavertolet: It makes sense, will do it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoComponent.kt
sanyavertolet: I see no reason why we might want to move it to the backend as far as we do not fetch test data from backend - only result file and warnings

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoRequest.kt
```diff
@@ -0,0 +1,6 @@
+package com.saveourtool.save.demo
+
+/**
+ * Interface that should be implemented by all data classes that are meant to be some demo service run request.
+ */
+interface DemoRequest
```
Cheshiriks, line 6: Is an empty parent interface needed now if it doesn't add any common logic?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoComponent.kt
nulls: we might want/have to to fetch it from backend for demo tool which doesn't support kotlin

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoRequest.kt
sanyavertolet: Yes, as far as we require it as an argument. Require Any seems to be a bad practice. Moreover, if we want to require more, we will simply do that

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DiktatDemoRunRequest.kt
sanyavertolet: Done

PR #1428, opened 2022-11-09
Conflict status when saving data in a profile.

### LINKED ISSUE #1412: Conflict status when saving data in a profile.

Cheshiriks:
Fixed bug updating user

Close #1412


sanyavertolet (approved):

petertrr (changes requested):
Please add a unit test for this logic, since it's getting more complicated and there are three branches already. Also, in which case will `user.name` be `null`?

Cheshiriks:
> Please add a unit test for this logic, since it's getting more complicated and there are three branches already. Also, in which case will `user.name` be `null`?

Added tests.
There is a check for null at a higher level, so it is not needed here.

petertrr (approved):

Issue #527, opened 2022-01-10
Move `Project` class into `jvmMain`, introduce `ProjectDto`


petertrr:

nulls:
it's partially done in #988: `ProjectDto` and `OrganizationDto` are introduced. Need to extract `Organization` and `Project` to `jvmMain`

PR #1433, opened 2022-11-10
Removed ownerId from Organization

nulls:
It's part of #527


sanyavertolet (approved):
we have finally removed it!

nulls:
> we have finally removed it!

according to the current logic -- it's already not used.
so this change only add a missed during migration roles

Cheshiriks (approved):

nulls:
Need to fix foreign key:
```
Caused by: java.sql.SQLException: Cannot drop column 'owner_id': needed in a foreign key constraint 'fk_organization_user'
```

PR #1435, opened 2022-11-10
Run `diktat` (`ktlint`) demo as CLI

### LINKED ISSUE #1434: [DEMO] fix mode is not supported
Fix mode is not supported due some infra issues, need to fix it

sanyavertolet:
This PR closes #1434

### What's done:
* Added CliRunner interface
* Added DiktatCliRunner implementation of CliRunner which uses diktat-cli and ktlint-cli (for now, will be changed)
* Supported fix mode for diktat and ktlint

### DIFF
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DiktatDemoService.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DiktatDemoService.kt
@@ -13,9 +13,11 @@ import org.cqfn.diktat.ruleset.rules.DiktatRuleSetProvider
 import org.springframework.stereotype.Service
 
 import java.io.File
+import java.nio.file.Path
 import java.util.*
 
 import kotlin.collections.ArrayList
+import kotlin.io.path.*
 
 /**
  * Demo service implementation for ktlint-demo/diktat-demo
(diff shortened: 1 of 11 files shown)


save-demo/src/main/resources/diktat.cmd
```diff
@@ -0,0 +1,19 @@
+@echo off
```
kgevorkyan, line 1: What's the difference between `ktlint.cmd`?

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DiktatDemoService.kt
```diff
@@ -5,119 +5,62 @@ import com.saveourtool.save.demo.diktat.DiktatDemoMode
...
-
 /**
  * Demo service implementation for ktlint-demo/diktat-demo
  */
 @Service
 class DiktatDemoService : AbstractDemoService<DiktatDemoAdditionalParams, DiktatDemoResult> {
-    private fun generateFileName() = UUID.randomUUID().toString()
-    private fun generateDemoFile(generatedName: String): File = File("demo-input-$generatedName")
-    private fun generateDemoConfig(generatedName: String): File = File("demo-config-$generatedName")
+    private fun generateName() = UUID.randomUUID().toString()
```
nulls, line 19: suggestion to append `tmpDir` and set it via property

sanyavertolet: This is what will be done in the next PR. Also it is planned to use file based storage in order to store all the executables

save-demo/src/main/resources/diktat
```diff
@@ -0,0 +1,578 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
sanyavertolet: It will be removed in the next PR and replaced by downloading executables to the storage

save-demo/src/main/resources/diktat.cmd
sanyavertolet: Well if I got it right the text in these files is the same, but it depends on file name so it works fine

Cheshiriks (approved):

PR #1437, opened 2022-11-11
Add delete, recover and banned Organization logic in backend

### LINKED ISSUE #1436: Add ```banned``` logic for Organization in backend
* Add status ```BANNED``` in OrganizationStatus.kt 
* Add change organization Status

Arrgentum:
## Whats added:
- Add field ```banned``` in OrganizationStatus and ProjectStatus
- Add deleted, banned and recovery Organization logic
- Add new permission ```BAN```
- Correct tests

## Issue
This pull request close #1436

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -328,6 +328,7 @@ internal class OrganizationController(
         Parameter(name = "status", `in` = ParameterIn.QUERY, description = "type of status being set", required = true),
     )
     @ApiResponse(responseCode = "200", description = "Successfully change status an organization.")
+    @ApiResponse(responseCode = "400", description = "Invalid new status of the organization.")
     @ApiResponse(responseCode = "403", description = "Not enough permission for this action on organization.")
     @ApiResponse(responseCode = "404", description = "Could not find an organization with such name.")
     @ApiResponse(responseCode = "409", description = "There are projects connected to organization. Please delete all of them and try again.")
@@ -341,6 +342,10 @@ internal class OrganizationController(
         }
         .switchIfEmptyToNotFound {
             "Could not find an organization with name $organizationName."
+        }.filter {
+            it.status != status
+        }.switchIfEmptyToResponseException(HttpStatus.BAD_REQUEST) {
+            "Invalid new status of the organization $organizationName"
         }
         .filter {
             organizationPermissionEvaluator.hasPermissionToChangeStatus(authentication, it, status)
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -342,9 +342,11 @@ internal class OrganizationController(
         }
         .switchIfEmptyToNotFound {
             "Could not find an organization with name $organizationName."
-        }.filter {
+        }
+        .filter {
             it.status != status
-        }.switchIfEmptyToResponseException(HttpStatus.BAD_REQUEST) {
+        }
+        .switchIfEmptyToResponseException(HttpStatus.BAD_REQUEST) {
             "Invalid new status of the organization $organizationName"
         }
         .filter {
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -330,7 +330,7 @@ internal class OrganizationController(
     @ApiResponse(responseCode = "403", description = "Not enough permission for this action on organization.")
     @ApiResponse(responseCode = "404", description = "Could not find an organization with such name.")
     @ApiResponse(responseCode = "409", description = "There are projects connected to organization. Please delete all of them and try again.")
-    fun deleteOrganization(
+    fun changeOrganizationStatus(
         @PathVariable organizationName: String,
         @RequestParam status: OrganizationStatus,
         authentication: Authentication,
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -315,19 +315,19 @@ internal class OrganizationController(
             ResponseEntity.ok("Organization updated")
         }
 
-    @DeleteMapping("/{organizationName}/change-status")
+    @PostMapping("/{organizationName}/change-status")
     @RequiresAuthorizationSourceHeader
     @PreAuthorize("isAuthenticated()")
     @Operation(
-        method = "DELETE",
-        summary = "Delete existing organization.",
-        description = "Delete existing organization by its name.",
+        method = "POST",
+        summary = "Change status existing organization.",
+        description = "Change status existing organization by its name.",
     )
     @Parameters(
         Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of an organization", required = true),
     )
-    @ApiResponse(responseCode = "200", description = "Successfully deleted an organization.")
-    @ApiResponse(responseCode = "403", description = "Not enough permission for deleting this organization.")
+    @ApiResponse(responseCode = "200", description = "Successfully change status an organization.")
+    @ApiResponse(responseCode = "403", description = "Not enough permission for this action on organization.")
     @ApiResponse(responseCode = "404", description = "Could not find an organization with such name.")
     @ApiResponse(responseCode = "409", description = "There are projects connected to organization. Please delete all of them and try again.")
     fun deleteOrganization(
@@ -342,7 +342,7 @@ internal class OrganizationController(
             "Could not find an organization with name $organizationName."
         }
         .filter {
-            organizationPermissionEvaluator.permissionChangeOrganizationStatus(authentication, it, status)
+            organizationPermissionEvaluator.hasPermissionToChangeStatus(authentication, it, status)
         }
         .switchIfEmptyToResponseException(HttpStatus.FORBIDDEN) {
             "Not enough permission for this action with organization of organization $organizationName."
@@ -353,18 +353,18 @@ internal class OrganizationController(
         .switchIfEmptyToResponseException(HttpStatus.CONFLICT) {
             "There are projects connected to $organizationName. Please delete all of them and try again."
         }
-        .map {
-            when(status) {
+        .map {organization ->
+            when (status) {
                 OrganizationStatus.BANNED -> {
-                    organizationService.banOrganization(it)
+                    organizationService.banOrganization(organization)
                     ResponseEntity.ok("Organization banned")
                 }
                 OrganizationStatus.DELETED -> {
-                    organizationService.deleteOrganization(it)
+                    organizationService.deleteOrganization(organization)
                     ResponseEntity.ok("Organization deleted")
                 }
                 OrganizationStatus.CREATED -> {
-                    organizationService.recoverOrganization(it)
+                    organizationService.recoverOrganization(organization)
                     ResponseEntity.ok("Organization recovered")
                 }
             }
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -315,7 +315,7 @@ internal class OrganizationController(
             ResponseEntity.ok("Organization updated")
         }
 
-    @DeleteMapping("/{organizationName}/delete")
+    @DeleteMapping("/{organizationName}/change-status")
     @RequiresAuthorizationSourceHeader
     @PreAuthorize("isAuthenticated()")
     @Operation(
@@ -332,6 +332,7 @@ internal class OrganizationController(
     @ApiResponse(responseCode = "409", description = "There are projects connected to organization. Please delete all of them and try again.")
     fun deleteOrganization(
         @PathVariable organizationName: String,
+        @RequestParam status: OrganizationStatus,
         authentication: Authentication,
     ): Mono<StringResponse> = Mono.just(organizationName)
         .flatMap {
@@ -341,20 +342,32 @@ internal class OrganizationController(
             "Could not find an organization with name $organizationName."
         }
         .filter {
-            organizationPermissionEvaluator.hasPermission(authentication, it, Permission.DELETE)
+            organizationPermissionEvaluator.permissionChangeOrganizationStatus(authentication, it, status)
         }
         .switchIfEmptyToResponseException(HttpStatus.FORBIDDEN) {
-            "Not enough permission for deletion of organization $organizationName."
+            "Not enough permission for this action with organization of organization $organizationName."
         }
         .filter {
-            !organizationService.hasProjects(organizationName)
+            status == OrganizationStatus.DELETED && !organizationService.hasProjects(organizationName)
         }
         .switchIfEmptyToResponseException(HttpStatus.CONFLICT) {
             "There are projects connected to $organizationName. Please delete all of them and try again."
         }
         .map {
-            organizationService.deleteOrganization(it.name)
-            ResponseEntity.ok("Organization deleted")
+            when(status) {
+                OrganizationStatus.BANNED -> {
+                    organizationService.banOrganization(it)
+                    ResponseEntity.ok("Organization banned")
+                }
+                OrganizationStatus.DELETED -> {
+                    organizationService.deleteOrganization(it)
+                    ResponseEntity.ok("Organization deleted")
+                }
+                OrganizationStatus.CREATED -> {
+                    organizationService.recoverOrganization(it)
+                    ResponseEntity.ok("Organization recovered")
+                }
+            }
         }
 
     @GetMapping("/{organizationName}/list-git")
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -345,10 +345,10 @@ internal class OrganizationController(
             organizationPermissionEvaluator.hasPermissionToChangeStatus(authentication, it, status)
         }
         .switchIfEmptyToResponseException(HttpStatus.FORBIDDEN) {
-            "Not enough permission for this action with organization of organization $organizationName."
+            "Not enough permission for this action with organization $organizationName."
         }
         .filter {
-            status == OrganizationStatus.DELETED && !organizationService.hasProjects(organizationName)
+            status != OrganizationStatus.DELETED || !organizationService.hasProjects(organizationName)
         }
         .switchIfEmptyToResponseException(HttpStatus.CONFLICT) {
             "There are projects connected to $organizationName. Please delete all of them and try again."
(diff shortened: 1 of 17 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -341,20 +342,32 @@ internal class OrganizationController(
             "Could not find an organization with name $organizationName."
         }
         .filter {
-            organizationPermissionEvaluator.hasPermission(authentication, it, Permission.DELETE)
+            organizationPermissionEvaluator.hasPermissionToChangeStatus(authentication, it, status)
         }
         .switchIfEmptyToResponseException(HttpStatus.FORBIDDEN) {
-            "Not enough permission for deletion of organization $organizationName."
+            "Not enough permission for this action with organization of organization $organizationName."
```
sanyavertolet, line 348: Fix the message please

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -341,20 +342,32 @@ internal class OrganizationController(
...
         }
         .filter {
-            !organizationService.hasProjects(organizationName)
+            status == OrganizationStatus.DELETED && !organizationService.hasProjects(organizationName)
```
sanyavertolet, line 351: So if `status` is not `OrganizationStatus.DELETED` user will get message `There are projects connected to $organizationName. Please delete all of them and try again"`? Is it what was planned or did I miss something?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -341,20 +342,32 @@ internal class OrganizationController(
...
+                    organizationService.banOrganization(organization)
+                    ResponseEntity.ok("Organization banned")
+                }
+                OrganizationStatus.DELETED -> {
+                    organizationService.deleteOrganization(organization)
+                    ResponseEntity.ok("Organization deleted")
+                }
+                OrganizationStatus.CREATED -> {
+                    organizationService.recoverOrganization(organization)
+                    ResponseEntity.ok("Organization recovered")
```
sanyavertolet, line 368: Please do something like `Successfully ***ACTION*** the organization.`

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
```diff
@@ -123,3 +145,6 @@ class OrganizationPermissionEvaluator(
         val contestCreatorMinimalRole = Role.ADMIN
     }
 }
+
```
sanyavertolet, line 148: It is not that obvious when the receiver is `OrganizationStatus`. Can status be banned (meaning someone banned the status)?

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/TestSuitePermissionEvaluator.kt
```diff
@@ -30,7 +30,7 @@ class TestSuitePermissionEvaluator(
     ): Boolean = lnkOrganizationTestSuiteService.getDto(organization, testSuite).rights.let { currentRights ->
         when (permission) {
             Permission.READ -> testSuite.isPublic || canAccessTestSuite(currentRights)
-            Permission.WRITE, Permission.DELETE -> canMaintainTestSuite(currentRights)
+            Permission.WRITE, Permission.DELETE, Permission.BAN -> canMaintainTestSuite(currentRights)
```
sanyavertolet, line 33: Probably `Permission.BAN` should lead to `IllegalStateException` in this case

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/LnkUserOrganizationControllerTest.kt
```diff
@@ -220,6 +220,7 @@ class LnkUserOrganizationControllerTest {
                 Permission.READ -> organizationRole.priority >= Role.VIEWER.priority
                 Permission.WRITE -> organizationRole.priority >= Role.ADMIN.priority
                 Permission.DELETE -> organizationRole.priority >= Role.OWNER.priority
+                Permission.BAN -> organizationRole.priority== Role.OWNER.priority
```
sanyavertolet, line 223: Probably something went wrong here. Shouldn't it be `Role.SUPER_ADMIN`?

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/PermissionControllerTest.kt
```diff
@@ -255,6 +255,7 @@ class PermissionControllerTest {
                 Permission.READ -> permission != null
                 Permission.WRITE -> permission == Permission.WRITE || permission == Permission.DELETE
                 Permission.DELETE -> permission == Permission.DELETE
+                Permission.BAN -> permission == Permission.BAN
```
sanyavertolet, line 258: Looks suspicious

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationAdminView.kt
```diff
@@ -170,15 +165,16 @@ internal class OrganizationAdminView : AbstractView<Props, OrganizationAdminStat
      * the externally supplied [ErrorHandler].
      *
      * @param organization the project to delete.
-     * @return the lambda which deletes [organization].
+     * @return the lambda which delentes [organization].
```
sanyavertolet, line 168: fix typo

sanyavertolet (commented):
Much better

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -43,25 +43,70 @@ class OrganizationService(
      * Mark organization with [organizationName] as deleted
      *
      * @param organizationName an [Organization]'s name to delete
+     * @param newStatus is new status for [organizationName]
      * @return deleted organization
      */
     @Suppress("UnsafeCallOnNullableType")
-    fun deleteOrganization(organizationName: String): Organization = getByName(organizationName)
+    fun changeOrganizationStatus(organizationName: String, newStatus: OrganizationStatus): Organization = getByName(organizationName)
```
Cheshiriks, line 50: Why do we need to make a request to the database twice to get the `organization`? Let's pass the organization here, not the `organizationName`.

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
Arrgentum: Yes, thanks, I changed it to
``status != OrganizationStatus.DELETED || !organizationService.hasProjects(organizationName)```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
Arrgentum: fixed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
Arrgentum: done

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
sanyavertolet: See it is also done

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
```diff
@@ -33,6 +34,23 @@ class OrganizationPermissionEvaluator(
...
+     * @param authentication [Authentication] describing an authenticated request
+     * @param organization is organization in which we want to change the status
+     * @param newStatus is new status in [organization]
+     * @return whether user described by [authentication] can have permission on change [organization] status on [newStatus]
+     */
+    fun hasPermissionToChangeStatus(authentication: Authentication?, organization: Organization, newStatus: OrganizationStatus): Boolean {
+        val oldStatus = organization.status
+
+        return when {
+            oldStatus == newStatus -> false
```
nulls, line 47: why false here?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -40,28 +40,73 @@ class OrganizationService(
...
+
+    /**
+     * Mark organization with [organization] as created.
+     * If an organization was previously banned, then all its projects become deleted.
+     *
+     * @param organization an [Organization] to create
+     * @param organization
+     * @return deleted organization
+     */
+    fun recoverOrganization(organization: Organization): Organization {
```
nulls, line 78: it changes two tables: need to mark as Transactional

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -40,28 +40,73 @@ class OrganizationService(
...
+     */
+    fun deleteOrganization(organization: Organization): Organization = if (!hasProjects(organization.name)) {
+        changeOrganizationStatus(organization, OrganizationStatus.DELETED)
+    } else {
+        organization
+    }
+
+    /**
+     * Mark organization with [organization] as created.
+     * If an organization was previously banned, then all its projects become deleted.
```
nulls, line 72: is it requirement or some limitation?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -40,28 +40,73 @@ class OrganizationService(
...
+        return changeOrganizationStatus(organization, OrganizationStatus.CREATED)
+    }
+
+    /**
+     * Mark organization with [organization] and all its projects as banned.
+     *
+     * @param organization an [Organization] to ban
+     * @return deleted organization
+     */
+    fun banOrganization(organization: Organization): Organization {
```
nulls, line 94: need to mark as Transactional

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -40,28 +40,73 @@ class OrganizationService(
...
+
     /**
      * @param organizationName the unique name of the organization.
      * @return `true` if this organization has at least one non-deleted project,
      *   `false` otherwise.
      */
     fun hasProjects(organizationName: String): Boolean =
             projectService.getAllByOrganizationName(organizationName).any { project ->
-                project.status != DELETED
+                project.status == CREATED
```
nulls, line 109: `hasProjects` doesn't count BANNED projects, it means that it's allowed to remove organization with banned projects -- is it requirement?

save-backend/src/test/kotlin/com/saveourtool/save/backend/security/ProjectPermissionEvaluatorTest.kt
```diff
@@ -75,18 +77,19 @@ class ProjectPermissionEvaluatorTest {
 
     @Test
     fun `permissions for project owners`() {
+        ownerPermissions.forEach { println(it.name) }
```
nulls, line 80: debug log

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
Arrgentum: in this case, the status of the organization and the status that we want to set are the same and therefore the action is not necessary

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Arrgentum: If the organization was ```banned``` by the administrator, then all its projects also immediately become ```banned```, and after it is ```restored```, we will not be able to restore all projects with the statuses that they had before. Therefore, such a solution is proposed so that the user can ```restore``` his own projects that he will need.

Arrgentum: of course, it is possible not to ban all its projects when the organization is banished, but to add more checks so that when the project is called, the organization is also astive, but are these additional checks needed at every step?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Arrgentum: yes

Arrgentum: Only active projects are taken into account here

save-backend/src/test/kotlin/com/saveourtool/save/backend/security/ProjectPermissionEvaluatorTest.kt
Arrgentum: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Arrgentum: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Arrgentum: Done

sanyavertolet (approved):
LGTM, take a look at other conversations

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -315,23 +315,24 @@ internal class OrganizationController(
...
-    @ApiResponse(responseCode = "200", description = "Successfully deleted an organization.")
-    @ApiResponse(responseCode = "403", description = "Not enough permission for deleting this organization.")
+    @ApiResponse(responseCode = "200", description = "Successfully change status an organization.")
+    @ApiResponse(responseCode = "403", description = "Not enough permission for this action on organization.")
     @ApiResponse(responseCode = "404", description = "Could not find an organization with such name.")
     @ApiResponse(responseCode = "409", description = "There are projects connected to organization. Please delete all of them and try again.")
-    fun deleteOrganization(
+    fun changeOrganizationStatus(
         @PathVariable organizationName: String,
+        @RequestParam status: OrganizationStatus,
```
nulls, line 335: need to add to Parameters in swagger's annotation

Arrgentum: If I understood correctly, then the ```OrganizationController```
has import ```io.swagger.v3.oas.annotations.Parameters```

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
nulls: got the problem, but maybe left them as banned instead of deleted?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
nulls: need to add here (line 326):
```kotlin
    @Parameters(
        Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of an organization", required = true),
    )
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Arrgentum: It's not always that all projects will be banned. 
At the same time, it makes no sense to unban an organization if all projects in it are banned. Here it is easier for the user to create a new one. Yes, and ```super_admin``` will know about it and if anything, it's easy to ban some project

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
Arrgentum: Done

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
nulls: so even user has permission to do it, user gets 401. it can mislead

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
nulls: > It's not always that all projects will be banned.

```kotlin
    @Transactional
    fun banOrganization(organization: Organization): Organization {
        projectService.getAllByOrganizationName(organization.name).forEach {
            it.status = BANNED
            projectService.updateProject(it)
        }
        return changeOrganizationStatus(organization, OrganizationStatus.BANNED)
    }
```

When we ban organization, we ban all projects.
I guess we can't unban project if organization is banned (btw, maybe add such validation?), so project can be banned or deleted for banned organization. Maybe left status of project as is when organization is unbanned?

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
Arrgentum: Well, this operation is useless. In principle, the button to restore the organization will not be displayed on the frontend if the organization is already ``CREATED``

Arrgentum: Or is it better for me to allow any action, even if a person wants to do so?

nulls: suggest to check oldStatus and newStatus in controller and return 404 to user

Arrgentum: Or is it better to check for statuses earlier? And in case of a match, issue a 404?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
nulls: discussed offline: owner of organization can restore deleted project, but can't unban project. And we don't store previous state before banning. So it's the easier solution: delete project after unbanning organization

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
nulls: > Or is it better to check for statuses earlier? And in case of a match, issue a 404?

yes

> Or is it better for me to allow any action, even if a person wants to do so?

nope, I think we need to check permission in all cases even old status == new status in this method.
probably we can throw exception here if we check changing permission to same value

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/OrganizationPermissionEvaluator.kt
Arrgentum: Done

Cheshiriks (approved):

PR #1440, opened 2022-11-11
Moved Project and Organization

### LINKED ISSUE #527: Move `Project` class into `jvmMain`, introduce `ProjectDto`

nulls:
It closes #527


save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/AgentsControllerTest.kt
```diff
@@ -68,7 +68,7 @@ class AgentsControllerTest {
             .accept(MediaType.APPLICATION_JSON)
             .bodyValue(
                 listOf(
-                    AgentStatusDto(LocalDateTime.now(), AgentState.IDLE, "container-1")
+                    AgentStatusDto(AgentState.IDLE, "container-1")
```
akuleshov7, line 71: Time is not needed on AgentStatus? Why?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Contest.kt
```diff
@@ -52,12 +54,12 @@ class Contest(
     fun toDto() = ContestDto(
         name,
         status,
-        startTime!!,
-        endTime!!,
+        startTime?.toKotlinLocalDateTime()!!,
```
akuleshov7, line 57: I think the previous version, when we just had a typealias with expect/actual implementation looked better...

akuleshov7 (approved):
Looks good, we will definitely get errors because of this change, but the logic get's better, I love it

akuleshov7:
Will need to stop other's development, until you merge it

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/AgentsControllerTest.kt
nulls: it's set as LocalDateTime.now() in constructor. We set only in two places another values and it's a test

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Contest.kt
nulls: yeap, it will be great to have alias in kotlinx.serialization instead of a dedicated class

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationAdminView.kt
```diff
@@ -158,14 +158,15 @@ internal class OrganizationAdminView : AbstractView<Props, OrganizationAdminStat
...
-    private suspend fun getOrganizations(): MutableList<Organization> {
+    private suspend fun getOrganizations(): MutableList<OrganizationDto> {
         val response = get(
             url = "$apiUrl/organizations/all?onlyActive=${true}",
             headers = jsonHeaders,
             loadingHandler = ::classLoadingHandler,
         )
 
         return when {
+            // FIXME
```
nulls, line 169: fixed

nulls:
> Will need to stop other's development, until you merge it

will merge after #1454 -- will wait to avoid merge conflicts

PR #1422, opened 2022-11-07
Adding save-demo to docker-compose

sanyavertolet:
### What's done:
* Added `save-demo` to `docker-compose`

PR #1442, opened 2022-11-11
Add `save-demo` path to `api-gateway`

sanyavertolet:
This PR fixes #1422 

### What's done:
* Fixed `application.yml` and `application-dev.yml` in order to support redirecting to `/demo/api/**`

### DIFF
--- a/api-gateway/src/main/resources/application-dev.yml
+++ b/api-gateway/src/main/resources/application-dev.yml
@@ -6,6 +6,8 @@ gateway:
     url: http://localhost:8080
   sandbox:
     url: http://localhost:5400
+  demo:
+    url: http://localhost:5401
 spring:
   security:
     oauth2:
@@ -58,6 +60,6 @@ spring:
             client-secret: 2c05929edd949d670c3454258bbdbfefb7ba3eb7
 logging:
   level:
-    org.springframework.web: DEBUG
-    org.springframework.security: DEBUG
+    org.springframework.web: TRACE
+    org.springframework.security: TRACE
     com.nimbusds: TRACE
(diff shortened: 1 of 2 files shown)


api-gateway/src/main/resources/application.yml
```diff
@@ -60,6 +62,14 @@ spring:
...
             - RemoveRequestHeader=Cookie
             - ConvertAuthorizationHeader=
+        - id: demo-api_route
+          uri: ${gateway.demo.url}
+          predicates:
+            - Path=/demo/api/**
+          filters:
+            # If SESSION cookie is passed to downstream, it is then removed, because downstream discards it
+            - RemoveRequestHeader=Cookie
+            - ConvertAuthorizationHeader=
```
petertrr, line 72: Do we need to pass authentication to demo? Or will it be useful in the future?

api-gateway/src/main/resources/application-dev.yml
```diff
@@ -58,6 +60,6 @@ spring:
             client-secret: 2c05929edd949d670c3454258bbdbfefb7ba3eb7
 logging:
   level:
-    org.springframework.web: DEBUG
-    org.springframework.security: DEBUG
+    org.springframework.web: TRACE
```
petertrr, line 63: This will be a lot of logging, maybe make it `DEBUG` again?

petertrr (approved):

api-gateway/src/main/resources/application.yml
sanyavertolet: I guess it might be useful as far as we might want to store some user files on save-demo

api-gateway/src/main/resources/application-dev.yml
sanyavertolet: Yes, sure, my bad

sanyavertolet: Done

api-gateway/src/main/resources/application.yml
nulls: big brother is watching you

PR #1445, opened 2022-11-11
ConfigProperties and Storage for demo

sanyavertolet:
### What's done:
 * Added `ToolKey`
 * Implemented `ToolStorage`
 * Updated `application.properties`
 * Added `application-mac.properties`
 * Added `save-demo` configuration

### DIFF
--- a/save-demo/src/main/resources/application-mac.properties
+++ b/save-demo/src/main/resources/application-mac.properties
@@ -1,0 +1,1 @@
+demo.fileStorage.location=/Users/Shared/.save-cloud/cnb/files
--- a/save-demo/src/main/resources/application.properties
+++ b/save-demo/src/main/resources/application.properties
@@ -1,3 +1,4 @@
 server.port=5401
 management.endpoints.web.exposure.include=health,info,prometheus
 spring.codec.max-in-memory-size=100MB
+demo.fileStorage.location=${user.home}/.save-cloud/cnb/files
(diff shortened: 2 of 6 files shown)


save-demo/src/main/resources/application-mac.properties
```diff
@@ -0,0 +1 @@
+demo.fileStorage.location=/Users/Shared/.save-cloud/cnb/files
```
petertrr, line 1: BTW, what's wrong with `user.home` on Mac?

petertrr (approved):

save-demo/src/main/resources/application-mac.properties
sanyavertolet: I do not want to move existing .save_cloud from /Users/Shared to user.home

save-demo/src/main/resources/application.properties
```diff
@@ -1,3 +1,4 @@
 server.port=5401
 management.endpoints.web.exposure.include=health,info,prometheus
 spring.codec.max-in-memory-size=100MB
+demo.fileStorage.location=${user.home}/.save-cloud/cnb/files
```
nulls, line 4: what's about prod configuration?

nulls: yeap, but I'm asking more about using `${user.home}` in prod

nulls: sandbox and backend use `/home/cnb/files` as file storage location. better to use the similar location

sanyavertolet: Will do it as a part of #1450

sanyavertolet: Done

PR #1450, opened 2022-11-14
Remove unused deps from save-demo

sanyavertolet:
### What's done:
 * Removed `diktat` from `save-demo` deps
 * Removed `ktlint` from `save-demo` deps
 * Removed ktlint version from libs.versions.toml
 * Fixed storage location configs for save-demo

### DIFF
--- a/save-demo/build.gradle.kts
+++ b/save-demo/build.gradle.kts
@@ -22,13 +22,4 @@ val diktatVersion: String = libs.versions.diktat.get()
 dependencies {
     implementation(projects.saveCloudCommon)
     implementation(libs.save.common.jvm)
-
-    implementation(libs.ktlint.core)
-    implementation(libs.ktlint.rulesets.standard)
-    implementation("org.cqfn.diktat:diktat-common:$diktatVersion") {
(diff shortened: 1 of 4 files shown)


save-demo/build.gradle.kts
```diff
@@ -22,13 +22,4 @@ val diktatVersion: String = libs.versions.diktat.get()
 dependencies {
     implementation(projects.saveCloudCommon)
     implementation(libs.save.common.jvm)
-
-    implementation(libs.ktlint.core)
-    implementation(libs.ktlint.rulesets.standard)
```
petertrr, line 27: Are these still needed in `libs.versions.toml`? Also, where do we store actual versions (of diktat-cli)?

sanyavertolet: I guess we do not need to store ktlint version. Will also remove it.

For now actual version of diktat-cli and ktlint-cli should be stored in resources dir, but it will be changed to storage in next PR

petertrr (approved):

nulls (approved):

PR #1457, opened 2022-11-17
Added demo-cpg

nulls:


api-gateway/src/main/resources/application.yml
```diff
@@ -13,6 +13,8 @@ gateway:
     url: http://sandbox:5400
   demo:
     url: http://demo:5401
+  demo-cpg:
+    url: http://demo-cpg:5402
```
nulls, line 17: old port

save-cloud-charts/save-cloud/values.yaml
```diff
@@ -54,6 +54,18 @@ gateway:
...
     # Any additional options for Service `spec` block
     spec: {}
+demo-cpg:
+  name: demo-cpg
+  profile: dev
+  imageName: save-demo-cpg
+  containerPort: 5500
+  managementPort: 5501
+  clusterIP: null
+  db:
```
nulls, line 64: they are not used

save-cloud-charts/cpg/templates/demo-cpg-deployment.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: demo-cpg
```
petertrr, line 4: I actually thought that it can be simply included into the same chart as other services. There shouldn't be any problems with accessing the database from a different namespace

petertrr (approved):

save-cloud-charts/cpg/templates/demo-cpg-deployment.yaml
nulls: I thought since we extracted neo4j to a dedicated chart, backend for neo4j also should be extracted

PR #1458, opened 2022-11-17
Create CPG view

Cheshiriks:
Create CPG view
prod175334

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/BuilderComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/BuilderComponent.kt
@@ -1,0 +1,94 @@
+/**
+ * Function component for build window
+ */
+
+@file:Suppress("FILE_NAME_MATCH_CLASS")
+
+package com.saveourtool.save.frontend.components.basic
+
+import com.saveourtool.save.frontend.components.basic.codeeditor.codeEditorComponent
+import com.saveourtool.save.frontend.externals.reactace.AceModes
+import com.saveourtool.save.frontend.externals.reactace.AceThemes
+import com.saveourtool.save.frontend.utils.*
+import csstype.ClassName
+import react.FC
+import react.Props
+import react.dom.html.ReactHTML.div
+import react.useState
+
+val builderComponent = builderComponent()
+
+/**
+ * DemoComponent [Props]
+ */
+@Suppress("TYPE_ALIAS")
+external interface BuilderComponentProps : Props {
+    /**
+     * Theme for Ace Editor
+     */
+    var selectedTheme: AceThemes
+
+    /**
+     * Lambda for run request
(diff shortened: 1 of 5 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/BuilderComponent.kt
```diff
@@ -0,0 +1,94 @@
...
+                    onDraftTextUpdate = { code ->
+                        setCodeLines(code)
+                    }
+                    isDisabled = false
+                }
+            }
+
+            div {
+                className = ClassName("col-8 card mt-4")
+                // TODO: need to added window for graph
```
sanyavertolet, line 68: You need to add props with type `(ChildrenBuilder, R) -> Unit` and use here 
```Kotlin
props.resultBuilder(this, result)
```
R should be generic (probably <R: DemoResult>)
It is needed in order to make this component reusable for any further demos

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/reactace/AceModes.kt
```diff
@@ -4,52 +4,59 @@ package com.saveourtool.save.frontend.externals.reactace
...
-     * C/C++ highlight mode
+     * C highlight mode
      */
-    CPP("c_cpp"),
+    C("c_cpp", "C++"),
+
+    /**
+     * C++ highlight mode
+     */
+    CPP("c_cpp", "C"),
```
sanyavertolet, line 19: C for CPP and vice versa

sanyavertolet (approved):
I would like you to make a general component in order to escape code duplications (guess it can be done as Phase2)

akuleshov7:
1. Big white blocks are not on the same level at the top

akuleshov7:
2. The Selection of theme for IDE is very ugly: items should be in lower case, put the selector of theme and language on the same line ABOVE the code editor block

akuleshov7:
3. Same about send request button - move it to a single line: this line is good enough 
image

akuleshov7:
4. Make a small margin from the right and left sides of blocks

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/reactace/AceModes.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/BuilderComponent.kt
Cheshiriks: Fixed

akuleshov7 (approved):

Arrgentum (approved):

PR #1472, opened 2022-11-21
Removed explicit usage of spring version

nulls:


gradle/libs.versions.toml
```diff
@@ -100,11 +99,10 @@ spring-boot-gradle-plugin = { module = "org.springframework.boot:spring-boot-gra
 spring-cloud-starter-gateway = { module = "org.springframework.cloud:spring-cloud-starter-gateway", version.ref = "spring-cloud" }
 spring-cloud-starter-kubernetes-client-config = { module = "org.springframework.cloud:spring-cloud-starter-kubernetes-client-config", version.ref = "spring-cloud-kubernetes" }
 spring-boot-starter-oauth2-client = { module = "org.springframework.boot:spring-boot-starter-oauth2-client" }
-spring-context-indexer = { module = "org.springframework:spring-context-indexer", version.ref = "spring" }
```
petertrr, line 103: That's weird, looks like I declared a dependency but didn't add it to any module's classpath...

petertrr (approved):

gradle/libs.versions.toml
nulls: do we need to add?

petertrr: It can be added as `kapt("org.springframework:spring-context-indexer")`, I think, in `spring-boot-configuration.gradle.kts`. I think we can experiment with it later as a separate PR

PR #1473, opened 2022-11-21
Exclusions for spring-boot 2.7.5

nulls:


renovate.json
```diff
@@ -19,7 +19,8 @@
         "*"
       ],
       "excludePackagePatterns": [
-        "^org\\.jetbrains\\.kotlin[.:]"
+        "^org\\.jetbrains\\.kotlin[.:]",
+        "^org\\.springframework\\.boot(:[a-z0-9-]+)+\\:2.7.5"
```
petertrr, line 23: I think it'll start creating a separate PR for this pattern, but it'll still be more convenient than constantly editing the file

petertrr (approved):

renovate.json
nulls: I don't think, it's only exclude packages from this `group name`, but I don't add it to another group name (as it was done for kotlin packages).

But I think we can extract spring dependencies to a separate task for renovate

PR #1476, opened 2022-11-21
Delete execution event

Cheshiriks:
Delete execution event

### DIFF
--- a/db/v-2/events/event_scheduler.xml
+++ b/db/v-2/events/event_scheduler.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="event-delete-execution-1" author="frolov" context="prod">
+        <sql>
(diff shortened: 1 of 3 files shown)


db/v-2/events/event_scheduler.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="event-delete-execution-1" author="frolov" context="prod">
```
nulls, line 8: what's about context `scheduler`?

Cheshiriks: If the context is non-existent, then the event is not created. But we used to have a Prod context, and after we changed it, the base probably didn’t drop, so the events weren’t deleted.

nulls: maybe then change `scheduler` to `prod` for existed scripts?

akuleshov7: how liquibase will work with such change of a context?

akuleshov7 (approved):

db/v-2/events/event_scheduler.xml
nulls: yeap, DATABASECHANGELOG contains the column `contexts` . So changing context will be a problem

nulls (approved):

PR #1478, opened 2022-11-21
Reset current page after filtering in history view

kgevorkyan:
### What's done:
* Reset current page after filtering in history view


nulls (approved):

petertrr:
Addressed this (probably) in #1488 by using `useNavigate` instead of changing `window.location`

PR #1488, opened 2022-11-22
Migrate from `kotlin-react-table` (v7) to `kotlin-tanstack-react-table` (v8)

### LINKED ISSUE #1243: Switch from `kotlin-react-table` to `kotlin-tanstack-react-table` (i.e. v7 -> v8)
As stated at https://github.com/JetBrains/kotlin-wrappers/tree/master/kotlin-react-table, this library is archived (i.e. fixed at react-table v7?) and should be replaced with `kotlin-tanstack-react-table`, which is a wrapper around react-table v8.
Migration guide for JS/TS: https://github.com/TanStack/table#migrating-from-react-table-v7

petertrr:
* Switched dependency from `kotlin-react-table` to `kotlin-tanstack-react-table`
* Added column builders similar to what used to be
* Added some utility getters
* Switched `tableComponent` to the new API; supported all previously used functionality: paging, sorting, expanding
* Removed calls to react-table's filtering, because we are using only our own filtering
* Removed reload after filtering on ExecutionView; use `useNavigate` instead
* Removed exclusion in renovate.json; bumped kotlin-wrappers version

Closes #1243


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
```diff
@@ -197,52 +224,46 @@ fun <D : Any, P : TableProps<D>> tableComponent(
...
-                                            this[key] = value
-                                        }
-                                    }))
+                                spread(getRowProps(row))
+                                row.getVisibleCells().map { cell ->
+                                    child(renderCell(cell))
                                 }
                             }
-                            if (row.isExpanded) {
+                            if (row.getIsExpanded()) {
```
nulls, line 266: maybe add extension in `TableUtil` to `get`?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/tables/TableComponent.kt
petertrr: Added getters for `isExpanded` and `can(Next/Previous)Page`

PR #1493, opened 2022-11-23
Remove maven repo for sarif4k

petertrr:
Actually release of save-core is required


settings.gradle.kts
```diff
@@ -8,18 +8,6 @@ dependencyResolutionManagement {
                 includeGroup("com.saveourtool.save")
             }
         }
-        maven {
```
nulls, line 11: need to upgrade version to 0.3.0 also

nulls:
can be closed in favor to https://github.com/saveourtool/save-cloud/pull/1542

petertrr:
The only thing that is missing in master rn is snapshot repo for save-cli, which was there to make switching to snapshot version easier. However, we tend to always use release these days, so I guess it's not that needed.

PR #1496, opened 2022-11-24
Routing for the neo4j

akuleshov7:


api-gateway/src/main/resources/application.yml
```diff
@@ -15,6 +15,9 @@ gateway:
     url: http://demo:5401
   demo-cpg:
     url: http://demo-cpg:5500
+  neo4j-browser:
+    url: http://save-cloud:7474
```
nulls, line 19: why save-cloud? does neo4j have such name?

api-gateway/src/main/resources/application.yml
```diff
@@ -79,6 +82,10 @@ spring:
           filters:
             # If SESSION cookie is passed to downstream, it is then removed, because downstream discards it
             - RemoveRequestHeader=Cookie
+        - id: neo4j_route
+          uri: ${gateway.neo4j-browser.url}
+           predicates:
+             - Path=/neo4j/browser/**
```
nulls, line 88: maybe add
```
filters:
            # If SESSION cookie is passed to downstream, it is then removed, because downstream discards it
            - RemoveRequestHeader=Cookie
```
?

nulls (approved):

api-gateway/src/main/resources/application.yml
petertrr: Weirdly enough, it does. From `neo4j-svc.yaml`:
```yaml
metadata:
  name: "{{ $.Release.Name }}"
```

api-gateway/src/main/resources/application.yml
petertrr, line 88:
```suggestion
          predicates:
            - Path=/neo4j/browser/**
```
Extra indentation?

api-gateway/src/main/resources/application.yml
nulls: can we override it? or will leave as is?

akuleshov7: how we can override their names? I think we need to report it

Issue #1425, opened 2022-11-09
Extract (functional) JVM-only code from `save-cloud-common` into a separate Gradle module


petertrr:
As we've faced in #1408, for example. It makes it trickier to add new services or to add new dependencies to a shared module.

One solution would be to split `save-cloud-common` and extract smaller specialized modules. `save-cloud-common` is getting too large anyway, and it was mostly planned as a common data classes module. These smaller modules could be `save-cloud-common-spring-data` and `save-cloud-common-spring-security`. Domain classes, if any (i.e. with JPA annotations) can be left there though.

This also slows down builds, because neither frontend nor agent require JVM-only part of `save-cloud-common`, however their build doesn't start until `save-cloud-common` is fully built

PR #1497, opened 2022-11-24
Remove most of JPA related dependencies from auth service

petertrr:
* Remove `spring-data-configuration` from applied plugins it's not needed mostly; keep only depednecy needed to resolve one class
* Remove excludes on authoconfiguration classes that are no longer present

`auth-service` is a library and not a service so all Srping Boot starters are a bit redundant there
Related: #1408, #1425


authentication-service/build.gradle.kts
```diff
@@ -25,6 +24,7 @@ dependencies {
     implementation(projects.saveCloudCommon)
     implementation(libs.spring.boot.starter.security)
     implementation(libs.spring.security.core)
+    implementation("org.springframework:spring-jdbc")
```
kgevorkyan, line 27: Did you tried to start some of applications, that used `auth-service` as dependency?
I remember, that I tried to use such solution sometime ago, but got errors.
Does it all okay now?

kgevorkyan: Build was successful in that case, but nor the `backend` not `sandbox` didn't started

petertrr: Double checked, but I think for backend and sandbox it's safe - they have all spring-data related dependencies anyway. Also checked for gateway, which also has this dependency, and it's starting fine too.

petertrr:
* Remove `spring-data-configuration` from applied plugins it's not needed mostly; keep only depednecy needed to resolve one class
* Remove excludes on authoconfiguration classes that are no longer present

PR #1523, opened 2022-11-28
POC for a mobile WeclomeView

akuleshov7:


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/MobileRouting.kt
```diff
@@ -0,0 +1,31 @@
...
+import react.router.Route
+import react.router.Routes
+
+/**
+ * Just put a map: View -> Route URL to this list
+ */
+val mobileRoutes = VFC {
+    Routes {
+        listOf(
+            AboutUsMobileView::class.react.create() to FrontendRoutes.ABOUT_US.path,
```
akuleshov7, line 22: @petertrr @nulls @sanyavertolet  if you will have any ideas of how to cast it, so I won't need to have `.react.create()`, will be great. Simple map does not work, due to receiver type...

```
.map { it.react.create() }
```

nulls: migrate to `FC`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/welcome/MarketingTitles.kt
```diff
@@ -0,0 +1,87 @@
...
+
+import react.ChildrenBuilder
+import react.dom.html.ReactHTML
+
+/**
+ * @param textColor
+ * @param isDark
+ */
+fun ChildrenBuilder.welcomeMarketingTitle(textColor: String, isDark: Boolean = false) {
+    ReactHTML.div {
```
sanyavertolet, line 20:
```suggestion
    div {
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/welcome/MarketingTitles.kt
```diff
@@ -0,0 +1,87 @@
...
+        className = ClassName("col-lg-4 ml-auto mt-3 mb-5 mr-5 ml-0 $textColor")
+        marketingTitle("Software", isDark)
+        marketingTitle("Analysis", isDark)
+        marketingTitle("Verification &", isDark)
+        marketingTitle("Evaluation", isDark)
+        ReactHTML.h3 {
```
sanyavertolet, line 26:
```suggestion
        h3 {
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/welcome/pagers/WelcomePager.kt
```diff
@@ -4,14 +4,15 @@ import com.saveourtool.save.frontend.externals.animations.Animation
 import react.ChildrenBuilder
 
 val allWelcomePagers = listOf(
-    listOf(HighLevelSave),
+    // listOf(HighLevelSave),
```
sanyavertolet, line 7: Should it be commented?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/MobileRouting.kt
sanyavertolet: Maybe wrap it into `VFC`s? In that way it seems possible to easily iterate over the list and create components

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/MobileRouting.kt
sanyavertolet: > migrate to `FC`

It seems to be a long journey, can we do it now?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/welcome/pagers/WelcomePager.kt
akuleshov7: yeah, will think of how to draw it - in the next PR

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/MobileRouting.kt
akuleshov7: > Maybe wrap it into `VFC`s? In that way it seems possible to easily iterate over the list and create components

wrapper looks much more ugly, I want to remove existing actually

petertrr: > Simple map does not work, due to receiver type...

What type is inferred in this case? `.react` is an extension on `KClass<out Component<P, *>>`, maybe list can be cast somehow?

Issue #659, opened 2022-04-02
Support the algorithm for regression and flaky tests detection


akuleshov7:
Save should be able to detect regressions in tests. This information should be visualized on the history view.
Save also should be able to detect flaky tests, this should be shown on the run page of a particular execution.

Probably, the visualisation can be done in a separate issue, it depends on the author.

 * [X] Core #1525 
    * Test run statistics can be collected, using a sliding window of a configurable size.
    * Unique test ids are generated using SHA-2.
    * Scalar test metrics can now be collected:
       ![regular-test](https://user-images.githubusercontent.com/73111822/213238253-cfbc5a42-c49e-4070-8307-649815c647e3.png)
       * successful & failed run count,
       * flip rate,
       * failure rate,
       * average & median test duration.
    * Flaky test detection based on a flip rate (30% threshold).
       ![flaky-test](https://user-images.githubusercontent.com/73111822/213238210-d83f0c3a-c9ac-4f0a-a8fa-6fcde0b0d3ef.png)
    * Regression detection.
      ![regression](https://user-images.githubusercontent.com/73111822/213238245-5eb864b5-9983-475f-804a-6b8d1e243a79.png)
    * Permanent failure detection.
      ![permanent-failure](https://user-images.githubusercontent.com/73111822/213238235-fddd78ff-65ce-4dc0-b127-9f6cd26f3a42.png)
    * The list of individual test runs within the given batch exposed via the REST API.
    * [X] Data types are serializable #1678
 * [X] [Test data generator](https://github.com/0x6675636b796f75676974687562/save-flaky-tests)
 * [x] Back-end
    * [X] #1707
    * [x] #1724
    * [x] #1728
    * [x] #1732
    * [x] #1734
    ![jmx](https://user-images.githubusercontent.com/73111822/213238224-a0b919fa-338e-4daf-aa88-d6482fa943a5.png)
 * [x] Front-end
    * [X] #1712
    * [x] #1735

### Known issues

* #1726

PR #1525, opened 2022-11-29
Add the core test analysis functionality

0x6675636b796f75676974687562:
# What's done:

 * Test run statistics can be collected, using a sliding window of
   a configurable size.
 * Unique test ids are generated using SHA-2.
 * Scalar test metrics can now be collected:
    * successful & failed run count,
    * flip rate,
    * failure rate,
    * average & median test duration.
 * Flaky test detection based on a flip rate (30% threshold).
 * Regression detection.
 * The list of individual test runs within the given batch exposed via the REST API.
 * See #659.

### DIFF
--- a/save-cloud-charts/save-cloud/.gitignore
+++ b/save-cloud-charts/save-cloud/.gitignore
@@ -1,0 +1,1 @@
+/save-cloud-*.tgz
--- a/test-analysis-core/README.adoc
+++ b/test-analysis-core/README.adoc
@@ -1,0 +1,26 @@
+= Test Analysis
+:toc:
+
+== Features
+
+* Collection of scalar xref:#metrics[test metrics].
+* _Sliding window_ of a configurable size.
+* xref:#algorithms[Test analysis].
+
+[#metrics]
+=== Test Metrics
+
+* The number of successful runs;
+* The number of failures;
+* The number of status "flips" (i.e. "success -> failure" and "failure -> success"
+transitions across consecutive test runs);
+* Failure rate;
+* Flip rate;
+* Average test duration;
+* Median test duration.
+
+[#algorithms]
+=== Algorithms
+
+* _Flaky test_ detection based on _flip rate_ (the default threshold is *30%*).
+* _Regression_ detection.
--- a/test-analysis-core/build.gradle.kts
+++ b/test-analysis-core/build.gradle.kts
@@ -1,11 +1,9 @@
-import com.saveourtool.save.buildutils.configureJacoco
-import com.saveourtool.save.buildutils.configurePublishing
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
 
 plugins {
     id("com.saveourtool.save.buildutils.kotlin-jvm-configuration")
     id("com.saveourtool.save.buildutils.code-quality-convention")
-    `maven-publish`
+    id("com.saveourtool.save.buildutils.publishing-configuration")
 }
 
 java {
@@ -53,6 +51,3 @@ publishing {
         }
     }
 }
-
-configureJacoco()
-configurePublishing()
--- a/test-analysis-core/build.gradle.kts
+++ b/test-analysis-core/build.gradle.kts
@@ -1,0 +1,58 @@
+import com.saveourtool.save.buildutils.configureJacoco
+import com.saveourtool.save.buildutils.configurePublishing
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+
+plugins {
+    id("com.saveourtool.save.buildutils.kotlin-jvm-configuration")
+    id("com.saveourtool.save.buildutils.code-quality-convention")
+    `maven-publish`
+}
+
+java {
+    withSourcesJar()
+}
+
+dependencies {
+    /*
+     * Necessary, because otherwise Gradle will try to resolve "org.slf4j:slf4j-api:."
+     * (the transitive dependency of "projects.saveCloudCommon")
+     * and fail to find the undefined version.
+     */
+    implementation(project.dependencies.platform(libs.spring.boot.dependencies))
+    api(projects.saveCloudCommon)
+
+    testApi(libs.assertj.core)
+    testApi(libs.mockito.kotlin)
+    testApi(libs.mockito.junit.jupiter)
+    testApi(libs.junit.jupiter.api)
+    testApi(libs.junit.jupiter.params)
+    testRuntimeOnly(libs.junit.jupiter.engine)
+}
+
+tasks.withType<Test> {
+    useJUnitPlatform()
+
+    testLogging {
+        showStandardStreams = true
+        showCauses = true
+        showExceptions = true
+        showStackTraces = true
+        exceptionFormat = FULL
+        events("passed", "skipped")
+    }
+
+    filter {
+        includeTestsMatching("com.saveourtool.save.test.analysis.*")
+    }
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            from(components["java"])
+        }
+    }
+}
+
+configureJacoco()
+configurePublishing()
--- a/test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
+++ b/test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
@@ -8,10 +8,30 @@ import com.saveourtool.save.test.analysis.api.results.IrregularTest
 /**
  * _Flip-rate_ based flaky test detection algorithm.
  *
+ * A test _flip_ is a status change (either from _successful_ to a _failed_ or
+ * vice versa) over two consecutive test runs.
+ *
+ * A _flip rate_ is the ratio of the actual flip number to the maximum possible
+ * flip number (i.e. run count minus one, see
+ * [RegularTestMetrics.flipRate] and [RegularTestMetrics.flipRatePercentage]):
+ *
+ * ```
+ *         N          + N
+ *          pass➔fail    fail➔pass
+ * R     = ────────────────────────
+ *  flip          N    - 1
+ *                 run
+ * ```
+ *
+ * If the _flip rate_ exceeds a certain [threshold][flipRateThreshold], the test
+ * is considered _flaky_.
+ *
  * @param minimumRunCount the minimum run count a sample should have to be
  *   representative.
  * @param flipRateThreshold the _flip rate_ threshold.
  *   If the threshold is exceeded, the test is considered _flaky_.
+ * @see RegularTestMetrics.flipRate
+ * @see RegularTestMetrics.flipRatePercentage
  */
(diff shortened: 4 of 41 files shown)


save-cloud-charts/save-cloud/.gitignore
```diff
@@ -0,0 +1 @@
+/save-cloud-*.tgz
```
akuleshov7, line 1: I am making a weekly report now. Is this all the work you have done in this week? :D

0x6675636b796f75676974687562: Why are you reviewing draft PRs?

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/metrics/TestMetrics.kt
```diff
@@ -0,0 +1,10 @@
+package com.saveourtool.save.test.analysis.api.metrics
+
+import com.saveourtool.save.test.analysis.api.TestStatisticsStorage
+
+/**
+ * Scalar test metrics returned by [TestStatisticsStorage.getTestMetrics].
+ *
+ * @see TestStatisticsStorage.getTestMetrics
+ */
+sealed interface TestMetrics
```
petertrr, line 10: Should these classes (`TestMetrics` and `AnalysisResults` and their descendants) be located in `save-cloud-common` from the very beginning since we'll probably be storing this information in the database in the future?

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/TestId.kt
```diff
@@ -0,0 +1,13 @@
...
+ * A test executed within a particular project (the same test executed in a
+ * different project even within the same organization will have a different
+ * [hash]).
+ *
+ * To create a [TestId] instance, use [TestIdGenerator.testId].
+ *
+ * @property hash the SHA hash that uniquely identifies this test.
+ */
+@JvmInline
+value class TestId internal constructor(private val hash: String)
```
petertrr, line 13: Maybe call it `TestExecutionId` in order not to confuse with `Test.id`, because it has the same meaning as `TestExecution.id`? Also, why can't we use `TestExecution.id` as a key here?

petertrr (approved):

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/TestId.kt
0x6675636b796f75676974687562: No, this is not a `TestExecutionId`.
This is actually similar to `com.saveourtool.save.entities.Test` (which has a `hash`), but keep reading.

The same `save-cli` test (Git source with a specific version, test suite, TOML file) executed multiple times (in different executions) will have the same `TestId`.

In other words, `TestId` identifies a series of test runs (so that they can be statistically analyzed) rather than a single test. 

Moreover, as you may have noticed, the same test executed for a different project (e.g.: _KtLint_ vs _Diktat_, or _Diktat 1.2.3_ vs _Diktat 1.2.4_) or within a different organization will have a different `TestId`.

For example, on the production database, 

```sql
select count(distinct(test_id)) from test_execution;
```

yields **28**, despite the same [standard test suite](https://github.com/saveourtool/save-cli/tree/main/examples/kotlin-diktat) (mirrored [here](https://github.com/0x6675636b796f75676974687562/save-kotlin-diktat)) has about 22 tests which are executed for **multiple** SAVE-Cloud projects (Diktat, Diktat-Integration, etc.).

So if `Test.hash` were the entity I needed, I would expect a `test_id` count which is equal to or slightly exceeds a **multiple** of 22, e.g. 44+ (at least), 66+, 88+ etc.

In terms of CI and unit testing, my intention was to create something like a JUnit/TestNG _test name_ (e. g.:  [`org.apache.commons.lang3.AnnotationUtilsTest.testEquivalence`](https://gitbox.apache.org/repos/asf?p=commons-lang.git;a=blob_plain;f=src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java;hb=HEAD)), partitioned by organization name and project name.

petertrr: Awesome explanation, thanks

0x6675636b796f75676974687562: It seems actually possible to use a triplet of `(Test.hash, OrganizationName, ProjectName)`, but this is merely a matter of implementing a yet another `com.saveourtool.save.test.analysis.api.TestIdGenerator`.

Speaking of naming, the entity is actually very similar to `Test.hash` and this may well be retain the `TestId` name. If you don't like that too much, how about renaming it to `TestNameId`?

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/metrics/TestMetrics.kt
0x6675636b796f75676974687562: Both `TestMetrics` and `AnalysisResults` are actually denormalized data.

So my initial idea was to read the test run history in the background as the back-end service starts, and then keep continously updating the data as new tests are run.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/TestId.kt
petertrr: No, I think `TestId` is fine in a sense that it's a unique ID of a test and `TestNameId` would also be confusing with `Test.name`. Maybe `TestExecutionSeriesId` would be more appropriate, but we can keep it simpler.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/metrics/TestMetrics.kt
petertrr: Yes, I see that it's denormalized, but are you suggesting to keep statistics in memory only? Even if these algorithms require the whole series of execution to calculate or update, wouldn't it slow the things too much when statistics are only being read?

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
```diff
@@ -0,0 +1,42 @@
+package com.saveourtool.save.test.analysis.algorithms
+
+import com.saveourtool.save.test.analysis.api.TestRuns
+import com.saveourtool.save.test.analysis.api.metrics.RegularTestMetrics
+import com.saveourtool.save.test.analysis.api.results.FlakyTest
+import com.saveourtool.save.test.analysis.api.results.IrregularTest
+
+/**
+ * _Flip-rate_ based flaky test detection algorithm.
```
nulls, line 9: can you please add more information about algorithm here?

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun invoke(runs: TestRuns, metrics: RegularTestMetrics): IrregularTest? {
+        require(runs.size == metrics.runCount) {
+            "${runs.size} != ${metrics.runCount}"
+        }
+
+        return when {
+            metrics.flipRate > flipRateThreshold && metrics.runCount >= minimumRunCount ->
+                FlakyTest("Flip rate of ${metrics.flipRatePercentage}% over ${metrics.runCount} run(s)")
+
+            else -> null
```
kgevorkyan, line 39: Should we return `null` here? Or since it is **not** `IrregularTest`, we can return `RegularTest`? Thus, invoke will return `AnalysisResult` without nullability

test-analysis-core/README.adoc
```diff
@@ -0,0 +1,26 @@
+= Test Analysis
```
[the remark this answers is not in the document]
0x6675636b796f75676974687562: Plenty of reasons:

1. It's more powerful than Markdown, yet fully supported by IDEA (via a plug-in) and mostly by GitHub
2. You get a TOC for free
3. Internal cross-references are possible
4. History is preserved when editing large tables (remember what a PITA editing `diktat/available-rules.md` is)
5. Easily exported into PDF ([example](https://saveourtool.github.io/benedikt/ebook.pdf))
6. Exportable to GH-pages (via a separate action).

So, from now on, unless I have a really good reason to create a Markdown document, I tend to write AsciiDoctor instead.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
0x6675636b796f75676974687562: Already implemented in `DefaultTestAnalysisService`.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/metrics/TestMetrics.kt
0x6675636b796f75676974687562: > ... but are you suggesting to keep statistics in memory only

Well, _TeamCity_ does exactly that, keeping the last N runs of hundreds of thousands of tests in-memory.

The idea is to keep only a sub-volume of the whole statistical data available (the _sliding window_ discards older test runs), but maintain fast access to this data (when multiple clients are requesting analysis for tests X, Y, Z and so on).

There's the only limitation: your data structures should be as small as possible to avoid OOM.

If the volume of the underlying test run history is not that large, and the number of concurrent users is low, indeed we can read directly from the database, but at this (early prototype) stage the final decision is actually not that important.

We can implement initial support from the front-end and the back-end now, play for a while, and add extra persistent entities later.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/algorithms/FlipRateAnalysis.kt
0x6675636b796f75676974687562: Addressed with 7607e24e9dd1034bc245dc517f53154d79d8aa5e.

test-analysis-core/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
+import com.saveourtool.save.buildutils.configureJacoco
+import com.saveourtool.save.buildutils.configurePublishing
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+
+plugins {
+    id("com.saveourtool.save.buildutils.kotlin-jvm-configuration")
+    id("com.saveourtool.save.buildutils.code-quality-convention")
+    `maven-publish`
```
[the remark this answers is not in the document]
0x6675636b796f75676974687562: Addressed with 13ed6feb0d429f05d57391ad55dd46cde1a518a8.

PR #1539, opened 2022-11-30
[CPG] exception handler on connection to database

nulls:
### What's done:
* refactoring creation of connection to database: now we throw exception in case of problems
* introduced logback captured which registers additional appender which captures logs to list and returns as result
* DB exception returns as exception, exception in CPG returns as 200 with logs message

**NOTE**: `LogbackCaptures` is not thread safe -- it means that multi requests can erase logs for each callers


sanyavertolet (approved):

Cheshiriks (approved):

save-demo-cpg/src/main/kotlin/com/saveourtool/save/demo/cpg/controller/CpgController.kt
```diff
@@ -139,59 +138,32 @@ class CpgController(
...
             "Count base nodes to save [components: ${result.components.size}, additionalNode: ${result.additionalNodes.size}]"
         }
 
-        session.beginTransaction().use {
-            // FixMe: for each user we should keep the data
-            session.purgeDatabase()
+        sessionWithFactory.use { session ->
+            session.beginTransaction().use {
+                // FixMe: for each user we should keep the data
+                session.purgeDatabase()
```
akuleshov7, line 151: pls delete it also

akuleshov7 (approved):

save-demo-cpg/src/main/kotlin/com/saveourtool/save/demo/cpg/controller/CpgController.kt
nulls: got you

PR #1541, opened 2022-12-01
[Helm] Create separate PVCs for backend's and sandbox's storage

petertrr:
- Create separate PVCs for backend's and sandbox's storage
- Add missing property to backend's ConfigMap

### DIFF
--- a/save-cloud-charts/save-cloud/templates/fs-storage-pvc.yaml
+++ b/save-cloud-charts/save-cloud/templates/fs-storage-pvc.yaml
@@ -1,21 +1,24 @@
+{{- range tuple .Values.backend.name .Values.sandbox.name }}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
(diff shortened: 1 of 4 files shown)


save-cloud-charts/save-cloud/templates/fs-storage-pvc.yaml
```diff
@@ -1,21 +1,24 @@
+{{- range tuple .Values.backend.name .Values.sandbox.name }}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    io.kompose.service: save-fs-storage
-  {{- with .Values.storage.annotations  }}
+    io.kompose.service: {{ . }}-fs-storage
+  {{- with $.Values.storage.annotations  }}
```
petertrr, line 7: Default context (`.`) becomes overridden by `range`, so `$` can be used to access root context

nulls (approved):

PR #1543, opened 2022-12-01
Remove diktat-cli and change demo port

sanyavertolet:
### What's done:
 * Removed `diktat` from `demo` resources
 * Removed `diktat.cmd` from `demo` resources
 * Removed `ktlint-cli` from `demo` resources
 * Removed `ktlint-cli.cmd` from `demo` resources
 * Changed `save-demo` port from `5401` to `5421`

### Todo:
 * Support kubernetes deployment


api-gateway/src/main/resources/application-dev.yml
```diff
@@ -7,7 +7,7 @@ gateway:
   sandbox:
     url: http://localhost:5400
   demo:
-    url: http://localhost:5401
+    url: http://localhost:5421
```
nulls, line 10: can it be just 5420?
because we have additional port: `managementPort` which is `port` + 1 usually

nulls (approved):

api-gateway/src/main/resources/application-dev.yml
sanyavertolet: I would not like to allow ports that ends with `0` as far as they seem to be more popular.
`managementPort` will be `5422`, why not? Is it occupied?

petertrr (changes requested):
Please also update port in demo-deployment.yaml for Kubernetes

sanyavertolet:
> Please also update port in demo-deployment.yaml for Kubernetes

Is there already one? I planned to add those kubernetes yamls in the next PR

petertrr (approved):
> > Please also update port in demo-deployment.yaml for Kubernetes

> Is there already one? I planned to add those kubernetes yamls in the next PR 

My bad

api-gateway/src/main/resources/application-dev.yml
nulls: common pattern for all application: port ends on `0`, `managementPort` ends on `1`.
Now we have exception for demo project, which is not a temporary application in our ecosystem

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
```diff
@@ -18,9 +19,15 @@ class ToolStorage(
...
     @Suppress("DestructuringDeclarationWithTooManyEntries")
     override fun buildKey(rootDir: Path, pathToContent: Path): ToolKey {
-        val (_, ownerName, toolName, vcsTagName, executableName) = pathToContent.pathNamesTill(rootDir)
+        val (ownerName, toolName, vcsTagName, executableName) = pathToContent.pathNamesTill(rootDir)
         return ToolKey(ownerName, toolName, vcsTagName, executableName)
     }
 
     override fun buildPathToContent(rootDir: Path, key: ToolKey): Path = rootDir / key.ownerName / key.toolName / key.vcsTagName / key.executableName
+
+    override fun isKey(rootDir: Path, pathToContent: Path): Boolean = pathToContent.name != DS_STORE
```
nulls, line 28: suggestion to path `pathPartsCount` in constructor instead of overriding `isKey` for mac specific use case

api-gateway/src/main/resources/application-dev.yml
sanyavertolet: Suggestion: use 5-10 ports per service (one is container port, one is management port and others are reserved) going one by one: let's say from `12120` or some other port that is not that nice as `5000`

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
sanyavertolet: Oh sure

sanyavertolet: And what if `.DS_Store` is in the directory with the tool?

api-gateway/src/main/resources/application-dev.yml
nulls: looks ok for me
@petertrr @akuleshov7

petertrr: I think it doesn't matter that much, especially in Kubernetes, where ports just need to match in `application.properties` and `Deployment`, and then the port can be referenced by name and Service provides access on port 80.
We can even keep existing ports for development and then just override `server.port` in ConfigMap with `{{ .service.containerPort }}` to make change smaller

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
```diff
@@ -18,9 +19,15 @@ class ToolStorage(
...
+
+    companion object {
+        private const val DS_STORE = ".DS_Store"
```
0x6675636b796f75676974687562, line 31: I suggest this is converted to a list, because there may be other ignore patterns like `*~` (backup files created by text editors) or `Thumbs.db` (Windows Explorer metadata) and renamed to smth like `IGNORE_PATTERNS`:

```suggestion
        private val IGNORE_PATTERNS = listOf(".DS_Store")
```

0x6675636b796f75676974687562 (approved):

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
sanyavertolet: Fixed it in a bit different way - improved AbstractFileBasedStorage to have `storageIgnore` (by analogy with `gitignore`) - list of filenames that should not be considered as files from storage

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
sanyavertolet: Fixed it

api-gateway/src/main/resources/application-dev.yml
nulls: > We can even keep existing ports for development and then just override `server.port` in ConfigMap with `{{ .service.containerPort }}` to make change smaller

we can, but need to add ports for dev profile to be able run locally

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
nulls: got you, discussed offline --- agreed to add this fix in `AbstractFileBasedStorage`

api-gateway/src/main/resources/application-dev.yml
sanyavertolet: Anyway it seems to be a good practice for local deploy. It is better to reserve `n` ports going one by one then to reserve randomly-chosen ports

PR #1548, opened 2022-12-02
[DEMO] kubernetes and demo friends forever

sanyavertolet:
### What's done:
 * Added config files for `kubernetes` + `save_demo`


save-cloud-charts/save-cloud/templates/demo-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name: demo
+  labels:
+    {{- include "common.labels" (merge (dict "service" .Values.demo) .) | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      io.kompose.service: demo
+  strategy:
+    # Because of shared volume with multi-attach problem
+    type: Recreate
```
petertrr, line 13: `demo` doesn't have persistent volume mounts that can cause problems, like for backend. So specifying strategy explicitly is redundant (default value is rolling update)

save-cloud-charts/save-cloud/values.yaml
```diff
@@ -202,3 +202,10 @@ demo_cpg:
   containerPort: 5500
   managementPort: 5501
   clusterIP: null
+demo:
+  name: demo
+  profile: dev
```
nulls, line 207: why do we enable `dev` profile on k8s?

save-demo/src/main/resources/bootstrap.yml
```diff
@@ -0,0 +1,29 @@
+spring:
+  cloud:
+    kubernetes:
+      enabled: false
```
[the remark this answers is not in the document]
sanyavertolet: No, we do not, will remove it

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
petertrr: Yes, if it's intended to be persisted between redeployments

sanyavertolet: oh god, will do it

save-cloud-charts/save-cloud/templates/demo-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.demo.name }}-config
+data:
+  application.properties: |
+    management.endpoints.web.exposure.include=*
```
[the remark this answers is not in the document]
petertrr: Yes, we can introduce `application-kubernetes.properties` everywhere. I'm not sure, however, whether it'd be a good idea - now we have kubernetes-specific properties in a form of kubernetes-specific resource, which is ConfigMap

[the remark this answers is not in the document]
sanyavertolet: > now we have kubernetes-specific properties in a form of kubernetes-specific resource, which is ConfigMap
Which seems to be a good idea - to keep all the configuration at **one** (if only in one) place

save-cloud-charts/save-cloud/values.yaml
petertrr: Just a default value, for Minikube for example. It's overridden for actual prod

nulls: oh.. it's overridden by `values-cce.yaml`. Got it

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
petertrr: You also need to link pod template from Deployment and PV, see `volumes` and `volumeMounts` in backend-deployment

save-demo/src/main/resources/bootstrap.yml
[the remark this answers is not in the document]
petertrr: Not quite what I've meant - the whole `bootstrap.[yml|properties]` is only read by spring-cloud or spring-cloud-kubernetes. So rn this file will be ignored completely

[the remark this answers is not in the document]
sanyavertolet: Oh ok, will remove the file

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
sanyavertolet: Seems that it is done, but not sure, please, double-check

save-cloud-charts/save-cloud/templates/demo-configmap.yaml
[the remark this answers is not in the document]
petertrr: `ConfigMap` acts as an additional property source on top of common profile-agnostic properties (with `-Dspring.config.additional-location=optional:file:/home/cnb/config/application.properties` and `spring-boot.config-volume-mount`). We only have a single `application-kubernetes.properties` file in the project in save-orchestrator-common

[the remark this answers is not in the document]
sanyavertolet: Removed logging configuration so that config map now contains only `management.*`

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
sanyavertolet: Yep, saw that, the problem is with the order of lines

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+      restartPolicy: Always
+      containers:
+        - name: demo
+          {{- include "spring-boot.common" (merge (dict "service" .Values.demo) .) | nindent 10 }}
+          env:
+            {{- include "spring-boot.common.env" (merge (dict "service" .Values.demo) .) | nindent 12 }}
+            - name: JAVA_TOOL_OPTIONS
+              value: -XX:ReservedCodeCacheSize=48M
+          volumeMounts:
+          {{- include "spring-boot.management" .Values.demo | nindent 10 }}
```
petertrr, line 30: This thing should be on the same level as `volumeMounts` (and not part of it). You can see how it's expanded in `_helpers.tpl`

sanyavertolet: Fixed it

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
sanyavertolet: Not it seems it is done lol
Problems of copy-paste

petertrr (approved):
LGTM

PR #1560, opened 2022-12-05
Cacheable avatars

nulls:
### What's done:
* checking that avatar does exist for returning it
* added controller for avatars


save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -79,4 +79,26 @@ data class AvatarKey(
         AvatarType.ORGANIZATION -> "/$objectName/$imageName"
         AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$objectName/$imageName"
     }
+
+    companion object {
+        /**
+         * @param relativePath [AvatarKey.getRelativePath]
+         * @return [AvatarKey] created from [relativePath]
+         */
+        operator fun invoke(relativePath: String): AvatarKey = relativePath.split("/")
```
nulls, line 88: now it's not needed

Cheshiriks (approved):

PR #1563, opened 2022-12-05
Filtering when avatar doesn't exist

nulls:


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
```diff
@@ -86,7 +87,8 @@ internal class AvatarController(
             "Not found avatar for organization $organizationName with name $imageName"
         }
 
-    private fun AvatarKey.toMonoResponse(): Mono<ByteBufferFluxResponse> = avatarStorage.doesExist(this)
+    private fun AvatarKey.toMonoResponse(): Mono<ByteBufferFluxResponse> = this.toMono()
+        .filterWhen(avatarStorage::doesExist)
```
petertrr, line 91: Do we need an explicit `emptyToNotFound` here?

petertrr (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
nulls: fixed

PR #1564, opened 2022-12-05
logging.level.org.springframework=DEBUG

nulls:
backend, orchestrator, demo and demo cpg


sanyavertolet (approved):
Isn't it too much?

nulls:
> Isn't it too much?

it's only for spring.
will remove after stabilization

PR #1565, opened 2022-12-05
Initial 0.1 rework of App.kt

akuleshov7:
### What's done:
- Possibly breaking, risky change
- Moving App.kt routs to a separate file (initial version)

Will be done in 3 phazes


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/BasicRouting.kt
```diff
@@ -0,0 +1,242 @@
...
+import react.router.Route
+import react.router.Routes
+
+val testExecutionDetailsView = testExecutionDetailsView()
+
+/**
+ * Just put a map: View -> Route URL to this list
+ */
+val basicRouting: FC<AppProps> = FC { props ->
+    val contestView: FC<Props> = withRouter { location, params ->
```
petertrr, line 43: I think all these `FC`s need to be declared on top level, not inside another `FC`

akuleshov7: That’s not possible when you have `props` and need to pass them…

PR #1582, opened 2022-12-07
[DEMO] Major diktat-demo improvements

sanyavertolet:
### What's done:
 * Fixed an issue with crashes of diktat fix - set env's `JAVA_TOOL_OPTIONS` and `_JAVA_OPTIONS` to empty string
 * Added `Demo` dropdown to `top bar` at frontend
 * Improved diktat temp configuration - now there always is default `diktat-analysis.yaml` and `src` is in path to working dir
 * Added `Constants.kt` file with demo's constants

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
@@ -18,7 +18,6 @@ data class DemoAdditionalParams(
     val language: Languages = Languages.KOTLIN,
 ) : DemoAdditionalParams {
     companion object {
-        @Language("YAML")
         val defaultDiktatConfig = """
             |- name: DIKTAT_COMMON
             |  enabled: true
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
@@ -18,6 +18,7 @@ data class DemoAdditionalParams(
     val language: Languages = Languages.KOTLIN,
 ) : DemoAdditionalParams {
     companion object {
+        @Language("YAML")
         val defaultDiktatConfig = """
             |- name: DIKTAT_COMMON
             |  enabled: true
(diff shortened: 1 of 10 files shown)


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
```diff
@@ -14,6 +14,109 @@ import kotlinx.serialization.Serializable
...
     val mode: DiktatDemoMode = DiktatDemoMode.WARN,
     val tool: DiktatDemoTool = DiktatDemoTool.DIKTAT,
-    val config: String? = null,
-    val language: Languages = Languages.KOTLIN
-) : DemoAdditionalParams
+    val config: List<String> = defaultDiktatConfig,
+    val language: Languages = Languages.KOTLIN,
+) : DemoAdditionalParams {
+    companion object {
+        val defaultDiktatConfig = """
```
nulls, line 21: maybe extract to a file in /resources?

0x6675636b796f75676974687562, line 21:
```suggestion
        @Language("YAML")
        val defaultDiktatConfig = """
```

Bacisally, both `YaMl` and `yMl` (in any case) work.

0x6675636b796f75676974687562 (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/diktat/DemoAdditionalParams.kt
sanyavertolet: I think of adding it to storage, but in this case I will need to get it from storage if and only if there is no config provided. Will be done later in next PRs

sanyavertolet: sure, thanks

sanyavertolet: Seems that commonMain does not have `@Language` annotation so will remove as was

petertrr: https://www.jetbrains.com/help/idea/using-language-injections.html#use-language-injection-comments

PR #1600, opened 2022-12-12
[CPG] Additional Node info on frontend

sanyavertolet:
### What's done:
 * Added `CpgNodeAdditionalInfo` data class into `CpgNodeAttributes`
 * Added card on frontend that displays the additional info
 * Modified `Node#toCpgNode` to save additional params to `CpgNode`

### How it looks: 
screenshot

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
@@ -6,6 +6,7 @@
 
 package com.saveourtool.save.frontend.components.views.demo
 
+import com.saveourtool.save.demo.cpg.CpgNodeAdditionalInfo
 import com.saveourtool.save.demo.cpg.CpgResult
 import com.saveourtool.save.frontend.components.basic.cardComponent
 import com.saveourtool.save.frontend.components.basic.cpg.graphEvents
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
```diff
@@ -98,11 +109,31 @@ val cpgView: VFC = VFC {
...
+                                            )
+                                        }
                                     }
                                     graphLoader {
                                         cpgGraph = cpgResult.cpgGraph
                                     }
                                 }
+                                div {
+                                    id = "collapse"
+                                    val show = selectedNodeName?.let { nodeName ->
```
Cheshiriks, line 124: Can you add some button to close this additional information window? Something like the X in the top right?

sanyavertolet: The window is closed when clicking on the same node once again. Do you think it might be better to add such a button?

Cheshiriks: Can it happen that this window will appear on top of the node that was clicked on?

sanyavertolet: It can, but the graph layout is moveable so there is no problem with that

sanyavertolet: Added close button to the top of the window

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
```diff
@@ -136,3 +177,82 @@ val cpgView: VFC = VFC {
...
+                        onClick = { setSelectedNodeName(null) }
+                        style = jso {
+                            top = "1%".unsafeCast<Top>()
+                            right = "1%".unsafeCast<Right>()
+                        }
+                    }
+                }
+            }
+            tbody {
+                listOf(
```
petertrr, line 215: Maybe make this a map of columns names to accessor functions and extract into a file-level variable? Then each `value` of the map can consume `additionalInfo` and return its result, and the map will be reused between rerenders.

petertrr (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
sanyavertolet: Ok, will do it

sanyavertolet: Done

PR #1607, opened 2022-12-13
Fix configuration for fs-storage mount paths in Minikube

petertrr:
* Sync mount paths of fs-storage PVs and properties in `ConfigMap`s for backend and demo. These values are in `_constants.tpl` instead of `values.yaml` because they refer to entities _inside_ the container and there is no need to configure them externally, but they need to match in `application.properties` (ConfigMap`) and Deployment spec
* Fix values-minikube.yaml: remove deprecated property, set agent count to 1
* Comment out `dockerTag` placeholder from `values.yaml` to force its explicit provision

Due to `dev` profile being activated in Minikube, files were being stored at an ephemeral path and lost after pod recreations


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestSuitesSourceController.kt
```diff
@@ -605,12 +605,6 @@ class TestSuitesSourceController(
                 .contentType(MediaType.APPLICATION_OCTET_STREAM)
                 .body(testSuitesSourceSnapshotStorage.download(it))
         }
-        .onErrorReturn(
```
petertrr, line 608: Checks for existence of keys are performed earlier and converted to `NOT_FOUND` there. This `onErrorReturn` basically was wrapping `ResponseStatusException`s with `404`s into another `404`, and message was being lost.

I believe if there is an error in `testSuitesSourceSnapshotStorage.download` after several existence checks, then it's okay to return a `500` error

cc @nulls

nulls (approved):

PR #1609, opened 2022-12-13
Replaced FileNotFoundException and FileAlreadyExistsException to checking storage

nulls:


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -162,23 +161,18 @@ class DownloadFilesController(
...
+        .switchIfEmptyToNotFound {
+            "File with key $fileKey is not found"
         }
-        .onErrorReturn(
-            FileNotFoundException::class.java,
-            ResponseEntity
-                .status(HttpStatus.NOT_FOUND)
+        .map {
+            logger.info("Sending file ${fileKey.name} to a client")
+            val content = fileStorage.download(fileKey)
```
akuleshov7, line 171: content is used once? - may be move it directly to a `.body(fileStorage.download(fileKey))`

nulls: done

petertrr (approved):

PR #1615, opened 2022-12-14
FC backgrounds

sanyavertolet:
### What's done:
* Moved all custom hooks to `CustomHooks.kt`
* Added custom hook to set background using css styles


nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/JsUtils.kt
```diff
@@ -75,3 +109,65 @@ external fun encodeURIComponent(str: String): String
...
+    jQuery('[data-toggle="tooltip"]').tooltip()
+""")
+}
+
+/**
+ * JS code lines to enable tooltip and popover.
+ *
+ * @return dynamic
+ */
+// language=JS
```
Cheshiriks, line 133: Is this comment necessary? If so, can you bring it to the Kdoc?

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/JsUtils.kt
sanyavertolet: No, as far as I know it is done just as it should be done In order to let Idea highlight the code

PR #1616, opened 2022-12-14
[CPG] Fix for wrong node click processing

sanyavertolet:
### What's done:
* Fixed an issue with node clicking - (problem was connected with react state updating, fixed by using `stateSetter((String?) -> String?)` instead of `stateSetter(String?)`)
* Moved closing `button` from `table` in order to prevent html warning

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
@@ -114,9 +114,9 @@ val cpgView: VFC = VFC {
                                     graphEvents {
                                         shouldHideUnfocusedNodes = true
                                         setSelectedNode = { newSelectedNodeName ->
-                                            setSelectedNodeName(
-                                                newSelectedNodeName.takeIf { it != selectedNodeName }
-                                            )
+                                            setSelectedNodeName { previousSelectedNodeName ->
+                                                newSelectedNodeName.takeIf { it != previousSelectedNodeName }
+                                            }
                                         }
                                     }
                                     graphLoader {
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
```diff
@@ -198,6 +198,16 @@ private fun ChildrenBuilder.displayCpgNodeAdditionalInfo(
 ) {
     div {
         className = ClassName("card card-body p-0")
+        button {
+            className = ClassName("btn p-0 position-absolute")
+            fontAwesomeIcon(faTimesCircle)
+            type = "button".unsafeCast<ButtonType>()
+            onClick = { setSelectedNodeName(null) }
+            style = jso {
+                top = "1%".unsafeCast<Top>()
```
akuleshov7, line 207: why exactly 1%?

akuleshov7 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/CpgView.kt
sanyavertolet: It was quite hard to come up with an idea of how to place the button in a right position. Iteratively it was found out that 1% is just what is needed. The position does not depend on screen size so it seems to be ok for now

sanyavertolet: Guess even `0%` might be better as far as the vertical position depends on the size of the whole div and it is not that good as far as it can change due to `code` section

PR #1621, opened 2022-12-14
Propagating CPG errors to the FE

akuleshov7:


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/cpg/CpgGraph.kt
```diff
@@ -91,16 +91,8 @@ data class CpgGraph(
...
-                CpgNode("2", CpgNodeAttributes("John", "#00FF00")),
-                CpgNode("3", CpgNodeAttributes("Sam", "#0000FF")),
-            ),
-            edges = listOf(
-                CpgEdge("1->2", "1", "2", CpgEdgeAttributes("KNOWS")),
-                CpgEdge("2->3", "2", "3", CpgEdgeAttributes("LIKES")),
-                CpgEdge("3->1", "3", "1", CpgEdgeAttributes("HATES")),
-            ),
+            nodes = emptyList(),
+            edges = emptyList(),
```
sanyavertolet, line 95: Probably we need to generate once the graph for code placeholder and put it here instead of an empty graph

sanyavertolet (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/cpg/CpgGraph.kt
akuleshov7: I do not get it. 

We work with a stub:
```
try {
     buildGraph
} catch {
     return stub
}
```

akuleshov7: Ah, you want to create some placeholder graph?

sanyavertolet: > Ah, you want to create some placeholder graph? For some placeholder code like "hello world"?

Yes, definitely

akuleshov7: Better to have crying cat instead of a stub :)

PR #1622, opened 2022-12-15
Move code analysis configuration to `gradle/plugins`, convert to script plugins

petertrr:
* Move configuration of diktat, detekt and spotless into `gradle/plugins`

Continuation of #948


sanyavertolet (approved):

nulls (approved):

nulls:
@petertrr after extracting it to `plugins`, can it be extracted to a separate repo and reuse in `save-cli`, `sarif-utils` and etc?

petertrr:
> @petertrr after extracting it to `plugins`, can it be extracted to a separate repo and reuse in `save-cli`, `sarif-utils` and etc?

Yes, it can, at least as git submodule it can be used as-is. I think converting these scripts to proper publishable plugins may require a bit more work, though.

PR #1624, opened 2022-12-15
added entity from common package to sandbox

nulls:


sanyavertolet (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxExecution.kt
```diff
@@ -27,6 +27,7 @@ class SandboxExecution(
     @Enumerated(EnumType.STRING)
     var status: ExecutionStatus,
     var sdk: String,
+    @Column(name = "user_id")
```
nulls, line 30: Hibernate fails with error: 
```
Table [execution] contains physical column name [user_id] referred to by multiple logical column names: [user_id], [userId]
```

It's due to `Execution` (which is loaded because of a common package).
Looks like `@JoinColumn`  should be set in camel case

akuleshov7 (approved):

save-sandbox/src/main/kotlin/com/saveourtool/save/sandbox/entity/SandboxExecution.kt
akuleshov7: where the ambiguity comes from?

akuleshov7 (approved):

PR #1630, opened 2022-12-16
[Orchestrator] Graceful stop of container

### LINKED ISSUE #1569: Graceful stop for agents in orchestrator
Need to add endpoint to send `com.saveourtool.save.agent.TerminateResponse` to agent.
Also, we cannot stop crashed agents in K8S. We need to send a terminate signal in such cases

~Also, need to rename `DockerService` to `AgentManager` or merge to existed `AgentService`~

nulls:
### What's done:
* Added ContainersCollection
* ...

It closes #1569


save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/HeartBeatInspector.kt
```diff
@@ -1,82 +0,0 @@
-package com.saveourtool.save.orchestrator.service
```
nulls, line 1: let's rollback this change to avoid a lot of changes

kgevorkyan (approved):

PR #1636, opened 2022-12-19
removed AgentState.STOPPED_BY_ORCH -- we never set it

nulls:


petertrr:
I guess `STOPPED_BY_ORCH` used to be the previous version of graceful shutdown, before we had graceful termination on `TerminateResponse`. Now agents either crash and are killed by `HeartbeatInspector` or stop gracefully.

petertrr (approved):

PR #1637, opened 2022-12-19
[DEMO] db support ph.II

sanyavertolet:
### What's done:
* Added entities for `GitRepo`, `Snapshot`, `Tool`
* Added repositories for `GitRepo`, `Snapshot`, `Tool`
* Added services for `GitRepo`, `Snapshot`, `Tool`
* Dropped `dependency` table


save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Tool.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.saveourtool.save.spring.entity.BaseEntity
+import javax.persistence.*
+
+/**
+ * @property gitRepo
+ * @property snapshot
+ */
+@Entity
+@Table(name = "tool")
+class Tool(
```
nulls, line 13: Tool relates only to github, maybe relates to our internal storage (file-based in first phase and S3 in second phase).
A service to download tool from github can be added as a separate service and as only one from several (manual upload for example)

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
```diff
@@ -0,0 +1,22 @@
...
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+/**
+ * @property coreTool dependant tool
+ * @property dependencyTool the tool that [coreTool] depends on
+ */
+@Entity
+@Table(name = "dependency")
+class Dependency(
```
nulls, line 15: it's a specific diktat cases -- do we really want to support it?
we can store a tool as a single artifact (even in case of diktat): as archive (tar, zip)

save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
```diff
@@ -3,13 +3,17 @@ package com.saveourtool.save.demo
...
+import org.springframework.boot.autoconfigure.domain.EntityScan
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 
 /**
  * An entrypoint for spring boot for save-demo
  */
 @EnableConfigurationProperties(ConfigProperties::class)
 @SpringBootApplication
+@EntityScan(
+    "com.saveourtool.save.demo.entity",
```
nulls, line 15: do we really need it? `@SpringBootApplication` should scan `com.saveourtool.save.demo` automatically

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/GitRepo.kt
```diff
@@ -0,0 +1,16 @@
...
+import javax.persistence.Entity
+import javax.persistence.Table
+
+/**
+ * @property organizationName
+ * @property toolName
+ */
+@Entity
+@Table(name = "git_repo")
+class GitRepo(
```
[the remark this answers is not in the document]
sanyavertolet: Makes sense, for now will just rename it

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DependencyService.kt
```diff
@@ -0,0 +1,24 @@
...
+@Service
+class DependencyService(
+    private val dependencyRepository: DependencyRepository,
+) {
+    /**
+     * @param coreTool
+     * @param dependencyTool
+     * @return [Dependency] entity that has been saved to database
+     */
+    fun addDependency(coreTool: Tool, dependencyTool: Tool): Dependency = dependencyRepository.findByCoreToolAndDependencyTool(
```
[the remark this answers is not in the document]
sanyavertolet: Oh sure

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/GitRepoService.kt
```diff
@@ -0,0 +1,31 @@
...
+    fun getRepos(): List<GitRepo> = gitRepoRepository.findAll()
+
+    private fun save(gitRepo: GitRepo): GitRepo = gitRepoRepository.save(gitRepo)
+
+    /**
+     * @param gitRepo entity to be saved
+     * @return [GitRepo] entity saved to database
+     */
+    @Transactional
+    fun saveIfNotPresent(gitRepo: GitRepo): GitRepo = gitRepoRepository.findByToolNameAndOrganizationName(
```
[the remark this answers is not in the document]
sanyavertolet: It is, will rename `addDependency`

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
sanyavertolet: Do we want to think of some universal format for saveourtool demo uploading for now?

sanyavertolet: I mean, probably we would like it to be an automated system. I am not sure if we would let developer upload an archive that contains everything in it.

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
```diff
@@ -0,0 +1,22 @@
...
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+/**
+ * @property coreTool dependant tool
+ * @property dependencyTool the tool that [coreTool] depends on
+ */
+@Entity
+@Table(name = "dependency")
```
Cheshiriks, line 14: This annotation is not needed here. It is needed when the class name is not equal to the table name.

sanyavertolet: Oh thanks, will remove it

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
nulls: We can automate uploading artifacts from github. Another system will be required another approach for uploading.

About universal format: I don't like the idea with dependencies. It's our specific case that we use ktlint which is a tool too.
Diktat can be converted to a single executable file (we actually already have a task for it). And as I remember, we wanted to limit our tools to a single executable file (but it's for save-cli, not a demo).

Can be left for now, but I think we will revise it in future

sanyavertolet: The truth is that I do not like the solution also but I could not come up with a better idea

save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
sanyavertolet: No, we do not, will remove it

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/GitRepo.kt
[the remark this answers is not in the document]
sanyavertolet: Renamed it to be `GithubRepo`

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DependencyService.kt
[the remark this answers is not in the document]
sanyavertolet: Done

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/GitRepoService.kt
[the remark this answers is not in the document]
sanyavertolet: Renamed

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
sanyavertolet: Completely removed `Dependency` and everything connected with it. Later will add two files: `setup.sh` and `run.sh` or something like that

save-demo/src/main/kotlin/com/saveourtool/save/demo/SaveDemo.kt
sanyavertolet: Done

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Dependency.kt
sanyavertolet: Removed useless `Table` annotations

nulls (approved):

PR #1638, opened 2022-12-19
Dependencies to MPP projects doesn't work from non MPP project

nulls:
### What's done:
- added sourceSets for commonMain for projects dependencies (MPP)
- moved spring boot bom to commonMain
- void error with default value of save-agent task


nulls:
It doesn't work: IDEA loads projects, but dependencies are missed in IDEA and gradle.

Using `main` instead of `commonMain` doesn't work too -- the original issue.

Looks like need to migrate all modules to MPP to resolve issue in IDEA

PR #1639, opened 2022-12-19
Fixed BOM for SLF4J

nulls:
### What's done:
* spring boot dependencies bom moved to commonMain
* avoid error in `downloadSaveAgentDistroTaskProvider`


nulls:
it doesn't work

PR #1640, opened 2022-12-19
Introduced ContainersCollection

nulls:
### What's done:
- hide all logic is related to container's info to a dedicated class

### DIFF
--- a/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/SaveOrchestratorCommonConfiguration.kt
+++ b/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/SaveOrchestratorCommonConfiguration.kt
@@ -1,7 +1,6 @@
 package com.saveourtool.save.orchestrator
 
 import com.saveourtool.save.orchestrator.config.ConfigProperties
-import com.saveourtool.save.orchestrator.utils.ContainersCollection
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.*
 import org.springframework.scheduling.annotation.EnableScheduling
@@ -16,8 +15,4 @@ import org.springframework.scheduling.annotation.EnableScheduling
 @EnableConfigurationProperties(ConfigProperties::class)
 @EnableScheduling
 @PropertySource("classpath:META-INF/save-orchestrator-common/application.properties")
-class SaveOrchestratorCommonConfiguration {
-    fun containersCollection(configProperties: ConfigProperties): ContainersCollection {
-        return ContainersCollection(configProperties.agentsHeartBeatTimeoutMillis)
-    }
-}
+class SaveOrchestratorCommonConfiguration
--- a/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/SaveOrchestratorCommonConfiguration.kt
+++ b/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/SaveOrchestratorCommonConfiguration.kt
@@ -1,6 +1,7 @@
 package com.saveourtool.save.orchestrator
 
 import com.saveourtool.save.orchestrator.config.ConfigProperties
+import com.saveourtool.save.orchestrator.utils.ContainersCollection
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.*
 import org.springframework.scheduling.annotation.EnableScheduling
@@ -15,4 +16,8 @@ import org.springframework.scheduling.annotation.EnableScheduling
 @EnableConfigurationProperties(ConfigProperties::class)
 @EnableScheduling
 @PropertySource("classpath:META-INF/save-orchestrator-common/application.properties")
-class SaveOrchestratorCommonConfiguration
+class SaveOrchestratorCommonConfiguration {
+    fun containersCollection(configProperties: ConfigProperties): ContainersCollection {
+        return ContainersCollection(configProperties.agentsHeartBeatTimeoutMillis)
+    }
+}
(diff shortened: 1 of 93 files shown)


save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/SaveOrchestratorCommonConfiguration.kt
```diff
@@ -15,4 +16,8 @@ import org.springframework.scheduling.annotation.EnableScheduling
 @EnableConfigurationProperties(ConfigProperties::class)
 @EnableScheduling
 @PropertySource("classpath:META-INF/save-orchestrator-common/application.properties")
-class SaveOrchestratorCommonConfiguration
+class SaveOrchestratorCommonConfiguration {
+    fun containersCollection(configProperties: ConfigProperties): ContainersCollection {
+        return ContainersCollection(configProperties.agentsHeartBeatTimeoutMillis)
```
nulls, line 21: it's not needed

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
```diff
@@ -104,32 +106,33 @@ class AgentService(
...
             }
-            .filter { (_, finishedContainerIds) -> finishedContainerIds.isNotEmpty() }
-            .flatMap { (executionId, finishedContainerIds) ->
+            .filter { finishedContainerIds -> finishedContainerIds.isNotEmpty() }
+            .flatMap { finishedContainerIds ->
                 log.info { "For execution id=$executionId all agents have completed their lifecycle" }
                 markExecutionBasedOnAgentStates(executionId, finishedContainerIds)
-                    .thenReturn(
+                    .then(Mono.fromCallable {
+                        orchestratorAgentStatusService.deleteAllByExecutionId(executionId)
```
kgevorkyan, line 129: May be rename this method, thus to point that we actually clear our collections and not removing containers from `db` or `docker`? As I remember, now `cleanup `is responsible for this

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
```diff
@@ -104,32 +106,33 @@ class AgentService(
...
                         containerRunner.cleanupAllByExecution(executionId)
-                    )
+                    })
             }
             .doOnSuccess {
                 if (it == null) {
-                    log.debug("Agents other than $containerId are still running, so won't try to stop them")
+                    log.debug("Agents for execution $executionId are still running, so won't try to stop them")
```
kgevorkyan, line 135: We cant determine now which concrete containers `are still running`?

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/ContainerService.kt
```diff
@@ -85,31 +81,22 @@ class ContainerService(
                     interval = configProperties.agentsStartCheckIntervalMillis.milliseconds,
                     numberOfChecks = configProperties.agentsStartTimeoutMillis / configProperties.agentsStartCheckIntervalMillis
                 ) {
-                    areAgentsHaveStarted.computeIfAbsent(executionId) { AtomicBoolean(false) }.get()
+                    !orchestratorAgentStatusService.containsAnyByExecutionId(executionId)
```
kgevorkyan, line 84: Also, can we rename this method is some manner it, to point out that we address here to our collection and not to `db`?

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/controller/HeartbeatController.kt
```diff
@@ -56,7 +56,6 @@ class HeartbeatController(private val agentService: AgentService,
         val containerId = heartbeat.agentInfo.containerId
         logger.info("Got heartbeat state: ${heartbeat.state.name} from $containerId under execution id=$executionId")
         return {
-            containerService.markAgentForExecutionAsStarted(executionId)
```
kgevorkyan, line 59: now we don't need any initialization? of collection maybe

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
nulls: no, we can't here. this method only waits that all agents are stopped by changing state in database

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
nulls: actually, I did it on purposes -- `HeartBeatInspector` with info about containers mimics\repeats `AgentStatusRepository` from backend, but now it's in-memory.

The class and field has name with prefix `orchestrator` to highlight that it's on orchestrator level

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/ContainerService.kt
nulls: I wanted to highlight it by prefix `orchestrator`. Can rename it to `agentStatusInMemoryRepository`

nulls: renamed the class and field name

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
nulls: renamed the class and field name

kgevorkyan (approved):
lgtm

PR #1641, opened 2022-12-19
Support uploading of avatars

### LINKED ISSUE #1558: Support uploading of media data properly (avatars)
1. Show "loading" button
2. Add some hints for users that IN THIS FORM YOU CAN UPLOAD MEDIA. At least change ON HOVER. Now none is able to understand, that they CAN change an avatar and they need to press on the image to upload it. Need to do it for project also.
3. Rename avatar, when you save it (so we won't have errors with Cyrillic symbols)
4. Make a single hierarchy for avatars of organizations and users
5. Add some editor: https://github.com/nhn/tui.image-editor https://javascript.plainenglish.io/how-to-make-a-personal-image-editor-with-react-in-less-than-10-minutes-dba0d45ff03d

Cheshiriks:
- added onHover for avatar
- renamed avatar, when we save it
- made a single hierarchy for avatars of organizations and users
IMG_5143


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
```diff
@@ -73,30 +69,23 @@ internal class AvatarController(
...
+            "Not found avatar for organization $organizationName"
         }
 
     private fun AvatarKey.toMonoResponse(): Mono<ByteBufferFluxResponse> = this.toMono()
         .filterWhen(avatarStorage::doesExist)
-        .flatMap {
-            avatarStorage.lastModified(this)
-                .map { lastModified ->
-                    ResponseEntity.ok()
-                        .cacheControl(CacheControl.maxAge(longExpirationTime.toJavaDuration()).cachePublic())
```
nulls, line 96: why do we removed caching?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractFileBasedStorage.kt
```diff
@@ -55,10 +55,6 @@ abstract class AbstractFileBasedStorage<K>(
         val contentPath = buildPathToContent(key)
         return Mono.fromCallable {
             contentPath.deleteIfExists()
-        }.doOnNext {
-            if (it) {
-                contentPath.parent.deleteDirectoriesTill(rootDir)
```
nulls, line 60: why do remove cleanup of directories ?

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
Cheshiriks: The cache was looking at changing the file path. But now the path does not change because the file name is equal to the organization. Only the content changes. Therefore, when changing the avatar, it did not change on the front.

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -51,32 +54,37 @@ class AvatarStorage(configProperties: ConfigProperties) :
...
      * Added for backward compatibility
      *
      * @return relative path to avatar image
      */
     fun getRelativePath(): String = when (type) {
-        AvatarType.ORGANIZATION -> "/$objectName/$imageName"
-        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$objectName/$imageName"
+        AvatarType.ORGANIZATION -> "/${AvatarStorage.ORGANIZATIONS_DIRECTORY}/$objectName"
+        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$objectName"
+        else -> ""
```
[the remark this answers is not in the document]
Cheshiriks: Yes. Changed it

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractFileBasedStorage.kt
nulls: `AbstractFileBasedStorage` is a common class -- it does affect all storages. Please rollback.

Also, even in case of `AvatarStorage`, this cleanup is required to delete folders when all avatars are removed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
nulls: so looks like we missed caching functionality for avatar now. Is there way to re-implement it with same file-name?
maybe introduce a version of file and use last modified time as version?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/AvatarType.kt
```diff
@@ -7,6 +7,11 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 enum class AvatarType {
+    /**
+     * default
+     */
+    NONE,
```
0x6675636b796f75676974687562, line 13: Please document what `NONE` actually means and what's its intended usage.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -276,6 +275,7 @@ class DownloadFilesController(
             when (type) {
                 AvatarType.ORGANIZATION -> organizationService.saveAvatar(owner, it)
                 AvatarType.USER -> userDetailsService.saveAvatar(owner, it)
+                else -> throw IllegalStateException("Not supported type: $type")
```
0x6675636b796f75676974687562, line 278: Better replace with NONE here. The whole `when`-expression will remain exhaustive, but we'll receive a compilation error immediately after we add a new avatar type (failing fast is good).

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -51,32 +54,37 @@ class AvatarStorage(configProperties: ConfigProperties) :
      * @return [Path] is built by [AvatarKey] object
      */
     override fun buildPathToContent(rootDir: Path, key: AvatarKey): Path = rootDir
-        .let { if (key.type == AvatarType.USER) it.resolve(USERS_DIRECTORY) else it }
+        .let {
```
0x6675636b796f75676974687562, line 57: For the sake of readability, this is better replaced with

```kotlin
run {
    // ...
    rootDir.resolve(/* ... */)
    // ...
}
```

Second, `resolve` can be replaced with `/` (`operator fun div`).

Up to you, though.

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -51,32 +54,37 @@ class AvatarStorage(configProperties: ConfigProperties) :
...
      * Added for backward compatibility
      *
      * @return relative path to avatar image
      */
     fun getRelativePath(): String = when (type) {
-        AvatarType.ORGANIZATION -> "/$objectName/$imageName"
-        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$objectName/$imageName"
+        AvatarType.ORGANIZATION -> "/${AvatarStorage.ORGANIZATIONS_DIRECTORY}/$objectName"
+        AvatarType.USER -> "/${AvatarStorage.USERS_DIRECTORY}/$objectName"
+        else -> throw IllegalStateException("Not supported type: $type")
```
0x6675636b796f75676974687562, line 88: Maybe

```suggestion
        AvatarType.NONE -> throw IllegalStateException("Not supported type: $type")
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -183,14 +182,15 @@ class RegistrationView : AbstractView<RegistrationProps, RegistrationViewState>(
     )
     private fun ChildrenBuilder.renderAvatar() {
         label {
+            className = ClassName("btn")
+            title = "Change user's avatar"
```
0x6675636b796f75676974687562, line 186:
```suggestion
            title = "Change the user's avatar"
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
nulls: @Cheshiriks, why we need NONE at all? it's invalid case and we don't need to support. Since `kotlin` supports validating `when` on compilation -- better to remove NONE and we will get `compilation` error when a new type of avatar will be added (as @0x6675636b796f75676974687562 suggested)

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractFileBasedStorage.kt
Cheshiriks: Rollbacked it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
nulls: looks like request with `cache-control` supports reload: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#reload_and_force_reload

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
Cheshiriks: Agree. Deleted it.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/AvatarType.kt
Cheshiriks: Deleted this type

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/AvatarStorage.kt
```diff
@@ -51,32 +54,37 @@ class AvatarStorage(configProperties: ConfigProperties) :
...
     val type: AvatarType,
     val objectName: String,
-    val imageName: String,
 ) {
     /**
      * Added for backward compatibility
      *
      * @return relative path to avatar image
      */
     fun getRelativePath(): String = when (type) {
```
nulls, line 85: btw, since now the path is common for all avatars and depends only on organization\user name -- maybe remove `ImageInfo` class and logic related to it

Cheshiriks: Ok. Agree. I'll refactor it in the next pr.

0x6675636b796f75676974687562 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -508,6 +508,7 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                 }
                 setState {
                     isUploading = false
+                    window.location.reload()
```
sanyavertolet, line 511: Why do you want to use it here? I am sure you can rerender the required part of a page instead of rendering the whole page for the second time

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -535,14 +536,15 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
                 label {
+                    className = ClassName("btn")
+                    title = "Change organization's avatar"
                     input {
                         type = InputType.file
                         hidden = true
                         onChange = { event ->
                             postImageUpload(event.target)
                         }
                     }
```
sanyavertolet, line 547: Suggestion to add a tooltip

Cheshiriks: The 'title' here displays a tooltip when you hover over it with the mouse. Or did you mean something else?

sanyavertolet: Yes, I meant `bootstrap`s tooltip, not the usual one (with data-toggle = tooltip). Is it here?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
Cheshiriks: When changing the avatar, we still have the same object with the same name. But the content of the bytearray is changing. Therefore, you need to re-render the page to get a new avatar.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Added screnshoot

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
Cheshiriks: Added version tag to url. In the next PR I will delete ImageInfo and change the cache.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
sanyavertolet: Yep, this is not `bootstrap` tooltip, to my mind it looks not that good. Check [this](https://getbootstrap.com/docs/4.6/components/tooltips/) out and find examples in our code (useTooltip hook can be used only inside `FC`s so either replace the hook with some code or use `(V)FC` wrapper).

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
sanyavertolet: Oh, I see
But what I suggest is to use some other way to re-render the page (only avatar). Thus there will be no useless re-render. Probably it might be a good idea to add FC wrapper for avatar so it could
1. re-render
2. display better tooltips

Can be done in next PR but if so then do it right after this one

Cheshiriks: OK. I will do it in the next PR. I won't refactor for now

sanyavertolet: Then please open a new issue and mention there both things I wrote above

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
sanyavertolet: As discussed above, will be done in next PR

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
```diff
@@ -170,10 +170,17 @@ class OrganizationService(
      * @throws NoSuchElementException
      */
     fun saveAvatar(name: String, relativePath: String) {
```
sanyavertolet, line 172: Shouldn't it be marked as `Transactional`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -509,6 +509,7 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                 setState {
                     isUploading = false
                 }
```
sanyavertolet, line 511: Is this state setter still required?

sanyavertolet (approved):

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
Cheshiriks: No. This annotation is needed when we have more than one method that changes the data in the database. In all other cases, it is bad to keep an open transaction to the database.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
Cheshiriks: No. Removed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/OrganizationService.kt
sanyavertolet: oh ok

PR #1651, opened 2022-12-22
[DEMO] hotfix for db support ph.IV

sanyavertolet:
### What's done:
* Added `docker-secrets` profile to `docker-compose.yaml` demo section
* Added `spring.factories` to enable `DockerSecretsDatabaseProcessor` in `save-demo`

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -109,7 +109,7 @@ services:
   demo:
     image: ghcr.io/saveourtool/save-demo:${DEMO_TAG}
     environment:
-      - "SPRING_PROFILES_ACTIVE=${PROFILE}"
+      - "SPRING_PROFILES_ACTIVE=${PROFILE},docker-secrets"
     ports:
       - "5421:5421"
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -109,7 +109,7 @@ services:
   demo:
     image: ghcr.io/saveourtool/save-demo:${DEMO_TAG}
     environment:
-      - "SPRING_PROFILES_ACTIVE=${PROFILE}"
+      - "SPRING_PROFILES_ACTIVE=${PROFILE},docker-secrets"
```
nulls, line 112: I guess you don't use authorization & authentication in save-demo yeat?

nulls (approved):

docker-compose.yaml
sanyavertolet: Yes, for now I do not use it as far as all endpoints either public or internal. Maybe later there will be endpoints that require authorization

PR #1657, opened 2022-12-23
Added checking execution status on finalization method

### LINKED ISSUE #1656: Orchestrator finalizes execution several times
Now it leads to an error is due to checking that orchestrator knows the stopped execution. Orchestrator forgets the execution after the first finalization

nulls:
### What's done:
- added checking execution status
- added/moved logging when containers are not stopped or execution is already in finale stage

It closes #1656

### DIFF
--- a/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
+++ b/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
@@ -121,8 +121,8 @@ class AgentService(
             }
             .filter { finishedContainerIds ->
                 finishedContainerIds.isNotEmpty()
-                    .also {
-                        if (!it) {
+                    .also { hasFinishedContainers ->
+                        if (!hasFinishedContainers) {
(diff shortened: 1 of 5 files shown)


save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/service/AgentService.kt
```diff
@@ -119,7 +119,25 @@ class AgentService(
...
+                    .also { hasFinishedContainers ->
+                        if (!hasFinishedContainers) {
+                            log.debug { "Agents for execution $executionId are still running, so won't try to stop them" }
+                        }
+                    }
+            }
+            .filterWhen {
+                orchestratorAgentService.getExecutionStatus(executionId)
+                    .map {
+                        it !in setOf(ExecutionStatus.FINISHED, ExecutionStatus.FINISHED)
```
kgevorkyan, line 133: set of `finished`, `finished`?

nulls: good catch, thanks. Should be `FINISHED` and `ERROR`

kgevorkyan (approved):

PR #1666, opened 2022-12-26
When serving binaries, report HTTP 404 on the back-end side, too

0x6675636b796f75676974687562:


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -182,10 +182,16 @@ class DownloadFilesController(
...
     // FIXME: backend should set version of save-agent here for agent
-    fun downloadSaveAgent(): Mono<out Resource> =
-            Mono.just(ClassPathResource("save-agent.kexe"))
-                .filter { it.exists() }
-                .switchIfEmptyToNotFound()
+    fun downloadSaveAgent(): Mono<out Resource> {
+        val executable = "save-agent.kexe"
+
+        return Mono.just(ClassPathResource(executable))
+            .filter { it.exists() }
```
sanyavertolet, line 189: Maybe so (and so on)? In order to keep the endpoint to be fully "chained"
```suggestion
    fun downloadSaveAgent(): Mono<out Resource> =  Mono.just("save-agent.kexe")

        .map { ClassPathResource(it) }
        .filter { it.exists() }
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -202,12 +208,16 @@ class DownloadFilesController(
...
     fun downloadSaveCliByVersion(
         @RequestParam version: String,
-    ): Mono<out Resource> =
-            Mono.just(ClassPathResource("save-$version-linuxX64.kexe"))
-                .filter { it.exists() }
-                .switchIfEmptyToNotFound {
-                    "Can't find save-$version-linuxX64.kexe with the requested version $version"
-                }
+    ): Mono<out Resource> {
+        val executable = "save-$version-linuxX64.kexe"
```
sanyavertolet, line 212: Same here

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
nulls: +1
even more -- it can be extracted to a common private method and reused in two methods

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -182,10 +182,16 @@ class DownloadFilesController(
     @ApiResponse(responseCode = "404", description = "File is not found.")
     @PostMapping(path = ["/internal/files/download-save-agent"], produces = [MediaType.APPLICATION_OCTET_STREAM_VALUE])
```
nulls, line 183: maybe add a GetMappint to the same path? just for testing in browser as you suggested

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
0x6675636b796f75676974687562: Addressed with d4bfadba7521d96c62a41600325762e1f23920c9

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
0x6675636b796f75676974687562: Addressed with d4bfadba7521d96c62a41600325762e1f23920c9

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
0x6675636b796f75676974687562: Addressed with 9aef36aa403eca710ac80fb661d42aee0ddbcc06

PR #1670, opened 2022-12-27
Remove `FileInfo`, `FileKey`

nulls:
### What's done:
- [x] #1672 -- now it's an abstract class which can be extended by DTO which wants to store ID from database
- [x] #1674
- [x] #1680
- [x] #1679
- [x] #1697
- [x] #1699
- [x] #1683
- [x] removed `FileInfo`, `FileKey`

It closes #1206

PR #1672, opened 2022-12-28
Refactored DtoWithId

nulls:
### What's done:
* refactored DtoWithId to abstract class
* changed TestSuite and TestSuitesSource

It's part of #1670


sanyavertolet (approved):
lgtm

PR #1683, opened 2022-12-29
Introduce a link between execution and file as a dedicated table

nulls:
### What's done:
- added table `lnk_execution_file`
- migrated `Execution.additonalFiles` to using link table

It's part of https://github.com/saveourtool/save-cloud/pull/1670


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/FileInternalController.kt
```diff
@@ -24,42 +26,21 @@ import reactor.kotlin.core.publisher.toMono
 @RestController
 @RequestMapping("/internal/files")
 class FileInternalController(
-    private val fileStorage: MigrationFileStorage,
+    private val fileStorage: NewFileStorage,
```
nulls: need to rollback, we will use the old endpoints in phase 1

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/FileStorage.kt
```diff
@@ -1,54 +1,60 @@
...
+import reactor.core.publisher.Mono
+import java.nio.ByteBuffer
+import java.time.Instant
+import javax.annotation.PostConstruct
 
 /**
  * Storage for evaluated tools are loaded by users
  */
 @Service
 class FileStorage(
```
nulls: need to rollback

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/OldFileStorage.kt
```diff
@@ -0,0 +1,76 @@
+package com.saveourtool.save.backend.storage
```
nulls: should be removed

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Execution.kt
```diff
@@ -85,7 +85,7 @@ class Execution(
 
     var sdk: String,
 
-    var additionalFiles: String,
+    var additionalFiles: String = "",
```
nulls: can be removed

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/FileInternalController.kt
nulls: done

save-backend/src/main/kotlin/com/saveourtool/save/backend/storage/OldFileStorage.kt
nulls: done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
```diff
@@ -287,6 +289,26 @@ class ExecutionService(
         return savedExecution
     }
 
+    /**
+     * Unlink provided [File] from all [Execution]s
+     *
+     * @param file
+     */
+    @Transactional
+    fun unlinkFileFromAllExecution(file: File) {
```
kgevorkyan, line 298: It's not actually an `unlink`, isn't it? We don't remove corresponding records from `lnkExecutionFileRepository`
So, maybe call it like `markeExecutionsCorrespondingToFileAsObsolete`?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
nulls: actually, it should delete `lnkExecutionFileRepository`.
Fixed the method

PR #1687, opened 2022-12-30
Refactoring: deleted ImageInfo

### LINKED ISSUE #1673: Delete ImageInfo
Now we do not store the original name of the image, but rename it by the name of the user or organization. Therefore, we no longer need this class.

- Optimize image caching.

Cheshiriks:
- deleted ImageInfo

Closes #1673


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -535,7 +527,7 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                     }
                     img {
                         className = ClassName("avatar avatar-user width-full border color-bg-default rounded-circle")
-                        src = state.image?.path?.let {
+                        src = state.organization?.avatar?.let {
```
nulls, line 530: does it mean that we should now re-load `state.organization` after uploading an image?
previously we set `image` and used it to take an url

nulls: or window.location.reload() does it?

Cheshiriks: Yes. Added reload() for it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -110,30 +111,23 @@ class DownloadFilesController(
...
+            AvatarType.USER -> userDetailsService.saveAvatar(owner, path)
+            else -> throw IllegalStateException("Not supported type: $type")
         }
-        ResponseEntity.status(
-            imageInfo.path?.let {
-                HttpStatus.OK
-            } ?: HttpStatus.INTERNAL_SERVER_ERROR
-        )
-            .body(imageInfo)
+        ResponseEntity.status(HttpStatus.OK).body("Saved")
```
kgevorkyan, line 130: "Image was successfully uploaded on the path: blabla"? Or smth like that

Cheshiriks: The text has changed, but the path is not needed here. It is now unified.

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DownloadFilesController.kt
```diff
@@ -110,30 +111,23 @@ class DownloadFilesController(
...
-            when (type) {
-                AvatarType.ORGANIZATION -> organizationService.saveAvatar(owner, it)
-                AvatarType.USER -> userDetailsService.saveAvatar(owner, it)
-                else -> throw IllegalStateException("Not supported type: $type")
-            }
+    }.map { path ->
+        when (type) {
+            AvatarType.ORGANIZATION -> organizationService.saveAvatar(owner, path)
+            AvatarType.USER -> userDetailsService.saveAvatar(owner, path)
+            else -> throw IllegalStateException("Not supported type: $type")
```
0x6675636b796f75676974687562, line 128: This last branch is redundant, isn't it?

0x6675636b796f75676974687562 (approved):

PR #1689, opened 2022-12-30
demo manager controller check permission in frontend

### LINKED ISSUE #1676: Implement permission check on `backend`'s `DemoManagerController`
Right now there is no permission check except `@PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")`. This annotation should be removed and `ProjectPermissionEvaluator` should be used in order to:

1. `READ` is required for getting `DemoStatus`;
2. `WRITE` is required for getting `DemoInfo`;
3. `DELETE` is required for demo adding.

This permissions should also be supported on frontend in order to lock some buttons (e.g. forbid users to create demo as well as edit input forms).

This task should be done in two steps (two PRs):
1. Backend part
2. Frontend part

Arrgentum:
## Whats added:
- added a modal window with information about request errors 

## Issues:
This pull request close frontend path in #1676
Close #1676

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
@@ -52,7 +52,7 @@ val projectDemoMenu: FC<ProjectDemoMenuProps> = FC { props ->
                         if (it.ok) {
                             setDemoStatus(DemoStatus.STARTING)
                         } else {
-                            props.updateErrorMessage(it, it.unpackMessage())
+                            props.updateErrorMessage(it.statusText, it.unpackMessage())
                         }
                     }
             }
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
```diff
@@ -211,4 +217,10 @@ external interface ProjectDemoMenuProps : Props {
      * Organization name
      */
     var organizationName: String
+
+    /**
+     * Callback to show error message
+     */
+    @Suppress("TYPE_ALIAS")
+    var updateErrorMessage: (Response, String) -> Unit
```
nulls, line 225: can we change it to `(String, String)` -- title and message

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
Arrgentum: Done

Issue #1575, opened 2022-12-07
Move scheduler events from the database to the backend.


Cheshiriks:

nulls:
`Execution` has a few related entities which should be deleted too:
- lnk_execution_test_suite
- lnk_execution_file
- agent
- agent_status
- test_execution

Currently, we delete them explicitly when `Execution` became `OBSOLETE`.
Since we remove an execution after 30 days, the links can be left, they will be deleted by cascade constrains.

Need to revise current cleanup logic after moving this logic to backend: we can mark `Execution` as `OBSOLETE` instead of deletion and have two values for retentions (OBSOLETE and remove)

PR #1697, opened 2023-01-09
Moved delete and marking as obsolete of Execution to ExecutionService

nulls:
It's part of #1670

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
@@ -66,8 +66,15 @@ class ExecutionService(
      */
     @Transactional
     fun updateExecutionStatus(srcExecution: Execution, newStatus: ExecutionStatus) {
-        log.debug("Updating status to $newStatus on execution id = ${srcExecution.requiredId()}")
-        val execution = executionRepository.findWithLockingById(srcExecution.requiredId()).orNotFound()
+        val executionId = srcExecution.requiredId()
+        log.debug("Updating status to $newStatus on execution id = $executionId")
+        if (newStatus == ExecutionStatus.OBSOLETE) {
+            log.info {
+                "Marking execution with id $executionId as obsolete. Additionally cleanup dependencies to this executions"
+            }
+            doDeleteDependencies(listOf(executionId))
+        }
+        val execution = executionRepository.findWithLockingById(executionId).orNotFound()
         val updatedExecution = execution.apply {
             status = newStatus
         }
@@ -290,24 +297,8 @@ class ExecutionService(
         log.info {
             "Deleting executions with id in $executionIds. Additionally cleanup dependencies to these executions"
         }
-        doDeleteDependencies(executionIds)
+        // dependencies will be cleanup by cascade constrains
         executionRepository.deleteAllById(executionIds)
-    }
-
-    /**
-     * Mark [Execution] as [ExecutionStatus.OBSOLETE] by ID
-     *
-     * @param executionIds
-     */
-    @Transactional
-    fun markAsObsoleteByIds(executionIds: List<Long>) {
-        log.info {
-            "Marking executions with id in $executionIds as obsolete. Additionally cleanup dependencies to these executions"
-        }
-        doDeleteDependencies(executionIds)
-        executionIds.forEach {
-            updateExecutionStatus(getExecution(it), ExecutionStatus.OBSOLETE)
-        }
     }
 
     private fun doDeleteDependencies(executionIds: List<Long>) {
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
@@ -117,17 +117,6 @@ class ExecutionService(
             executionRepository.findByProjectNameAndProjectOrganizationAndStartTimeBetween(name, organization, start, end)
 
     /**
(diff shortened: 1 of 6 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
```diff
@@ -302,6 +280,46 @@ class ExecutionService(
...
+        log.info {
+            "Marking executions with id in $executionIds as obsolete. Additionally cleanup dependencies to these executions"
+        }
+        doDeleteDependencies(executionIds)
+        executionIds.forEach {
+            updateExecutionStatus(getExecution(it), ExecutionStatus.OBSOLETE)
+        }
+    }
+
+    private fun doDeleteDependencies(executionIds: List<Long>) {
```
Cheshiriks, line 313: I don't think it's necessary. Tables have a cascade constraints. When deleting an execution, the data associated with it should be deleted from the rest of the tables.

nulls: Agree, actually in one place we forget to remove LnkExecutionTestSuite and it was being cleanup by a cascade constrains.
But we still need to have it for obsolete, we can remove it for deleting an execution.

@petertrr @akuleshov, what's do you think about stopping deleting related entities on deleting and rely on database mechanism?

akuleshov7: I already had some problems with such delete, do you remember?

.github/workflows/deploy_images.yml
```diff
@@ -101,6 +101,11 @@ jobs:
         with:
           java-version: 11
           distribution: temurin
+      - name: Set up Python 3.10.9
+        id: python
```
akuleshov7, line 105: oh, it was a bad idea to rebase. This code was removed in master

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
```diff
@@ -190,36 +185,19 @@ class TestSuitesService(
         ) {
             "Expected that we remove all test suites related to a single execution at once"
         }
-        executionIds.forEach { executionId ->
-            log.debug { "Delete link between execution $executionId and test suites" }
-            lnkExecutionTestSuiteService.deleteByExecution(executionId)
-        }
+        executionService.markAsObsoleteByIds(executionIds)
```
akuleshov7, line 188: What do you mean by "obsolete" ? Legacy? Deleted? Why we need to have obsolete executions? How we get obsolete executions? I do not get this logic...

akuleshov7: Ah, just moved logic to the Service?

nulls: > I do not get this logic...

It's existed logic of execution archiving 

> Ah, just moved existing logic to the Service?

yes

.github/workflows/deploy_images.yml
nulls: I guess the problem is in `force-pushed the master branch from d92be76 to 65c52ab`

nulls: rebased current PR

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
nulls: > I already had some problems with such delete, do you remember?

We have a problem when execution's history was cleaned up after recreating (on fetching the same version) tests suite source. `Cascade constrains` removes invalid links

nulls: > I do not remember what kind of problem, but we needed to rework delete at that time...

Current PR is just refactoring to avoid changes in #1670 when link to files will be added

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
nulls: will move this logic to `ExecutionStatus#updateExecutionStatus` to hide this method inside in service

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
nulls: Will remove explicit deletion of dependencies on deletion of execution

nulls: @Cheshiriks done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
nulls: removed

akuleshov7 (approved):

0x6675636b796f75676974687562 (approved):

Cheshiriks (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
nulls: added a note to https://github.com/saveourtool/save-cloud/issues/1575#issuecomment-1376928017

PR #1699, opened 2023-01-10
Added AbstractMigrationStorage

nulls:
### What's done:
- added AbstractMigrationStorage
- refactored MigrationFileStorage

It's part of #1670

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractMigrationStorage.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractMigrationStorage.kt
@@ -1,0 +1,100 @@
+package com.saveourtool.save.storage
+
+import com.saveourtool.save.utils.getLogger
+import com.saveourtool.save.utils.info
+import com.saveourtool.save.utils.warn
+import org.slf4j.Logger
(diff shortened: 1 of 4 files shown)


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractMigrationStorage.kt
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    private fun doMigrate() {
+        oldStorage.list()
+            .map { oldKey ->
+                oldKey to oldKey.toNewKey()
+            }
+            .filterWhen { (_, newKey) ->
+                newStorage.doesExist(newKey).map { !it }
```
kgevorkyan, line 44: Any debug info here, maybe?

kgevorkyan (approved):
lgtm

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractMigrationStorage.kt
nulls: added debug info when key already exists in new storage

PR #1703, opened 2023-01-10
Return HTTP 400 Bad Request if multiple sources or multiple versions are specified

0x6675636b796f75676974687562:


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
```diff
@@ -312,23 +317,29 @@ class ExecutionService(
...
 
-        private fun Collection<TestSuite>.singleVersion(): String = map { it.version }
+        private fun Collection<TestSuite>.singleVersion(): Either<ErrorMessage, String> = map { it.version }
             .distinct()
-            .also { versions ->
-                require(versions.size == 1) {
-                    "Only a single version is supported, but got: $versions"
+            .let { versions ->
+                when (versions.size) {
+                    1 -> versions[0]
```
nulls, line 337: maybe handle case when size is 0?

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/ExecutionService.kt
0x6675636b796f75676974687562: Leaving as-is as discussed.

PR #1707, opened 2023-01-10
When requesting text executions, return extended versions (`TestExecutionExDto`)

0x6675636b796f75676974687562:


kgevorkyan (commented):
btw, can we call `TestExecutionExDto`  ->`TestExecutionDtoWithTestMetcris` or smth like this?

Initially, I thought, that TestExecution Extended is equals, or partially equals to TestExecution, and `exteded()` is reverse operation for `toDto`
But this is different entity

Or even more, may be we can put test metrics to TestExecution and hold them in TestExecutionDto via `toDto()` method?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestExecutionController.kt
```diff
@@ -100,6 +101,9 @@ class TestExecutionController(
                     }
             }
         }
+        .map { testExecution ->
+            testExecution.extended()
```
sanyavertolet, line 105: maybe `toExtended()` or `extend()`? `extended()` might mean to return true (like in cpp)

sanyavertolet (approved):

save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
```diff
@@ -2,7 +2,7 @@
 
 package com.saveourtool.save.api
 
-import com.saveourtool.save.agent.TestExecutionDto
+import com.saveourtool.save.agent.TestExecutionExDto
```
sanyavertolet, line 5: Also it might be better to call it `TestExecutionExtDto` as far as `ext` is usually short for `extended` imo

nulls (approved):

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/TestExecutionController.kt
0x6675636b796f75676974687562: Renamed to `toExtended()` in cf3d3a453d455beeb84b178cf78b4e6d55fd8184.

save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
0x6675636b796f75676974687562: Renamed to `TestExecutionExtDto` in cf3d3a453d455beeb84b178cf78b4e6d55fd8184.

PR #1710, opened 2023-01-11
Removed checking contentSize in AbstractStorageWithDatabase

nulls:


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractStorageWithDatabase.kt
```diff
@@ -74,17 +71,12 @@ abstract class AbstractStorageWithDatabase<K : DtoWithId, E : BaseEntityWithDtoW
     }
         .flatMap { entity ->
             storage.upload(entity.requiredId(), content)
-                .filter { it > 0 }
```
kgevorkyan, line 77: We can't hold it, but with `>=0`?

nulls: we can't, because `storage` doesn't guarantee that such object will be cleanup.
Also, it's not a responsibility of this storage to check content size: negative or zero content size don't break logic of `AbstractStorageWithDatabase`. We should get `Mono.error` in case of invalid content.
now, we have `AbstractFileBasedStorage` which supports empty files and returns contentSize = 0.
Checking on content size we have on Controller layer (we return 500 to user in such case)

kgevorkyan: > I think we should not, because `storage` doesn't guarantee that such object will be cleanup when content size is invalid.
Why it doesn't guarantee this?

nulls: > Why it doesn't guarantee this?

Because if storage returns something (content size), it saves something (object). But I agree that it's negative content size is impossible due to nature of this value.

But still, it's not a responsibility of `AbstractStorageWithDatabase` -- it writes to another `storage` and writes the result to database

kgevorkyan (approved):

kgevorkyan (approved):

PR #1713, opened 2023-01-12
Fix calculating `passRate`

nulls:


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ExecutionLabels.kt
```diff
@@ -96,7 +96,8 @@ internal class ExecutionStatisticsValues(executionDto: ExecutionDto?) {
         this.failedTests = executionDto?.failedTests?.toString() ?: "0"
         this.runningTests = executionDto?.runningTests?.toString() ?: "0"
         this.passRate = executionDto
-            ?.let { "${calculateRate(it.passedTests, it.allTests)}" }
+            ?.let { calculateRate(it.passedTests, it.allTests) }
+            ?.toString()
```
akuleshov7, line 100: When have we broken it?

akuleshov7 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/ExecutionLabels.kt
nulls: I guess it was done in such way initially.
`allTests == 0` is wear case

Cheshiriks (approved):

PR #1715, opened 2023-01-12
Properly escape URL parameters when requesting `/test-executions`

0x6675636b796f75676974687562:


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
```diff
@@ -405,7 +406,12 @@ class ExecutionView : AbstractView<ExecutionProps, ExecutionState>(false) {
             filters = state.filters
             getData = { page, size ->
                 post(
-                    url = "$apiUrl/test-executions?executionId=${props.executionId}&page=$page&size=$size&checkDebugInfo=true",
+                    url = "$apiUrl/test-executions?" + URLSearchParams(jso {
```
kgevorkyan, line 409: Do we need to do the same with all other requests? Or why did you change only this one?

kgevorkyan (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
kgevorkyan: Could you create issue on this subject then? I also like this suggestion

nulls: If it's more safe/proper, can we move request params to `com.saveourtool.save.frontend.utils.RequestUtilsKt#post()` as map?

0x6675636b796f75676974687562: @nulls, please take a look -- I believe your suggestion is addressed with 647a52620d8d4da194a6e13194e83e2c37a534d7.

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ExecutionView.kt
0x6675636b796f75676974687562: > Could you create issue on this subject then? I also like this suggestion

@kgevorkyan, see #1716.

PR #1717, opened 2023-01-12
Add metadata about test suites source snapshots in database 

### LINKED ISSUE #1714: Store test suite snapshot using git commit as unique ID and branch name or tags as additional name

nulls:
### What's done:
- [x] #1722
- [x] #1723
- [x] #1725
- [x] #1727
- [x] #1729 
- [x] #1730
- [x] #1733
- [x] #1736
- [x] #1731
- [x] #1750 
- [x] #1751
- [x] #1752

It closes #1714 

----
This PR contains:
- cleanup repository's methods (unused) + renaming (`_` in function name only between entities)
- cleanup unused internal endpoints
- migrated `TestSuite` to `TestsSourceSnapshot`: now `TestSuite` is created using `TestsSourceSnapshot` as a source instead of `TestSuitesSource` + `version`.
- deletion of tests version deletes only `TestsSourceVersion`, when all `TestsSourceVersion` related to a single `TestsSourceSnapshot` are deleted, it triggers deletion of `TestsSourceSnapshot` with further deletion of `TestSuite` -> `Test` and marking `Execution`s as **OBSOLETE**
- `Execution.version` is `tests version` actually: changed it. It's a string and we validate that it's a valid version only when a new execution is created and version is provided (we do not provide it in contest mode). Version has a default value -- commitId
- Improved storage with database:  we validate that storage doesn't contain ids which are missed in database.


db/test-data/tests-source-snapshot-insert.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+    <changeSet id="tests-source-snapshot-insert" author="nulls" context="dev">
```
Cheshiriks, line 7: Let's load the data from the .csv file. It's shorter and more convenient.

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestsSourceVersionService.kt
```diff
@@ -122,11 +123,9 @@ class TestsSourceVersionService(
     }
 
     private fun doDelete(versionEntity: TestsSourceVersion) {
-        // clean-up [TestsSuite]
-        testSuitesService.deleteTestSuite(versionEntity.snapshot.source, versionEntity.name)
         versionRepository.delete(versionEntity)
         val snapshotEntity = versionEntity.snapshot
-        if (versionRepository.findAllBySnapshot(snapshotEntity).isEmpty()) {
+        if (versionRepository.findAllBySnapshotId(snapshotEntity.requiredId()).isEmpty()) {
```
Cheshiriks, line 128: There is no difference, the search is still carried out by ID.

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestsSourceVersionService.kt
nulls: ok, do you suggest return back to search by entity?

Cheshiriks: No need. Just saying for the future :)

db/test-data/tests-source-snapshot-insert.xml
nulls: migrated `tests-source-snapshot-insert.xml` and `tests-source-version-insert.xml`

kgevorkyan (commented):
btw, could you please also add some more represented description in such large PR's?
I see the general direction, but there are a lot of deletion and so on

So maybe like:
* Move blabla into DB
* Remove useless endpoints, due the direct interaction with DB
* ....

Cheshiriks (approved):

nulls:
> btw, could you please also add some more represented description in such large PR's?

updated the description

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TestSuitesDisplayer.kt
```diff
@@ -23,17 +25,69 @@ import react.dom.html.ReactHTML.small
...
+        pluginsGetter = TestSuiteVersioned::plugins,
+    )
+}
+
+/**
+ * @param testSuites
+ * @param selectedTestSuiteId
+ * @param onTestSuiteClick
+ */
+fun ChildrenBuilder.showAvailableTestSuites(
```
nulls, line 52: need to rollback this change, it's leftover

0x6675636b796f75676974687562 (approved):

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/TestSuitesControllerTest.kt
```diff
@@ -47,64 +46,59 @@ class TestSuitesControllerTest {
...
             expectBody<TestSuite>()
                 .consumeWith {
                     val body = it.responseBody!!
                     assertEquals(testSuite.name, body.name)
-                    assertEquals(testSuite.source.name, body.source.name)
-                    assertEquals(testSuite.source.organizationName, body.source.organization.name)
-                    assertTrue(testSuite.source.latestFetchedVersion != body.source.latestFetchedVersion)
-                    assertEquals(testSuite.version, body.source.latestFetchedVersion)
-                    assertEquals(testSuite.version, body.version)
+                    // FIXME
```
0x6675636b796f75676974687562, line 68: Are you expecting to fix this?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/ReactorUtils.kt
```diff
@@ -189,7 +189,7 @@ fun <T : Any> blockingToMono(supplier: () -> T?): Mono<T> = supplier.toMono()
  * @param supplier blocking operation like JDBC
  * @return [Flux] from result of blocking operation [List] of [T]
  */
-fun <T> blockingToFlux(supplier: () -> Iterable<T>): Flux<T> = blockingToMono(supplier).flatMapIterable { it }
+fun <T> blockingToFlux(supplier: () -> Iterable<T>?): Flux<T> = blockingToMono(supplier).flatMapIterable { it }
```
0x6675636b796f75676974687562, line 192: Where exactly do we need this signature change?

A nullable iterable should normally be replaced with an empty iterable instead.

save-backend/src/test/kotlin/com/saveourtool/save/backend/controller/TestSuitesControllerTest.kt
nulls: yes, doing it right now

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestsSourceVersionService.kt
```diff
@@ -81,22 +74,30 @@ class TestsSourceVersionService(
...
         sourceName
     ).mapTo(HashSet(), TestsSourceVersion::name)
 
+    /**
+     * @param snapshotId
+     * @return all fetched version of [TestsSourceVersion] found by provided values
+     */
+    fun getAllVersions(
+        snapshotId: Long,
+    ): Set<String> = versionRepository.findAllBySnapshotId(snapshotId).mapTo(HashSet(), TestsSourceVersion::name)
```
0x6675636b796f75676974687562, line 99: May I suggest that you replace a HS with an LHS here, so that the iteration order of versions is preserved?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
```diff
@@ -21,6 +20,7 @@ import org.springframework.transaction.annotation.Transactional
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.bodyToMono
 import reactor.core.publisher.Mono
+import kotlin.io.path.*
```
0x6675636b796f75676974687562, line 23: Unused import?

0x6675636b796f75676974687562 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
nulls: yeap. removed already

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestsSourceVersionService.kt
nulls: changed

PR #1723, opened 2023-01-16
Move checking tests source version on backend side

nulls:
### What's done:
* moved checking to backend side
* checking the full commit id -- currently it should return a false for all cases 

It's part of #1717


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
```diff
@@ -151,12 +153,45 @@ class TestSuitesSourceService(
...
+                    sourceName = testSuitesSource.name,
+                    version = version,
+                )
+                    .filter { it }
+                    .switchIfEmptyToResponseException(HttpStatus.INTERNAL_SERVER_ERROR) {
+                        "Failed to delete existed version $version in ${testSuitesSource.organizationName}/${testSuitesSource.name}"
+                    }
+                    .thenReturn(true)
+            } else if (doesContain) {
+                log.debug {
```
kgevorkyan, line 178: may be move into separate variable this msg? seems, it duplicates

kgevorkyan (approved):
lgtm

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesSourceService.kt
nulls: will extract to a dedicated method

0x6675636b796f75676974687562 (approved):

PR #1724, opened 2023-01-16
Refactor `TestExecutionController#getTestExecutions()` so that it enriches test executions with data from `TestAnalysisService`

0x6675636b796f75676974687562:
This is a part of Flaky Test Detector, see #659.


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/ReactorUtils.kt
```diff
@@ -128,6 +196,22 @@ fun <T> T.toFluxByteBufferAsJson(objectMapper: ObjectMapper): Flux<ByteBuffer> =
  */
 fun <T : Any, K : Comparable<K>> Flux<T>.sortBy(keyExtractor: (T) -> K): Flux<T> = sort(Comparator.comparing(keyExtractor))
 
+/**
+ * Exposes this tuple as a data class, allowing destructuring.
+ *
+ * @return the 1st component of this tuple.
+ */
+operator fun <T : Any> Tuple2<T, *>.component1(): T =
```
nulls, line 204: there is
```kotlin
import reactor.kotlin.core.util.function.component1
import reactor.kotlin.core.util.function.component2
```

why it doesn't fit for us?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/ReactorUtils.kt
```diff
@@ -179,3 +263,12 @@ fun downloadFromClasspath(
                 logger.error("$resourceName is not found on the classpath; returning HTTP 404...")
                 lazyResponseBody()
             }
+
+/**
+ * @param seed the starting value of this sequence.
+ * @return the sequence infinitely returning [seed].
+ */
+fun <T : Any> infiniteSequenceOf(seed: T): Sequence<T> =
```
nulls, line 271: why do we need this method? it replaces a single call of method with a single lambda to a single call of method with a a single parameter

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/ReactorUtils.kt
0x6675636b796f75676974687562: Was unaware of those, thanks!

Addressed with 9660cb5e85bef5e35aa3c733cd2c47c96893c653.

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/ReactorUtils.kt
0x6675636b796f75676974687562: Addressed with 8922ec9b9fc1eb42e0e5fcd3c715a9b0aa461675.

test-analysis-core/src/main/kotlin/com/saveourtool/save/test/analysis/api/TestIdGenerator.kt
```diff
@@ -42,3 +52,55 @@ fun interface TestIdGenerator {
...
+ * @param pluginName the name of the `save-cli` plug-in ("warn", "fix", etc.).
+ * @param filePath the path to the file within the test suite.
+ * @return the generated unique test id.
+ */
+@Suppress(
+    "LongParameterList",
+    "TOO_MANY_PARAMETERS",
+)
+fun TestIdGenerator.testId(
+    organizationName: OrganizationName,
```
nulls, line 73: why do we need all these names, if test has `id` from database?

0x6675636b796f75676974687562: As discussed, extracted into a separate issue: #1726.

PR #1727, opened 2023-01-16
Upload snapshot of tests in 2 steps

nulls:
Now we upload tests using `commitId` and after it upload it them again using `version` provided by user.
Current implementation copied snapshot to new version (we duplicate content)

### What's done:
* added `TestsSourceSnapshotInfo`
* refactored upload tests

It's part of #1717


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/TestSuitesSourceInternalController.kt
```diff
@@ -24,41 +27,62 @@ import reactor.kotlin.core.publisher.toMono
...
+        val key = TestSuitesSourceSnapshotKey(
+            organizationName = snapshotInfo.organizationName,
+            testSuitesSourceName = snapshotInfo.sourceName,
+            version = snapshotInfo.commitId,
+            creationTime = snapshotInfo.commitTime,
+        )
         val content = part.content().map { it.asByteBuffer() }
-        testSuitesSourceVersionService.upload(versionInfo, content).map { writtenBytes ->
-            log.info { "Saved ($writtenBytes bytes) snapshot of ${versionInfo.sourceName} in ${versionInfo.organizationName} with version ${versionInfo.version}" }
+        snapshotStorage.upload(key, content).map { writtenBytes ->
```
kgevorkyan, line 53: Could we move such logging into the storage class directly? I think we will need it for any type of `upload/copy/etc`

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/test/TestsSourceVersionInfo.kt
```diff
@@ -6,19 +6,13 @@ import kotlinx.serialization.Serializable
 typealias TestsSourceVersionInfoList = List<TestsSourceVersionInfo>
 
 /**
- * @property organizationName
- * @property sourceName
+ * @property snapshotInfo
  * @property version
```
kgevorkyan, line 10: Could you please document it

We have `commitId` which is actually `commitHash`, right? And what about `version` from now? What's the meaning of it

nulls: `commitId` is `commitHash` -- an unique id of snapshot (tests)
`version` is human readable format provided by user: `tag`, `branch` or `short commit id`. 
We will support fetching by human readable format and creating a snapshot with `commitId` for internal usage and `version` to show for user to select for running a test

Will extend KDoc

kgevorkyan: But, `version` from `TestSuitesSourceSnapshotKey` is still commit id, right?

0x6675636b796f75676974687562 (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/test/TestsSourceVersionInfo.kt
nulls: > But, `version` from `TestSuitesSourceSnapshotKey` is still commit id, right?

Not **still**: previously it was a version provided by user.
Now we will have two `TestSuitesSourceSnapshotKey`: one detected by preprocessor with `long commit id`, second -- with version provided by user

kgevorkyan: Oh, I see, ok

nulls: a note: it's a temporary till migration is finished.
We will remove `TestSuitesSourceSnapshotKey` and will use two entities:
- `TestsSourceSnapshot`: as a key to store in storage
- `TestsSourceVersion`: as a key\info for user. Will be stored only in database

nulls: > Now we will have two `TestSuitesSourceSnapshotKey`

forgot to copy existed `TestSuites` -- added

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/TestSuitesSourceInternalController.kt
nulls: moved

kgevorkyan (approved):

PR #1730, opened 2023-01-17
Introduced `TestsSourceFetchRequest`

nulls:
### What's done:
* added `TestsSourceFetchRequest`
* removed `TestsSourceSnapshotInfo` and replaced by `TestsSourceSnapshotDto`

It's part of #1717


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/TestSuitesSourceInternalController.kt
```diff
@@ -102,32 +97,40 @@ class TestSuitesSourceInternalController(
...
+                organizationName = source.organization.name,
+                testSuitesSourceName = source.name,
+                version = commitId,
+                creationTime = commitTime,
+            )
+        }
 
-        private fun TestsSourceSnapshotInfo.toKey() = TestSuitesSourceSnapshotKey(
+    companion object {
+        private fun TestsSourceVersionInfo.toKeyForSnapshot() = TestSuitesSourceSnapshotKey(
```
kgevorkyan, line 122: just to clarify, we will remove `TestSuitesSourceSnapshotKey` in future, right? as been said in https://github.com/saveourtool/save-cloud/pull/1727#discussion_r1072081276

nulls: yes, I want to leave it as long as possible during refactoring to move all public api changes on late stage

kgevorkyan (approved):

PR #1731, opened 2023-01-17
Introduced `tests_source_version`

nulls:
### What's done:
* added table `tests_source_version`
* added entity and dto `TestsSourceVersion` and `TestsSourceVersionDto`
* reworked storage to use new entities

It's part of #1717


save-backend/src/main/kotlin/com/saveourtool/save/backend/repository/TestSuiteRepository.kt
```diff
@@ -36,9 +35,8 @@ interface TestSuiteRepository : BaseEntityRepository<TestSuite> {
         "FunctionNaming",
         "FunctionName",
     )
-    fun findAllBySource_Organization_NameAndSource_NameAndVersion(
-        organizationName: String,
-        sourceName: String,
+    fun findAllBySource_IdAndVersion(
```
kgevorkyan, line 38: Why do you start to use `_` in names?

kgevorkyan: As for me, it reads like `find by source, id and version` and not like `find by sourceId and version`, and that is confusing
at least for me

nulls: yes, probably it's redundant here.
Will remove it here.
But I think `_`  is useful when we take entity inside another entity

0x6675636b796f75676974687562 (approved):

PR #1733, opened 2023-01-18
Introduced `TestSuiteVersioned`

nulls:
This extra data class is required to support different versions (alias) for a single TestSuite which will be loaded from snapshot version (commitId -- commit hash code)

### What's done:
* added `TestSuiteVersioned`
* replaced `TestSuiteDto` by `TestSuiteVersioned` in public api

It's part of #1717


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuiteVersioned.kt
```diff
@@ -0,0 +1,48 @@
...
+        fun TestSuiteDto.toVersioned(): TestSuiteVersioned = TestSuiteVersioned(
+            id = this.requiredId(),
+            name = this.name,
+            sourceName = this.source.name,
+            organizationName = this.source.organizationName,
+            isLatestFetchedVersion = this.version == this.source.latestFetchedVersion,
+            description = this.description ?: "",
+            version = this.version,
+            language = this.language ?: "",
+            tags = this.tags?.joinToString(", ") ?: "",
```
kgevorkyan, line 44: I remember, that we had some types of separators in `common`, can we use it here?

0x6675636b796f75676974687562 (approved):

kgevorkyan (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/testsuite/TestSuiteVersioned.kt
nulls: we have `const val DATABASE_DELIMITER = ","`, will add `PRETTY_DELIMITER`

nulls: also, replaced `?: ""` to `.orEmpty()`

PR #1734, opened 2023-01-18
Expose the `clear()`/`replay()` operations via JMX to ease stress testing

0x6675636b796f75676974687562:
This is a part of Flaky Test Detector, see #659.


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestAnalysisService.kt
```diff
@@ -100,42 +151,157 @@ class TestAnalysisService(
...
             .sorted(comparingLong(Execution::requiredId))
-            .mapToLong { execution ->
-                /*
-                 * Process test executions in parallel (slightly faster).
-                 */
-                val testExecutions = testExecutionService.getTestExecutions(execution.requiredId())
-                    .parallelStream()
-
-                val metadataStream = execution.metadata().let { metadata ->
-                    Stream.generate { metadata }
```
kgevorkyan, line 125: don't quite understand, seems, that previously we needed some few more actions
does `updateStatisticsInternal` cover all of them now?

0x6675636b796f75676974687562: It's okay, don't worry.

I just simplified the code -- the metadata is now moved out of the stream pipeline and extracted into a separate variable. So zipping two streams is no longer necessary.

See [`updateStatisticsInternal()`](https://github.com/saveourtool/save-cloud/pull/1734/files#diff-e4c4ff9b5144fdc716635af988bdc8402baf9b84fc3c49447b47876f4597b198R232)

nulls (approved):

PR #1740, opened 2023-01-19
Remove the `else`-branch: the `when`-expression is exhaustive

0x6675636b796f75676974687562:


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/ExecutionScoreUtils.kt
```diff
@@ -57,7 +57,6 @@ fun Int.isValidScore() = this in 0..100
 
 private fun ExecutionDto.calculateScoreForContestMode(scoreType: ScoreType): Double = when (scoreType) {
     ScoreType.F_MEASURE -> calculateFmeasure()
-    else -> TODO("Invalid score type for contest mode!")
```
kgevorkyan, line 60: Why do you prefer such way? Of course, now exhaustive, but if we will expand our data classes, we will need to remember all such places, in aim to update them
while in other hand we will get exceptions, if we will miss something

kgevorkyan (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/ExecutionScoreUtils.kt
0x6675636b796f75676974687562: This way, if we miss something, we'll get a compile error (which is better than a run-time error, isn't it?)

kgevorkyan: I expect warning in this case but

Capture

0x6675636b796f75676974687562: In your case, you have an expression of type `Unit`.
A non-exhaustive `when` evaluates to `Unit`, too

In the code (`ExecutionScoreUtils.kt`), there's an expression of type `Double`

kgevorkyan: Oh, okay, thought that we talking about general case, that each and every `when` should be exhaustive
here, I didn't pay attention into the return type

PR #1745, opened 2023-01-20
Avatar placeholders

### LINKED ISSUE #1562: Avatar placeholders
Right now we had an issue when avatars are unavailable for some reason and we didn't display the placeholder that does not depend on a storage. We need to display the placeholder image if real avatar in not fetched or fetched with an exception

Cheshiriks:
Closes #1562


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -203,6 +208,7 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
                 isEditDisabled = true
                 selfRole = highestRole
                 usersInOrganization = users
+                avatar = organizationLoaded.avatar?.let { "/api/$v1/avatar$it" } ?: run { "img/undraw_profile.svg" }
```
nulls, line 211: while the latest string is in `run`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
```diff
@@ -523,14 +529,14 @@ class OrganizationView : AbstractView<OrganizationProps, OrganizationViewState>(
...
-                        }
-                            ?: run {
-                                "img/company.svg"
-                            }
+                        src = state.avatar
                         height = 100.0
                         width = 100.0
+                        onError = {
+                            setState {
+                                avatar = "img/undraw_image_not_found.png"
```
nulls, line 537: maybe extract this variable to a separate constant?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/OrganizationView.kt
Cheshiriks: Added

PR #1754, opened 2023-01-24
Make sure artifact signing is applied *after* Maven configurations are created

0x6675636b796f75676974687562:
This fixes an issue when only artifacts (w/o *.asc digital signatures) get uploaded to Sonatype when a new release is created.

### DIFF
--- a/gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
+++ b/gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
@@ -10,15 +10,24 @@
 package com.saveourtool.save.buildutils
 
 import io.github.gradlenexus.publishplugin.NexusPublishExtension
+import org.gradle.api.Named
 import org.gradle.api.Project
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.tasks.bundling.Jar
+import org.gradle.internal.logging.text.StyledTextOutput
+import org.gradle.internal.logging.text.StyledTextOutput.Style.Failure
+import org.gradle.internal.logging.text.StyledTextOutput.Style.Success
+import org.gradle.internal.logging.text.StyledTextOutputFactory
 import org.gradle.kotlin.dsl.configure
 import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.getByType
+import org.gradle.kotlin.dsl.support.serviceOf
 import org.gradle.kotlin.dsl.withType
 import org.gradle.plugins.signing.SigningExtension
+
+private fun Project.styledOut(logCategory: String): StyledTextOutput =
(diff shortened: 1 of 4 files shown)


gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/publishing-configuration.gradle.kts
```diff
@@ -35,7 +35,12 @@ run {
 run {
     configurePublications()
 
-    if (hasProperty("signingKey")) {
-        configureSigning()
-    }
+    /*-
```
nulls, line 38: why `save-api` and `test-analysis-core` have another publications? maybe migrate them to common one?

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
```diff
@@ -65,11 +78,25 @@ internal fun Project.configurePublications() {
...
+        configure<SigningExtension> {
+            useInMemoryPgpKeys(property("signingKey") as String?, property("signingPassword") as String?)
+            val publications = extensions.getByType<PublishingExtension>().publications
+            val publicationCount = publications.size
+            val message = "The following $publicationCount publication(s) are getting signed: ${publications.map(Named::getName)}"
+            val style = when (publicationCount) {
+                0 -> Failure
+                else -> Success
+            }
+            styledOut(logCategory = "signing").style(style).println(message)
```
nulls: does `println` much better than `logger`?

0x6675636b796f75676974687562: It's a styled (colored) output. 

The message is important, and I'd prefer it to stand out of the rest of the output.

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/publishing-configuration.gradle.kts
0x6675636b796f75676974687562: `:save-cloud-common` is a multi-platform project, it just happens to have a default Maven publication readily available with no extra configuration.

nulls: discussed offline, will leave it for future

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/PublishingConfiguration.kt
nulls: discussed offline, logger doesn't fit here, so created custom println here

nulls (approved):

nulls:
btw, need to check that license, javadoc (dokkaJar), developers and scm are set in pom.xml for `save-api` and `test-anakysis-core`

PR #1758, opened 2023-01-24
Implement permission check to buttons render on ProjectDemoMenu

### LINKED ISSUE #1676: Implement permission check on `backend`'s `DemoManagerController`

Arrgentum:
## What added:
- added a check on the button rendering depending on the user's role in ProjectDemoMenu.kt
## Issues:
This pull request close frontend (last) path #1676. Close #1676


sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
```diff
@@ -217,9 +232,39 @@ external interface ProjectDemoMenuProps : Props {
      */
     var organizationName: String
 
+    /**
+     * User project role
+     */
+    var projectRole: Role
```
nulls, line 238: I guess we can be replaced by a single role:
```kotlin
val userRole: Role = when {
    props.globalRole.isSuperAdmin() -> Role.SUPER_ADMIN
    props.organizationRole.isHigherOrEqualThan(Role.OWNER) -> props.organizationRole
    else -> props.projectRole
}
```

Arrgentum: I think it is possible, but such back roles will cause a lot of questions in subsequent use.
I'm thinking of redoing it soon with backend requests in the future.

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
sanyavertolet: Take into consideration the fact that the server is in China so requests might take more time then just `when` that Nariman suggested

PR #1759, opened 2023-01-24
[DEMO] upload, get and delete files manually

### LINKED ISSUE #1645: [DEMO] entity fields refactoring
As I see, for now we have at least **2** different names for single entity, it's a `toolName` and `projectName`. While in terms of Github it's called in third way - `repository`

May be we can open an issue and unify this terms one time?

That's why I asked in other conversation about relation to the `github`, because `tool` associated to me with some executable and not `repository`

_Originally posted by @kgevorkyan in https://github.com/saveourtool/save-cloud/pull/1644#discussion_r1053449820_

sanyavertolet:
### What's done:
* Fixed `/upload-file` endpoint
* Added `/delete-file` endpoint both on backend (for permission check) and on demo
* Added `/list-file` endpoint both on backend (for permission check) and on demo
* Renamed fields of demo entities to match common practice (`organizationName` and `projectName`) (which closes #1645)


0x6675636b796f75676974687562 (approved):

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/ManagementController.kt
```diff
@@ -87,6 +94,64 @@ class ManagementController(
         .flatMap {
             demoService.loadFileToStorage(organizationName, projectName, version, file)
         }
+        .map {
+            FileDto(
```
nulls, line 98: `FileDto` supposes to store this key in database and use it as key of storage.
why do you need `FileDto` here? you have `ToolKey`

save-demo/src/main/kotlin/com/saveourtool/save/demo/storage/ToolStorage.kt
```diff
@@ -22,5 +22,5 @@ class ToolStorage(
         return ToolKey(ownerName, toolName, vcsTagName, executableName)
     }
 
-    override fun buildPathToContent(rootDir: Path, key: ToolKey): Path = rootDir / key.ownerName / key.toolName / key.vcsTagName / key.executableName
+    override fun buildPathToContent(rootDir: Path, key: ToolKey): Path = rootDir / key.organizationName / key.projectName / key.version / key.fileName
```
nulls, line 25: suggestion to store keys in database and use primary ID as path. It avoids problem with renaming and escaping a spec symbols in file name

sanyavertolet: Will be done when the frontend part of file uploading is done

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/ManagementController.kt
sanyavertolet: For now with use `ToolKey`, will be changed in PR with db-storage integration

sanyavertolet: Oh will leave as is as far as `ToolKey` is `save-demo` specific

PR #1767, opened 2023-01-25
[DEMO] make GitHub download optional

### LINKED ISSUE #1766: [DEMO] Make GitHub download optional

sanyavertolet:
This PR closes #1766

### What's done:
 * Made GitHub download optional - initialized only if `githubProjectCoordinates` are null
 * Changed hibernate logging level on save-demo to `WARN`
 * Added validation on `save-demo` creation

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
@@ -28,10 +28,7 @@ import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
 import org.springframework.web.bind.annotation.*
 import org.springframework.web.reactive.function.BodyInserters
-import org.springframework.web.reactive.function.client.WebClient
-import org.springframework.web.reactive.function.client.body
-import org.springframework.web.reactive.function.client.bodyToFlux
-import org.springframework.web.reactive.function.client.bodyToMono
+import org.springframework.web.reactive.function.client.*
 import org.springframework.web.server.ResponseStatusException
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
@@ -68,7 +65,7 @@ class DemoManagerController(
     @ApiResponse(responseCode = "200", description = "Successfully added demo.")
     @ApiResponse(responseCode = "403", description = "Not enough permission for accessing given project.")
     @ApiResponse(responseCode = "404", description = "Could not find project in organization.")
-    @ApiResponse(responseCode = "409", description = "Please provide github repository")
+    @ApiResponse(responseCode = "409", description = "Invalid demo creation request.")
     fun addDemo(
         @PathVariable organizationName: String,
         @PathVariable projectName: String,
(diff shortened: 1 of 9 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -97,16 +94,20 @@ class DemoManagerController(
...
             webClientDemo.post()
                 .uri("/demo/internal/add-tool")
                 .contentType(MediaType.APPLICATION_JSON)
                 .bodyValue(demoDto)
                 .retrieve()
+                .onStatus({ it == HttpStatus.CONFLICT }) {
+                    Mono.error(
+                        ResponseStatusException(
+                            HttpStatus.CONFLICT,
+                            "Demo creation request is invalid: fill project coordinates, run command and file name.",
```
kgevorkyan, line 107: Shouldn't it be `400 bad request`?

kgevorkyan: From docs

409 Conflict
Indicates that the request could not be processed because of conflict in the current state of the resource, **such as an between multiple simultaneous updates**

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoDto.kt
```diff
@@ -24,6 +24,12 @@ data class DemoDto(
     val configName: String? = null,
     val githubProjectCoordinates: ProjectCoordinates? = null,
 ) {
+    /**
+     * @return true of [DemoDto] is valid, false otherwise
+     */
+    @Suppress("FUNCTION_BOOLEAN_PREFIX")
+    fun validate(): Boolean = !projectCoordinates.consideredBlank() && runCommand.isNotBlank() && fileName.isNotBlank()
```
kgevorkyan, line 31: isValid? and remove `Suppress`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
sanyavertolet: Oh got it. We used to return `Conflict` when data is invalid.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/domain/ProjectCoordinates.kt
```diff
@@ -17,9 +17,9 @@ data class ProjectCoordinates(
     override fun toString() = "$organizationName/$projectName"
 
     /**
-     * @return true if both [organizationName] and [projectName] are empty, false otherwise
+     * @return true if both [organizationName] and [projectName] are blank, false otherwise
      */
-    fun consideredEmpty() = organizationName.isEmpty() && projectName.isEmpty()
+    fun consideredBlank() = organizationName.isBlank() || projectName.isBlank()
```
kgevorkyan, line 22: same

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/demo/DemoDto.kt
sanyavertolet: Problem with Jackson serialization - isValid will be serialized (dnw)

0x6675636b796f75676974687562: > Problem with Jackson serialization

1. You can register a mix-in class and add a `@JsonIgnore`-annotated property &mdash; see `TestMetricsMixin`.
2. You can annotate a properly-named function with an unusual `@JvmName` and have Jackson effectively ignoring it, even w/o any mix-in.

PR #1775, opened 2023-01-26
fix flaky test

icemachined:


save-orchestrator-common/src/test/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManagerTest.kt
```diff
@@ -53,7 +53,7 @@ class DockerContainerManagerTest {
         val testFile = createTempFile().toFile()
         testFile.writeText("wow such testing")
         dockerAgentRunner.createAndStart(
-            executionId = 42,
+            executionId = 43,
```
kgevorkyan, line 56: how this change fix `flacky`?

icemachined: 42 is used in AgentsControllerTest, it's not always managed to shutdown container so the next test can't use it and failed.

sanyavertolet: suggestion to generate such ids instead of using it hardcoded

sanyavertolet (approved):

save-orchestrator-common/src/test/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManagerTest.kt
kgevorkyan: but as I remember, we used `42` more, than in one (current) place, isn't it? shouldn't we change other places too?

kgevorkyan (approved):

save-orchestrator-common/src/test/kotlin/com/saveourtool/save/orchestrator/docker/DockerContainerManagerTest.kt
icemachined: i found only 2 files

PR #1778, opened 2023-01-27
Wrap blocking operations into `blockingToMono {}`

0x6675636b796f75676974687562:
`@Blocking`/`@NonBlocking` annotations have been added as necessary.

### DIFF
--- a/save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestDiscoveringService.kt
+++ b/save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestDiscoveringService.kt
@@ -3,6 +3,7 @@ package com.saveourtool.save.preprocessor.service
 import com.saveourtool.save.core.config.TestConfig
 import com.saveourtool.save.core.files.ConfigDetector
 import com.saveourtool.save.core.plugin.GeneralConfig
+import com.saveourtool.save.core.plugin.PluginException
 import com.saveourtool.save.core.utils.buildActivePlugins
 import com.saveourtool.save.core.utils.processInPlace
 import com.saveourtool.save.entities.TestSuite
@@ -13,6 +14,7 @@ import com.saveourtool.save.test.TestsSourceSnapshotDto
 import com.saveourtool.save.test.collectPluginNames
 import com.saveourtool.save.testsuite.TestSuiteDto
 import com.saveourtool.save.utils.EmptyResponse
+import com.saveourtool.save.utils.blockingToMono
 import com.saveourtool.save.utils.debug
 import com.saveourtool.save.utils.info
 import com.saveourtool.save.utils.thenJust
@@ -20,6 +22,8 @@ import com.saveourtool.save.utils.thenJust
 import okio.FileSystem
 import okio.Path
 import okio.Path.Companion.toPath
+import org.jetbrains.annotations.Blocking
+import org.jetbrains.annotations.NonBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import reactor.core.publisher.Flux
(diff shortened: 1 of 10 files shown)


sanyavertolet (approved):

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestDiscoveringService.kt
```diff
@@ -52,18 +56,22 @@ class TestDiscoveringService(
         log.info { "Starting to save new test suites for root test config in $repositoryPath" }
         return Mono.just(repositoryPath)
             .map { it.resolve(testRootPath) }
-            .map { getRootTestConfig(it.absolutePathString()) }
+            .flatMap { blockingToMono { getRootTestConfig(it.absolutePathString()) } }
```
nulls, line 59: Combine previous steps to a single map:
```suggestion
            blockingToMono { getRootTestConfig(repositoryPath.resolve(testRootPath).absolutePathString()) }
```

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestDiscoveringService.kt
```diff
@@ -52,18 +56,22 @@ class TestDiscoveringService(
...
-                    rootTestConfig,
-                    sourceSnapshot,
-                ).toMono()
+                {
+                    log.info { "Starting to discover test suites for root test config ${rootTestConfig.location}" }
+                    discoverAllTestSuites(
+                        rootTestConfig,
+                        sourceSnapshot,
+                    )
+                }.toMono()
```
nulls, line 67: I think here should be `blockingMono` since `discoverAllTestSuites` operates with FS

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestsPreprocessorToBackendBridge.kt
```diff
@@ -58,45 +61,49 @@ class TestsPreprocessorToBackendBridge(
...
      * @param sourceId
      * @param commitId
      * @return [TestsSourceSnapshotDto] found by provided values
      */
+    @NonBlocking
     fun findTestsSourceSnapshot(sourceId: Long, commitId: String): Mono<TestsSourceSnapshotDto> = webClientBackend.get()
         .uri("/test-suites-sources/find-snapshot?sourceId={sourceId}&commitId={commitId}", sourceId, commitId)
         .retrieve()
-        .bodyToMono()
+        .blockingBodyToMono()
```
nulls, line 75: why blockingBodyToMono?

nulls (changes requested):

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestsPreprocessorToBackendBridge.kt
0x6675636b796f75676974687562: A network operation (from the _preprocessor_ to the _backend_) is a blocking I/O operation, just like any other I/O operation (e.g.: a file system access).

This change offloads network requests the preprocessor **initiates** (not the ones it **receives**) from the `reactor-http-*` pool (which serves the **received** requests) to the `bounded-elastic-*` pool.

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestDiscoveringService.kt
0x6675636b796f75676974687562: As you can see, `discoverAllTestSuites()` is annotated with `@NonBlocking`.

Indeed, if I'm not mistaken, it only processes TOML content once it's already read into memory.

0x6675636b796f75676974687562: There're three `@Blocking` operations, however: getRootTestConfig(), `getAllTests()`, and `discoverAllTestsIntoMap()`.

save-preprocessor/src/main/kotlin/com/saveourtool/save/preprocessor/service/TestsPreprocessorToBackendBridge.kt
nulls: it means that all network operation should be blockingBodyToMono?

nulls (approved):

PR #1780, opened 2023-01-27
use prebuild jep distro

icemachined:


nulls:
does it resolve the issue with using latest `paketo-buildpacks`?

nulls (approved):

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/SystemUtils.kt
```diff
@@ -0,0 +1,26 @@
+/**
+ * Utility functions for System
+ */
+
+package com.saveourtool.save.buildutils
+
+private fun String.isWindows() = startsWith("Windows", ignoreCase = true)
```
0x6675636b796f75676974687562, line 7: It's better to use Gradle's `DefaultNativePlatform` and `DefaultOperatingSystem` here.

PR #1808, opened 2023-02-02
S3 Operations

### LINKED ISSUE #1798: Add a custom `Execution` and `Scheduler` for S3 operations

nulls:
### What's done
* create a custom `Execution`
* create a custom `Scheduler` from custom `Executor`
* use this `Executor` in http client in AWS S3 SDK and `Scheduler` for further processing
* Extracted this logic to common interface `S3Operations`

It closes #1798


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/S3OperationsProperties.kt
```diff
@@ -0,0 +1,66 @@
...
+     * @property maxPoolSize
+     * @property queueSize
+     * @property ttl
+     */
+    data class AsyncProperties(
+        val minPoolSize: Int = Schedulers.DEFAULT_POOL_SIZE,
+        val maxPoolSize: Int = Schedulers.DEFAULT_BOUNDED_ELASTIC_SIZE,
+        val queueSize: Int = Schedulers.DEFAULT_BOUNDED_ELASTIC_QUEUESIZE,
+        val ttl: Duration = 60.seconds.toJavaDuration(),
+        // reactor.core.scheduler.BoundedElasticScheduler.DEFAULT_TTL_SECONDS
```
kgevorkyan, line 64: ?

kgevorkyan (approved):

0x6675636b796f75676974687562 (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/S3OperationsProperties.kt
nulls: it's comment to `ttl` where I took the default value. `diktatFix` moved it to next line. will move on 1 line up

nulls: will do it in next PR

kgevorkyan: hm, we should catch such situations in diktat, I didn't remember how, but we definitely have logic for inline comments
I suppose, that we shouldn't trigger on such cases. But may be it depends on configuration

nulls: ok, strange then, but will move to KDoc anyway. It's more proper place for it

nulls: fixed here d9545ee84223c1a4a3fe8973dc6addc01f12fc80

Issue #1096, opened 2022-08-24
Test suite source validation


petertrr:
After test suite source is fetched, we should store and display validations, including among others:
* Whether source contains suites with plugins that have zero tests
* Whether source contains suites with wildcard mode (not handled well by save-cloud currently?)

This includes calculating these validations, storing them in the storage and support them on the frontend

---

Tasks, as discussed with @nulls:
 * Front-end
   * [X] #1810
   * [ ] Display the current fetch status ("Fetching...", "Completed", "Errors encountered", "Time-out", "Preprocessor unavailable") when (re-)fetching
   * [ ] Display the last fetch status ("Completed", "Errors encountered") when viewing tests (the _Tests_ tab within an organization), using data from the `tests_source_version` and `tests_source_snapshot` tables.
 * Back-end
   * TBD
 * Preprocessor
   * [X] #1778
   * [x] #1955 
   * [x] #1956 
   * [ ] Refactor `TestSuitesPreprocessorController#doFetchTests()` and the underlying calls
      * [ ] The snapshot should be always saved (`TestPreprocessorToBackendBridge#saveTestsSuiteSourceSnapshot()`), even in case of validation errors. This is necessary, so that a malformed snapshot is never re-fetched again.
      * [ ] The `tests_source_snapshot` table, along with the corresponding DTO, should be enriched with `isSuccessful` and `errorMessage` fields
      * [ ] `TestDiscoveringService#detectAndSaveAllTestSuitesAndTests()` should perform the following operations, in order:
         1. detect test suites and tests
         2. validate test suites and tests
         3. update the saved snapshot with status and error messages, if any
         4. save test suites and tests, if the validation succeeded

akuleshov7:
Andrey did a lot of related things, probably we will open a new issue, when needed

PR #1810, opened 2023-02-02
Implement the `useEventStream()` (SSE, HTTP GET only) and `useNewlineDelimitedJson()` React hooks

0x6675636b796f75676974687562:
### What's done:

 - This change allows live event streaming from the back-end (`Flux<T>` or
   `ParallelFlux<T>`) to the font-end, using either `text/event-stream`
   (arbitrary text data, incl. JSON) or `application/x-ndjson` (JSON only).
 - On the server side, _Spring_ has support for both content types.
 - This is a part of https://github.com/saveourtool/save-cloud/issues/1096.


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/CustomHooks.kt
```diff
@@ -152,3 +161,153 @@ fun useBackground(colorStyle: Style) {
...
+ *   server closes the connection.
+ *   Allowed to change the state of components.
+ * @param onError invoked when the event source reports an error.
+ *   Allowed to change the state of components.
+ * @param onEvent invoked when a new event arrives.
+ *   Allowed to change the state of components.
+ * @return a lambda that triggers the callback.
+ * @see useEventStream
+ */
+internal fun useNewlineDelimitedJson(
```
nulls, line 282: maybe name it as `Jsonl`:
```suggestion
internal fun useJsonl(
```

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/HttpUtils.kt
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+/**
+ * @param timings the server-side timings.
+ * @param lazyResponse the lazy HTTP response.
+ * @return [lazyResponse] wrapped with HTTP headers.
+ */
+@OptIn(ExperimentalContracts::class)
+fun <T : Any> withHttpHeaders(
```
sanyavertolet, line 90: Shouldn't `with*` function be an extension function?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
```diff
@@ -531,6 +542,122 @@ suspend fun WithRequestStatusContext.loadingHandler(request: suspend () -> Respo
...
+}
+
+/**
+ * Decodes this byte flow into a flow of strings, assuming UTF-8 encoding.
+ *
+ * Malformed byte sequences are replaced with `\uFFFD`.
+ *
+ * @see ByteArray.decodeToString
+ */
+fun Flow<Byte>.decodeToString(): Flow<String> =
```
nulls, line 616: all these util methods are used only here -- maybe mark them as private?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
```diff
@@ -531,6 +542,122 @@ suspend fun WithRequestStatusContext.loadingHandler(request: suspend () -> Respo
...
+         * available.
+         */
+        while (true) {
+            @Suppress(
+                "GENERIC_VARIABLE_WRONG_DECLARATION",
+                "TYPE_ALIAS",
+            )
+            val resultAsync = reader
+                .read()
+                .unsafeCast<Promise<ReadableStreamDefaultReadValueResult<Uint8Array>>>()
```
sanyavertolet, line 576: Maybe some alias for that?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/CustomHooks.kt
0x6675636b796f75676974687562: Addressed with e5f501d371cbcad8f71880d4780555fc8fb2e1d2

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
0x6675636b796f75676974687562: Addressed with 46bfea9aaa9d4370f5e294e80980d97b0d504232

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/HttpUtils.kt
0x6675636b796f75676974687562: No, as it, as a side-effect, measures the execution time of the lambda it accepts and sets the `Server-Timing` header.

So it can be only made an extension on a lambda: `fun (() -> Response<T>).withHttpHeaders()`, which makes the function impractical.

sanyavertolet: oh ok got it

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/RequestUtils.kt
```diff
@@ -682,13 +689,10 @@ private suspend fun Response.inputStream(): Flow<Byte> {
          * available.
          */
         while (true) {
-            @Suppress(
-                "GENERIC_VARIABLE_WRONG_DECLARATION",
-                "TYPE_ALIAS",
-            )
+            @Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
```
nulls, line 692:
```suggestion
            val resultAsync: ResultAsync<Uint8Array> = reader
```
does it resolve suppression?

0x6675636b796f75676974687562: Yes it does, thanks.

Addressed with 7fd1bd570113cb3581186b6bcf497f088ac1f418

PR #1811, opened 2023-02-02
Toml instead of Properties

sanyavertolet:
### What's done:
 * Moved `fs` to `commonMain` source set of `save-cloud-common` as `expected val` with `actual`s on all the platforms
 * Added `nativeMain` source set to `save-cloud-common`
 * Deleted `.properties` parsing logic
 * Moved from `Properties` configuration to `Toml` configuration (for agents)


0x6675636b796f75676974687562 (approved):

save-cloud-common/src/jsMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
```diff
@@ -0,0 +1,15 @@
+/**
+ * js implementation for FileUtils - Not implemented
+ */
+
+package com.saveourtool.save.utils
+
+import okio.FileSystem
+import okio.Path
+
+const val NOT_IMPLEMENTED_ON_JS = "Cannot be used in js."
```
nulls, line 10: it cannot be used in legacy JS

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
```diff
@@ -97,12 +107,16 @@ fun Path.pathNamesTill(stop: Path): List<String> = generateSequence(this, Path::
...
-                "The path is not absolute: $this"
-            }
-        }
+fun Path.requireIsAbsolute(): Path = apply {
+    require(isAbsolute) {
+        "The path is not absolute: $this"
+    }
+}
+
+actual inline fun <reified C : Any> parseConfig(configPath: okio.Path): C {
```
nulls, line 116: maybe move to `commonNonJsMain`?

save-cloud-common/build.gradle.kts
```diff
@@ -81,6 +87,17 @@ kotlin {
                 runtimeOnly(libs.junit.jupiter.engine)
             }
         }
+
+        val linuxX64Main by getting
+        val macosX64Main by getting
+        val mingwX64Main by getting
+
+        val nativeMain by creating {
```
nulls, line 95: why do we need `nativeMain` here?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
nulls: discussed offline, that it doesn't resolve the code duplication for different targets

save-cloud-common/src/jsMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
sanyavertolet: It cannot be used in js as well as js legacy, can it?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
sanyavertolet: Discussed offline - decided to leave as is

save-cloud-common/build.gradle.kts
sanyavertolet: There is an implementation for `linuxX64`, `macosX64` and `mingwX64`

nulls: discussed offline: gradle has problem with long hierarchy

save-cloud-common/src/jsMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
nulls: yes, you're right. I confused with serialization

nulls (approved):

PR #1831, opened 2023-02-06
Supported download on save-agent required files from S3

### LINKED ISSUE #1830: [save-agent] Download files directly from S3

nulls:
### What's done
* supported pre-signed url
* migrated to GET on save-agent
* removed AbstractFileBasedStorage

It closes #1830


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/DefaultS3Operations.kt
```diff
@@ -25,8 +33,13 @@ import java.util.concurrent.*
  */
 class DefaultS3Operations(
     properties: S3OperationsProperties,
-) : S3Operations {
+) : S3Operations, Closeable {
```
0x6675636b796f75676974687562, line 36: Probably, it's better to use `AutoCloseable` here (almost the same signature, more modern API).

nulls: sure

sanyavertolet (approved):

0x6675636b796f75676974687562 (approved):

PR #1836, opened 2023-02-07
TestSuiteSelector search mode fix

### LINKED ISSUE #1834: Search in the selector view does not work
image

sanyavertolet:
This PR closes #1834

### What's done:
 * Fixed `/filtered` endpoint - @nulls has broken it by adding new field to `TestSuite` forgetting about `ExampleMatcher`'s `ignorePaths`

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
@@ -91,7 +91,7 @@ class TestSuitesService(
                 .withMatcher("name", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
                 .withMatcher("language", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
                 .withMatcher("tags", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
-                .withIgnorePaths("description", "source", "version", "dateAdded", "plugins", "sourceSnapshot")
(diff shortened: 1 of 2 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
```diff
@@ -91,7 +91,7 @@ class TestSuitesService(
                 .withMatcher("name", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
                 .withMatcher("language", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
                 .withMatcher("tags", ExampleMatcher.GenericPropertyMatchers.contains().ignoreCase())
-                .withIgnorePaths("description", "source", "version", "dateAdded", "plugins")
+                .withIgnorePaths("description", "source", "version", "dateAdded", "plugins", "sourceSnapshot")
```
nulls, line 94: `source` can be removed and please add a link/note about this runtime dependency to entity class

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TestSuitesService.kt
sanyavertolet: Replaced `source` with `sourceSnapshot` as well as added reminder to `TestSuite`'s KDoc

nulls (approved):

PR #1838, opened 2023-02-07
Upload save-demo-agent.kexe

sanyavertolet:
### What's done:
 * Supported `save-demo-agent.kexe` upload in `deploy_images.yml`

### DIFF
--- a/.github/workflows/deploy_images.yml
+++ b/.github/workflows/deploy_images.yml
@@ -156,6 +156,12 @@ jobs:
           name: save-agent-distribution
           path: 'save-cloud/save-agent/build/libs/save-agent-*-distribution.jar'
           retention-days: 5
+      - name: Upload save-demo-agent.kexe
+        uses: actions/upload-artifact@v3
+        with:
+          name: save-demo-agent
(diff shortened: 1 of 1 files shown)


kgevorkyan (approved):

.github/workflows/deploy_images.yml
```diff
@@ -156,6 +156,12 @@ jobs:
           name: save-agent-distribution
           path: 'save-cloud/save-agent/build/libs/save-agent-*-distribution.jar'
           retention-days: 5
+      - name: Upload save-demo-agent.kexe
+        uses: actions/upload-artifact@v3
+        with:
+          name: save-demo-agent
+          path: 'save-cloud/save-demo-agent/build/bin/linuxX64/releaseExecutable/save-demo-agent.kexe'
```
nulls, line 163: why `kexe`, but not `distribution.jar` as for save-agent? we will add `save-demo-agent` as dependency to `save-demo`

sanyavertolet: Yep, in next PR I will add Gradle task for that (as well as return agent.toml with server port _maybe_)

PR #1850, opened 2023-02-09
save-demo-agent-distribution

sanyavertolet:
### What's done:
 * Added `agent.toml` and `save-demo-agent.kexe` to `save-demo-agent-distribution.jar`
 * Implemented simple config parsing logic for `save-demo-agent`
 * Fixed workflow file in order to upload `save-demo-agent-distribution.jar`


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
```diff
@@ -43,3 +44,16 @@ expect inline fun <reified C : Any> parseConfig(configPath: Path): C
...
+ * Notice that [C] should be serializable
+ *
+ * @param defaultConfig config that should be set if no config was found ([FileNotFoundException])
+ * @param configName name of a toml config file, agent.toml by default
+ * @return [C] filled with configuration information
+ */
+inline fun <reified C : Any> parseConfigOrDefault(
+    defaultConfig: C,
+    configName: String = "agent.toml",
+): C = try { parseConfig(configName) } catch (e: FileNotFoundException) { defaultConfig }
```
0x6675636b796f75676974687562, line 59: I would still log a message that 

> Config file %s not found, falling back to the default config...

0x6675636b796f75676974687562 (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
sanyavertolet: Sure, will add

save-demo-agent/build.gradle.kts
```diff
@@ -58,4 +59,18 @@ kotlin {
             }
         }
     }
+
+    @Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
```
kgevorkyan, line 63: still need this suppress? you already provide a type `TaskProvider<KotlinNativeLink>`?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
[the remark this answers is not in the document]
sanyavertolet: Oh ok

save-demo-agent/build.gradle.kts
sanyavertolet: As I understand, due to kts features, we still need to pass a type to `tasks.named`, but diktat fixes it into 
```kotlin
val linkTask: TaskProvider<KotlinNativeLink> = tasks.named("linkReleaseExecutableLinuxX64")
```
which is wrong.

sanyavertolet: So yes, we still need it

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/FileUtils.kt
sanyavertolet: Done

nulls (approved):

PR #1853, opened 2023-02-09
save-demo-agent-configuration.gradle.kts

sanyavertolet:
### What's done:
 * Added `save-demo-agent-configuration.gradle.kts`
 * Added runtime dependency to `save-demo` (`save-demo` now depends on `save-demo-agent`)

### DIFF
--- a/gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
@@ -18,7 +18,7 @@ plugins {
 val downloadSaveDemoAgentDistroTaskProvider: TaskProvider<Download> = tasks.register<Download>("downloadSaveDemoAgentDistro") {
     enabled = findProperty("saveDemoAgentDistroFilepath") != null
 
-    val saveDemoAgentDistroFilepath = findProperty("saveDemoAgentDistroFilepath")?.toString() ?: "file:\\\\not-found"
+    val saveDemoAgentDistroFilepath = findProperty("saveDemoAgentDistroFilepath")?.toString() ?: "file://not-found"
     src { saveDemoAgentDistroFilepath }
     dest { "$buildDir/demoAgentDistro/${File(saveDemoAgentDistroFilepath).name}" }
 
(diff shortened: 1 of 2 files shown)


gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+plugins {
+    kotlin("jvm")
+    id("de.undercouch.download")
+}
+
+@Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
+val downloadSaveDemoAgentDistroTaskProvider: TaskProvider<Download> = tasks.register<Download>("downloadSaveDemoAgentDistro") {
+    enabled = findProperty("saveDemoAgentDistroFilepath") != null
+
+    val saveDemoAgentDistroFilepath = findProperty("saveDemoAgentDistroFilepath")?.toString() ?: "file:\\\\not-found"
```
0x6675636b796f75676974687562, line 21: Is the fallback value a `file://` URL?

```suggestion
    val saveDemoAgentDistroFilepath = findProperty("saveDemoAgentDistroFilepath")?.toString() ?: "file://not-found"
```

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+    src { saveDemoAgentDistroFilepath }
+    dest { "$buildDir/demoAgentDistro/${File(saveDemoAgentDistroFilepath).name}" }
```
0x6675636b796f75676974687562, line 23: Can you please use `Path` instead?

0x6675636b796f75676974687562 (approved):

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
sanyavertolet: yep, thx

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-demo-agent-configuration.gradle.kts
sanyavertolet: Done. Now using
```Kotlin
Path.of(saveDemoAgentDistroFilepath).fileName
```

PR #1854, opened 2023-02-09
s3-deployment.yaml

sanyavertolet:
### What's done:
 * Added `s3-deployment.yaml` that adds `s3-storage` configuration for `dev` profile if and only if `.Values.s3.external` is `false`


save-cloud-charts/save-cloud/templates/s3-deployment.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if not .Values.s3.external -}}
+
+  {{- if eq .Values.profile "dev" -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: s3-secrets
+stringData:
+  s3-storage.endpoint: "http://host.minikube.internal:9000"
```
nulls, line 9: need to add also chart to create s3 storage 
also, I think we will give some name to s3 service and should use it instead of `"http://host.minikube.internal:9000"`

nulls (changes requested):

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Oh ok, will do it

sanyavertolet: Done

0x6675636b796f75676974687562 (approved):

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if not .Values.s3.external -}}
+
+  {{- if eq .Values.profile "dev" -}}
```
nulls, line 3: I'm not sure we need this `if` here: if s3 is internal, we should create a secret and service, when it's external -- secret is created in values-cce

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+    spec:
+      containers:
+        - image: minio/minio:latest
+          name: minio
+          ports:
+            - containerPort: 9000
+              name: minio
+          env:
+            - name: MINIO_ROOT_USER
+              value: "admin"
```
nulls, line 40: suggestion to extract it to values and reuse in secret and here. actually I think we can use secret to set env variable for deploment

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if not .Values.s3.external -}}
+
+  {{- if eq .Values.profile "dev" -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: s3-secrets
+stringData:
+  s3-storage.endpoint: "http://s3:9000"
```
nulls, line 9: this port `9000` can be also moved to values.yaml and re-used

save-cloud-charts/save-cloud/templates/s3-service.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.s3.external -}}
```
nulls, line 1: I don't think we need a service to get access to S3. mysql requires it since we don't have access to its IP outside and create service to connect to mysql from local machine by port forwarding

nulls (changes requested):

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Makes sense, will remove it

save-cloud-charts/save-cloud/templates/s3-service.yaml
sanyavertolet: I guess port forwarding in this case might be convenient for debugging

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Ok will try to do it

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Let's try

save-cloud-charts/save-cloud/templates/s3-service.yaml
nulls: we won't be able to debug external service and we can connect directly to this service

sanyavertolet: Yep, I got my problem, will configure it to deal with minio locally

sanyavertolet:
@nulls please notice that `minio` password was changed from `12345678` to `adminadmin` due to wrong go templates stuff: `12345678` was treated like `1.2345678e+07`

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Done

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Extracted such data to `values-minikube.yaml`

save-cloud-charts/save-cloud/templates/s3-deployment.yaml
sanyavertolet: Done

nulls (approved):

PR #1860, opened 2023-02-10
Added view for all demo projects

nulls:


save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
```diff
@@ -25,20 +29,41 @@ import kotlin.io.path.createTempDirectory
...
+        demoService.getAllDemos().map { it.toDto() }
+    }
+
+    /**
+     * @param organizationName
+     * @param projectName
+     * @return configName or empty response
+     */
+    @GetMapping("/{organizationName}/{projectName}/config-name")
+    fun configName(
```
sanyavertolet, line 53: Can be replaced with `ManagementControllers`'s getInfo()

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
```diff
@@ -124,16 +139,16 @@ private fun ChildrenBuilder.displayAlertWithWarnings(result: DemoResult, flushWa
     "LongMethod",
     "TYPE_ALIAS"
 )
-private fun diktatDemoComponent() = FC<DiktatDemoComponentProps> { props ->
-    val (diktatRunRequest, setDiktatRunRequest) = useState(DemoRunRequest.diktatDemoRunRequest)
+private fun demoRunComponent() = FC<DemoRunComponentProps> { props ->
```
sanyavertolet, line 142:
```suggestion
val demoRunComponent: FC<DemoRunComponentProps> = FC { props ->
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
```diff
@@ -181,52 +196,54 @@ private fun diktatDemoComponent() = FC<DiktatDemoComponentProps> { props ->
...
+                                ?.let { file ->
+                                    reader.readAsText(file, "UTF-8")
+                                }
+                        }
                     }
+                    fontAwesomeIcon(icon = faUpload)
+                    val uploadOrReplace = if (demoRunRequest.config.isNullOrEmpty()) {
+                        "Upload"
+                    } else {
+                        "Replace"
```
sanyavertolet, line 243: This doesn't make sense for now

akuleshov7:
View on the FE from Nariman, seriously? :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
nulls: it doesn't make sense for Diktat only, `com.saveourtool.save.demo.DemoRunRequest` supports `config` == null

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
nulls: will make it as public to reuse it for `config-name`

nulls: actually, will leave it for you. will use endpoint from backend `/api/v1/demo`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
nulls: done

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
sanyavertolet: Well, let's leave as is. Not that important

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
sanyavertolet: Well, I mean I am not sure if the label works correctly. Seems that with diktat it didn't work as there was default config in data class. For now it should work fine probably

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
```diff
@@ -37,9 +37,7 @@ import web.html.InputType
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 
-val diktatDemoComponent = diktatDemoComponent()
-
-private val diktatDemoDefaultCode = """
+private val defaultCode = """
```
sanyavertolet, line 40: Not sure if we want to display this code as default. Probably should let demo owner upload file with name `inputFileName` in order to set default code. (should be done later)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
```diff
@@ -60,6 +58,21 @@ external interface DiktatDemoComponentProps : Props {
      * Mode for Ace Editor
      */
     var selectedMode: Languages
+
+    /**
+     * An initial value of [DemoRunRequest]
+     */
+    var emptyDemoRunRequest: DemoRunRequest
```
sanyavertolet, line 65: Shall we pass an `emptyDemoRunRequest` with `props`? Is it done in order to not to pass `organizationName` and `projectName`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
```diff
@@ -50,13 +52,40 @@ data class TopBarLink(
...
+
 /**
  * Displays the static links that do not depend on the url.
  */
 @Suppress("MAGIC_NUMBER", "LongMethod", "TOO_LONG_FUNCTION")
 private fun topBarLinks() = FC<TopBarLinksProps> { props ->
     val navigate = useNavigate()
     var isDemoDropdownActive by useState(false)
+    val deactivateDropdown = { isDemoDropdownActive = false }
+    var demos by useState(listOf<DemoDto>())
```
sanyavertolet, line 77: You should not want to use `useState` for lists with `by` initialization as setting state for lists usually means passing lambda to `StateSetter`, which as I know is impossible with delegation

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
```diff
@@ -79,19 +108,26 @@ private fun topBarLinks() = FC<TopBarLinksProps> { props ->
...
+                    navigate,
+                    deactivateDropdown,
+                )
+                demoDropdownEntry(
+                    "CPG",
+                    "/${FrontendRoutes.DEMO.path}/cpg",
+                    navigate,
+                    deactivateDropdown,
+                )
+                demos.forEach { demo ->
```
sanyavertolet, line 124:
```suggestion
                demos.take(3).forEach { demo ->
```
Just to look nice for now

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
```diff
@@ -0,0 +1,79 @@
...
+import react.router.useParams
+
+private val backgroundCard = cardComponent(hasBg = true, isPaddingBottomNull = true)
+
+val demoView = demoView()
+
+private fun demoView(): VFC = VFC {
+    useBackground(Style.WHITE)
+    val (selectedTheme, setSelectedTheme) = useState(AceThemes.preferredTheme)
+    val params = useParams()
```
sanyavertolet, line 28: Why do we need to use `useParams` with `VFC` instead of passing params from `react-router` into `FC` (as usual)? Is it because of diktat demo "legacy" support?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
nulls: it's done because `emptyDemoRunRequest` is not empty for diktat for now

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
nulls: I'm not sure -- I think we should leave it as empty (when it's be ready for production) or have a map where key is language

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
nulls: thanks for a note. Here `demos` is set a fully without manipulation with list, but will rewrite to `pair` to avoid possible problems in future

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
sanyavertolet: And we also should think of cookies, just as you suggested

sanyavertolet: > or have a map where key is language
The point is that we do not what language is connected with the demo. The only thing that could tell us about the language is `inputFileName` extension which is not that reliable

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
nulls: nope, `BasicRouting.kt` setups routing from url to some view. And view paths (as in Spring) are parsed by `useParams`.
It's the same as `withRouter` works inside or `com.saveourtool.save.frontend.components.views.TestExecutionDetailsViewKt#testExecutionDetailsView`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
nulls: are you sure, how we will look all of them? I guess it should be done in another way.
will add it for now with a `TODO` comment

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
sanyavertolet: Oh ok nice crutch for now

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
nulls: > > or have a map where key is language
> 
> The point is that we do not what language is connected with the demo. The only thing that could tell us about the language is `inputFileName` extension which is not that reliable

then -- empty by default

nulls: > And we also should think of cookies, just as you suggested

agree, we should in separate tasks: https://github.com/saveourtool/save-cloud/issues/1864

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
sanyavertolet: Yep, that's what I meant

nulls: added on request part

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
sanyavertolet: good

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
sanyavertolet: oh ok, but it seems that it is better to use unified high(er)-level params parsing logic where applicable

sanyavertolet (approved):

PR #1861, opened 2023-02-10
Correct the path/URL manupulation when `saveDemoAgentDistroFileUrl` is unset

0x6675636b796f75676974687562:
This change also renames the property from `saveDemoAgentDistroFilepath` to `saveDemoAgentDistroFileUrl`.


sanyavertolet (approved):
Same changes should be done with `save-agent`'s configuration

0x6675636b796f75676974687562:
> Same changes should be done with `save-agent`'s configuration

Thanks for the heads-up.
I've addressed this with 226d9447aa2578beeaf637bc543915cc70050fdb

PR #1865, opened 2023-02-13
demo-service.yaml

sanyavertolet:
### What's done:
 * Added `NodePort` service for `demo` when and only when profile is `dev`

### DIFF
--- a/save-cloud-charts/save-cloud/templates/demo-service.yaml
+++ b/save-cloud-charts/save-cloud/templates/demo-service.yaml
@@ -10,3 +10,26 @@ spec:
     {{- include "service.common.ports" (dict "service" .Values.demo) | nindent 4 }}
   selector:
     {{- include "service.common.selectors" (dict "service" .Values.demo) | nindent 4 }}
+
+---
+
+# Required for debug
+
+{{ if eq .Values.demo.profile "dev" }}
+
+apiVersion: v1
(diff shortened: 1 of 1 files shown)


save-cloud-charts/save-cloud/templates/demo-service.yaml
```diff
@@ -10,3 +10,26 @@ spec:
     {{- include "service.common.ports" (dict "service" .Values.demo) | nindent 4 }}
   selector:
     {{- include "service.common.selectors" (dict "service" .Values.demo) | nindent 4 }}
+
+---
+
+# Required for debug
+
+{{ if eq .Values.demo.profile "dev" }}
```
nulls, line 18: port forwarding doesn't fit this request?

nulls (approved):

save-cloud-charts/save-cloud/templates/demo-service.yaml
sanyavertolet: `port-forwarding` exposes port for a single machine while `NodePort` configures the port for every machine. Moreover, `port-forwarding` seems to configure cluster dynamically while `NodePort` does it statically.

Well, it seems that `port-forwarding` fits the request. In this case, this PR is not needed. Forgot about `port-forwarding`.

PR #1874, opened 2023-02-15
PodRunner implementation

### LINKED ISSUE #1873: Implement `PodRunner` interface
With #1869 merged we will have almost all infrastructure to support demo runs in a separate Kubernetes pod. In order to support runs in pod, we will need to implement `PodRunner` interface that should be implemented by `DemoPodRunner`, which `run` method would trigger execution in running pod.

sanyavertolet:
This PR closes #1873 

### What's done:
 * Added `PodRunner` interface
 * Added `DemoPodRunner` implementation of `PodRunner`
 * Unified `CliRunnerFactory` to be able to create also `DemoPodRunner`
 * Refactored `Runner` to make `run` method receive only one parameter of type `DemoRunRequest` instead of two: `Path` and `DemoRunRequest`


save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/DefaultRunnerFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.saveourtool.save.demo.runners.cli.DemoCliRunner
+import com.saveourtool.save.demo.runners.command.CommandBuilder
+import com.saveourtool.save.demo.runners.pod.DemoPodRunner
+import com.saveourtool.save.demo.service.KubernetesService
+import com.saveourtool.save.demo.storage.DependencyStorage
+
+import org.springframework.stereotype.Component
+
+/**
+ * Factory creates both [DemoCliRunner] and [DemoCliRunner].
```
kgevorkyan, line 14:
```suggestion
 * Factory creates both [DemoCliRunner] and [DemoPodRunner].
```

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/RunnerFactory.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        /**
+         * Runner that executes [Demo] in another pod
+         */
+        POD,
+
+        /**
+         * Runner that executes [Demo] in the same process as save-demo
+         *
+         * **Should not be used**
```
kgevorkyan, line 34: executes [Demo] in the same process, but shouldn't be used
may be describe why? or remove it at all? we hold this for back compatibility or what?

kgevorkyan (approved):
lgtm

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/RunnerFactory.kt
sanyavertolet: The point is that if we once decide to move `save-demo-cpg` to `save-demo`, it might be useful for some time, until it is started in external pod. 


Will update kdoc and say that this runner might be bad for `save-demo` performance

sanyavertolet: Done

PR #1875, opened 2023-02-17
save-demo-agent wrong exception catching

sanyavertolet:
The root of the problem is that we have `require` call in native implementation of `parseConfig` that checks the file presence (`require` throws `IllegalArgumentException`) while processing `FileNotFoundException`.

### What's done:
* Removed `require` check in `parseConfig`
* Fixed logging in `save-demo-agent`'s entry point


kgevorkyan:
> The root of the problem is that we have require call in native implementation of parseConfig that checks the file presence (require throws IllegalArgumentException) while processing FileNotFoundException.

May be then change `require` to manual `throw`, which file throw `FileNotFoundException`?
Or make an wrapper around this require? Which will catch `IllegalArgumentException` and then rethrow `FileNotFoundException`?

sanyavertolet:
Discussed offline - no need to check file existence as if it doesn't exist, `FileNotFoundException` will be thrown on lower level. All we need is to catch `FileNotFoundException`.

kgevorkyan (approved):

PR #1867, opened 2023-02-13
Add `InternalFileStorage` for *save-cli* and *save-agent*

nulls:
What's done:
- added `InternalFileStorage`
- removed endpoints to download *save-cli* and *save-agent* from backend
- updated tests
- updated `README.md`

It closes #1866

PR #1879, opened 2023-02-17
Migrated S3Operations to CompletableFuture

nulls:
Extracted from #1867


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/DefaultS3Operations.kt
```diff
@@ -210,11 +196,14 @@ class DefaultS3Operations(
             .build()
     }
 
-    private fun <T : Any> CompletableFuture<T>.toMonoAndPublishOn(): Mono<T> = toMono().publishOn(scheduler)
-
     companion object {
-        private fun <T : Any> Mono<T>.handleNoSuchKeyException(): Mono<T> = onErrorResume(NoSuchKeyException::class.java) {
-            Mono.empty()
+        private val region = Region.AWS_ISO_GLOBAL
```
kgevorkyan, line 200: some detalization maybe, like `defaultS3Region` or smth like this

nulls: added

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/DefaultS3Operations.kt
```diff
@@ -34,16 +33,19 @@ class DefaultS3Operations(
 ) : S3Operations, AutoCloseable {
     private val bucketName = properties.bucketName
     private val credentialsProvider: AwsCredentialsProvider = properties.credentials.toAwsCredentialsProvider()
-    private val executorService = with(properties.async) {
```
akuleshov7, line 37: I got very far from this, so a stupid question: if this was an implementation of S3Operations, why you didn't have these `override` methods here before? I do not understand what has changed

nulls: S3Operations returned `Mono` or `Flux` previously
Now we return `CompletableFuture` to implement implementation of `Storage` using coroutines.
It was done to avoid 2 casting: `CompletableFuture` to `Mono`/`Flux` to `suspend`/`Flow`

akuleshov7: And where `override` methods appeared from? 
`executorService` - initially you made a private method and now it's an override. Why it wasn't an override from a very beginning? Because of a different type?

akuleshov7: CompletableFuture is not a right way in case of Spring Flux integration, but I understand your problem with casting. 
Basically that should have been done in Spring S3 integration...

akuleshov7 (approved):
lgtm

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/s3/DefaultS3Operations.kt
nulls: make it private

PR #1884, opened 2023-02-20
Extracted `StorageProjectReactor` and `StoragePreSignedUrl` from `Storage`

nulls:
### What's done:
- Extracted `StorageProjectReactor` with methods which use `Mono` and `Flux`
- Extracted `StoragePreSignedUrl ` with methods to create pre-signed urls
- Introduced `StorageUsingProjectReactor` which implements `StorageProjectReactor` and `StoragePreSignedUrl `
- Moved init logic to common abstract classes: it checks that init method is done (or tried to be called) before calling methods of storages

It was done to simplify introducing `StorageUsingCoroutines`

Extracted from #1867


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractSimpleStorageUsingProjectReactor.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.saveourtool.save.storage.key.S3KeyManager
+
+/**
+ * S3 implementation of Storage
+ *
+ * @param s3Operations [S3Operations] to operate with S3
+ * @param prefix a common prefix for all S3 keys in this storage
+ * @param K type of key
+ */
+abstract class AbstractSimpleStorageUsingProjectReactor<K : Any>(
```
sanyavertolet, line 14: Maybe improve naming? Like `AbstractSimpleReactiveStorage`?

sanyavertolet (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/AbstractSimpleStorageUsingProjectReactor.kt
nulls: renamed `StorageUsingProjectReactor` to `ReactiveStorage`, but leave as is `StorageProjectReactor`

PR #1885, opened 2023-02-20
save-demo interacts with save-demo-agent

sanyavertolet:
Please note that `port-forwarding` is very slow and will be replaces with proper `Service` configuration in next PR

### What's done:
 * Added `demoAgentRequestWrapper` that is used in order to enable `port-forwarding` and close it
 * Added `Role` configuration for `save-demo` to `demo-service-account.yaml`
 * Tested demo-agent creation and run


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
```diff
@@ -18,15 +19,41 @@ typealias StringList = List<String>
...
-        delay(delayMillis)
-        retry(times - 1, delayMillis, action)
-    } else {
-        null
+    action: suspend (Int) -> T?,
+): Pair<T?, List<Throwable>> {
+    val caughtExceptions: MutableList<Throwable> = mutableListOf()
+    times.downTo(0).map { iteration ->
+        try {
+            delay(delayMillis)
```
nulls, line 36: you make delay before the first attempt -- is it required?

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -2,20 +2,26 @@
...
 import reactor.core.publisher.Flux
+import java.io.IOException
 import java.nio.ByteBuffer
 import kotlin.random.Random
 
 private const val DEFAULT_BUFFER_SIZE = 4096
+private const val PORT_RANGE_FROM = 20000
+private const val PORT_RANGE_TO = 30000
+
+private val logger = LoggerFactory.getLogger("c.s.s.d.u.HttpUtils")
```
nulls, line 24: why such name? logger prints `c.s.s.d.u.` because it's configured in such way. it should be proper class name

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -40,26 +46,47 @@ fun ByteReadChannel.toByteBufferFlux(): Flux<ByteBuffer> = DataBufferUtils.readI
...
-            }
-    }
+): R {
+    val podName = demoPod.metadata.name
+    val podResource = kc.pods().withName(podName)
+    val containerPort = demoPod.spec.containers.single()
+        .ports
+        .single()
+        .containerPort
+    val saveDemoPort = Random.nextInt(PORT_RANGE_FROM, PORT_RANGE_TO)
```
nulls, line 63: there is a guarantee that port is free? Maybe use `ServerSocket(0).localPort`?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
```diff
@@ -23,15 +24,41 @@ fun <T : Any> T?.isNotNull(): Boolean = this != null
  * @param action action that should be invoked
  * @return [T] if the result was fetched in [times] attempts, null otherwise
  */
+@Suppress("TooGenericExceptionCaught")
 suspend fun <T : Any> retry(
```
kgevorkyan, line 28: may be rename it now, it aim to indicate, that it also collect exceptions now
also, please update `@return` in kdoc

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Demo.kt
```diff
@@ -46,7 +46,10 @@ class Demo(
         }
     }
 
-    private fun projectCoordinates() = ProjectCoordinates(
```
kgevorkyan, line 49: could it be an class field, instead of manual crating a new instance any time?

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/pod/PodRunner.kt
```diff
@@ -28,35 +29,24 @@ interface PodRunner : Runner {
...
-        .let { response -> response.takeIf { it.status.isSuccess() }?.body() }
 
     /**
-     * @param demoRunRequest
-     * @return url of a pod that runs demo
+     * @param demoRunRequest run request
+     * @return [HttpResponse] if response received, null if something went wrong and response was not received
      */
-    fun getUrl(demoRunRequest: DemoRunRequest): Mono<String>
+    suspend fun sendRunRequest(demoRunRequest: DemoRunRequest): HttpResponse?
```
kgevorkyan, line 47: will it better to create a manual http response instead of null, in case of some errors?

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/KubernetesService.kt
```diff
@@ -72,72 +75,117 @@ class KubernetesService(
...
 
-        if (requestStatus.isSuccess()) {
-            logAndRespond(HttpStatusCode.OK, logger::info) {
-                "Job successfully started."
+    private suspend fun sendConfigurationRequestRetrying(
+        url: Url,
+        configuration: DemoAgentConfig,
+        retryNumber: Int,
+    ): StringResponse {
+        val requestStatusAndErrors = retry(retryNumber) { iteration ->
```
kgevorkyan, line 142: ?
```suggestion
        val (requestStatus, errors) = retry(retryNumber) { iteration ->
```

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: It is done in order to provide better information about the file that prints logs

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/KubernetesService.kt
kgevorkyan: thus we won't access `.first` and `.second`

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: For now there is no guarantee, but I reckon we should not improve port-forwarding solution as it is very slow. We should come up with a new solution.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
sanyavertolet: No, it is not
Will fix it

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/KubernetesUtils.kt
```diff
@@ -129,10 +127,23 @@ private fun demoAgentContainerSpec(
...
+    val envOptions = sequenceOf(
+        "KTOR_LOG_LEVEL" to "TRACE",
+    )
+
+    val startupCommand = downloadAndRunAgentCommand(agentDownloadUrl, AgentType.DEMO_AGENT, envOptions = envOptions)
 
     command = listOf("sh", "-c", startupCommand)
 
+    ports = listOf(
+        ContainerPort().apply {
```
kgevorkyan, line 139: maybe create some extension like `ContainerPort().default()` or smth similar with these values
could we?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
sanyavertolet: Moved `delay` to `finally` block

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Demo.kt
sanyavertolet: No, it couldn't as `Demo` is an `entity` and the only solution is to create a table for `ProjectCoordinates` which is weird

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/KubernetesService.kt
sanyavertolet: Yes, sure

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/KubernetesUtils.kt
sanyavertolet: Makes sense, will do it

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/pod/PodRunner.kt
sanyavertolet: Didn't get you point

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
sanyavertolet: oh sure

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/pod/PodRunner.kt
sanyavertolet: Discussed offline - return nullable is ok as far as it is processed on `Controller` level - wrapped into `Mono` and checked to be empty

save-demo/src/main/kotlin/com/saveourtool/save/demo/entity/Demo.kt
sanyavertolet: Oh I guess I got your point here, but not sure if we really want to do it in this PR. Probably should be done in one of next PRs with `Demo` refactoring (adding a new column)

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/KubernetesUtils.kt
sanyavertolet: Done

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
sanyavertolet: Updated Kdoc, don't want to change the name though

nulls (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
```diff
@@ -14,24 +14,53 @@ typealias StringList = List<String>
...
+    val caughtExceptions: MutableList<Throwable> = mutableListOf()
+    times.downTo(0).map { iteration ->
+        try {
+            action(iteration)?.let { result ->
+                return result to caughtExceptions
+            }
+        } catch (e: Throwable) {
+            caughtExceptions.add(e)
+        } finally {
+            delay(delayMillis)
```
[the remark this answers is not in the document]
sanyavertolet: If we get the result on first try, we will return the result from the function

save-cloud-charts/save-cloud/templates/demo-deployment.yaml
```diff
@@ -60,9 +60,9 @@ spec:
           {{- include "spring-boot.management" .Values.demo | nindent 10 }}
           resources:
             limits:
-              memory: 1000M
+              memory: 1200M
```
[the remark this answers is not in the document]
sanyavertolet: Rollbacked 1200/1000 to 1000/800

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/utils/KotlinUtils.kt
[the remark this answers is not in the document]
nulls: nope, finally will be called

PR #1886, opened 2023-02-20
excluded com.saveourtool.sarifutils:fixpatches

nulls:


nulls:
currently, sonatype snapshot is avaivale.
strategically, it's better to handle by #1887

PR #1890, opened 2023-02-22
[DEMO] make kubernetes optional

sanyavertolet:
### What's done:
 * Moved all logic connected with `kubernetes` interactions from controllers to `DemoService`
 * Made `KubernetesConfig` nullable
 * Added `Profile` annotations for all the kubernetes stuff

### DIFF
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/DefaultRunnerFactory.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/DefaultRunnerFactory.kt
@@ -17,11 +17,11 @@ import org.springframework.stereotype.Component
 class DefaultRunnerFactory(
     private val dependencyStorage: DependencyStorage,
     private val commandBuilder: CommandBuilder,
-    private val kubernetesService: KubernetesService,
+    private val kubernetesService: KubernetesService?,
 ) : RunnerFactory {
-    override fun create(demo: Demo, version: String, type: RunnerType): Runner = when (type) {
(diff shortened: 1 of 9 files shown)


save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/DefaultRunnerFactory.kt
```diff
@@ -17,11 +17,11 @@ import org.springframework.stereotype.Component
 class DefaultRunnerFactory(
     private val dependencyStorage: DependencyStorage,
     private val commandBuilder: CommandBuilder,
-    private val kubernetesService: KubernetesService,
+    private val kubernetesService: KubernetesService?,
```
nulls, line 20: does spring support it? maybe need to add something like `@Autowired(required = false)`

nulls (approved):

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/DefaultRunnerFactory.kt
sanyavertolet: Works fine when running outside Minikube, haven't test inside Kubernetes yet
Will test it

sanyavertolet: It seem to work fine as I use methods from `DemoSerivce` that require `KubernetesSerivce` check

PR #1896, opened 2023-02-27
okio-extras

sanyavertolet:
### What's done:
 * Added `saveourtool/okio-extras` to `settings.gradle.kts`

#### See also:
https://github.com/saveourtool/save-cli/commit/67267e58aff886359a9461b2da21a0b998e4bb80

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 kotlin = "1.8.10"
 jetbrains-annotations = "24.0.0"
-save-core = "0.3.7"
+save-core = "0.3.8"
 ktor = "2.2.2"
 okio = "3.3.0"
 serialization = "1.4.1"
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -9,6 +9,16 @@ dependencyResolutionManagement {
             }
         }
         mavenCentral()
+        maven {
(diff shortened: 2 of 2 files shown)


sanyavertolet:
@0x6675636b796f75676974687562 
telegram-cloud-photo-size-2-5469822743777231414-y

sanyavertolet:
https://scans.gradle.com/s/seosg3ubpcuyw/failure#1

akuleshov7 (approved):

0x6675636b796f75676974687562:
The cause of the problem is that both `save-cli` **0.3.7** and the newly-released `save-cli` **0.3.8** depend on the snapshot version of `sarif-utils` (**0.2.0-SNAPSHOT**), and the latest snapshot version now depends on `okio-extras`.

So, from the standpoint of `:save-agent`, the versions _haven't_ changed but the dependencies _have_.

kgevorkyan:
@sanyavertolet `save-cli` 0.3.8 released, so you can change `lib.versions` and retry

0x6675636b796f75676974687562 (approved):

0x6675636b796f75676974687562:
The build seems to pass (at least, locally).

Yet, I suggest we switch to the **released** `sarif-utils` version in `save-cli`: saveourtool/save-cli#507.

PR #1906, opened 2023-02-27
Agent Jobs should be deleted on completion

sanyavertolet:
### What's done:
 * Added `ttlSecondsAfterFinished` to job spec (for now hardcoded to be 3 mins = `180`)
 * Added `ttlAfterFinished` to `KubernetesSettings`

### DIFF
--- a/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
+++ b/save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
@@ -54,7 +54,7 @@ class KubernetesManager(
             }
             spec = JobSpec().apply {
                 parallelism = replicas
-                ttlSecondsAfterFinished = TTL_AFTER_COMPLETED
+                ttlSecondsAfterFinished = kubernetesSettings.ttlAfterFinished.toInt(DurationUnit.SECONDS)
                 // do not attempt to restart failed pods, because if we manually stop pods by deleting them,
                 // job controller would think that they need to be restarted
                 backoffLimit = 0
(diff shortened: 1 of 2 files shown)


save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
```diff
@@ -54,6 +54,7 @@ class KubernetesManager(
             }
             spec = JobSpec().apply {
                 parallelism = replicas
+                ttlSecondsAfterFinished = TTL_AFTER_COMPLETED
```
nulls, line 57: can we move it to `com.saveourtool.save.orchestrator.config.ConfigProperties.KubernetesSettings` ?
```
        val ttlAfterFinished: Duration = 180.seconds,
```
```suggestion
                ttlSecondsAfterFinished = kubernetesSettings.ttlAfterFinished.toInt(DurationUnit.SECONDS)
```

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
```diff
@@ -180,6 +181,7 @@ class KubernetesManager(
     companion object {
         private val logger = LoggerFactory.getLogger(KubernetesManager::class.java)
         private const val EXECUTION_ID_LABEL = "executionId"
+        private const val TTL_AFTER_COMPLETED = 1800
```
nulls, line 184: 30 minutes?

nulls (changes requested):

save-orchestrator-common/src/main/kotlin/com/saveourtool/save/orchestrator/kubernetes/KubernetesManager.kt
sanyavertolet: Oops, will remove odd nil

sanyavertolet: Done

nulls (approved):

PR #1910, opened 2023-02-28
Replace port-forwarding with Service

### LINKED ISSUE #1904: [DEMO] replace port-forwarding with Service
Right now when we want to send request to `save-demo-agent` we initialize port-forwarding (and stop it on request completion) which is extremely slow. We either need to create a `Service` for each `save-demo-agent` (inside `save-demo` namespace using Kubernetes Client for example) or come up with an idea of creating one unified Service, probably some kind of `LoadBalancer`. Maybe @petertrr has an idea on this?🥺

sanyavertolet:
This PR closes #1904 

### What's done:
 * Added `demo-agent-service.yaml`
 * Replaced `port-forwarding` with dns resolution
 * Added `subdomain` for all jobs that are created by `save-demo`


save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -21,6 +21,12 @@ private const val DEFAULT_BUFFER_SIZE = 4096
 private const val PORT_RANGE_FROM = 20000
 private const val PORT_RANGE_TO = 30000
 
+/**
+ * Subdomain of all save-demo-agents
+ */
+const val DEMO_SUBDOMAIN_NAME = "demo-agent-service"
+const val CLUSTER_DOMAIN = "cluster.local"
+
 private val logger = LoggerFactory.getLogger("c.s.s.d.u.HttpUtils")
```
nulls, line 30: maybe change to `LoggerFactory.getLogger(object {}.javaClass.enclosingClass::class.java)`?

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -43,31 +49,29 @@ fun ByteReadChannel.toByteBufferFlux(): Flux<ByteBuffer> = DataBufferUtils.readI
...
  * @param urlPathSegments list of endpoint path segments (e.g. ["run"])
  * @param demoPod [Pod] with save-demo-agent that runs requested demo
- * @param kc [KubernetesClient] that is required for kubernetes interactions
- * @param request callback that receives url and port
+ * @param request callback that receives [Url]
  * @return result of type [R]
  * @throws IOException rethrown from [request]
  */
-suspend fun <R> demoAgentRequestWrapper(
+private suspend fun <T> KubernetesClient.portForwardingRequest(
```
nulls, line 56: do we still need `portForwardingRequest`?

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -90,3 +94,42 @@ suspend fun <R> demoAgentRequestWrapper(
...
+    return request(
+        URLBuilder(
+            host = host,
+            port = SAVE_DEMO_AGENT_DEFAULT_PORT,
+            pathSegments = urlPathSegments,
+        ).build()
+    )
+}
+
+private fun addressToDnsResolution(podIp: String, namespace: String = "save-cloud") = listOf(
```
nulls, line 128: suggestion to extract to a public function and add tests

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -21,6 +21,12 @@ private const val DEFAULT_BUFFER_SIZE = 4096
 private const val PORT_RANGE_FROM = 20000
 private const val PORT_RANGE_TO = 30000
 
+/**
+ * Subdomain of all save-demo-agents
+ */
+const val DEMO_SUBDOMAIN_NAME = "demo-agent-service"
```
petertrr, line 27: You could actually avoid duplication of this string like this: You could place it into `values.yaml`, reference in YAML declaration for Service and pass to `save-demo` either via ConfigMap or env var.

save-cloud-charts/save-cloud/templates/demo-agent-service.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: demo-agent-service
+spec:
+  selector:
+    io.kompose.service: save-demo-agent
+  clusterIP: None
+  ports:
+    - port: 23456
```
petertrr, line 10: Is this matching specific port of agent? Could be extracted into `values.yaml` too

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
```diff
@@ -90,3 +94,42 @@ suspend fun <R> demoAgentRequestWrapper(
...
+ * @return result of type [R]
+ */
+suspend fun <R> demoAgentRequestWrapper(
+    urlPathSegments: List<String>,
+    demoPod: Pod,
+    request: suspend (Url) -> R,
+): R {
+    val host = addressToDnsResolution(demoPod.status.podIP)
+    return request(
+        URLBuilder(
```
nulls, line 120: do we really need to pass `listOf(String)`, maybe expects a single string with split by `/`?
```suggestion
        URLBuilder(
            host = host,
            port = SAVE_DEMO_AGENT_DEFAULT_PORT,
         )
         .appendPathSegments(components = arrayOf(endpoint), false)
         .build()
```

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: Ok, will do it

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: I would like to leave it here for now just in case. Do now want my time to be spent so useless

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: Ok, will do it

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: Ok, will change it, but a bit differently

save-cloud-charts/save-cloud/templates/demo-agent-service.yaml
nulls: or migrate to default 80 port for example

save-demo/src/main/kotlin/com/saveourtool/save/demo/utils/HttpUtils.kt
sanyavertolet: Done

save-cloud-charts/save-cloud/templates/demo-agent-service.yaml
sanyavertolet: Implemented the first solution

petertrr (approved):

Issue #780, opened 2022-05-19
Frontend: Migrate to Kotlin/JS IR compiler backend


petertrr:
Currently in `save-frontend/build.gradle.kts` we mention `as for `-pre.148-kotlin-1.4.21`, react-table gives errors with IR`. We should check whether these errors still persist and whether we have any other errors.

Another consideration: does `kotlin-react` have better IR compatibility than `kotlin-react-legacy`?

Migrating to IR can give us opportunity to use kotest for K/JS tests, as mentioned in https://github.com/analysis-dev/save-cloud/issues/537

petertrr:
With kotlin-wrappers `pre-330` and Kotlin 1.6.21 there are no errors with react-table. The only immediate error that I was able to spot: modal windows now occupy almost the whole screen.

akuleshov7:
Oh, that is a huge task that involves a lot of investigations of latest kotlin compiler.  @nulls @unix-junkie @petertrr @icemachined can be interesting for you.

nulls:
@akuleshov7 @icemachined I will take it

nulls:
Tried to migrate since we started to get a lot of warning about migration to Kotlin JS IR after upgrading to kotlin 1.8.x.

The main changes for us: JsExport is not applied by default. Also `KDoc` says that: `It is currently prohibited to export the following kinds of declarations: [...] enum classes`. It requires some refactoring (but I found here: https://github.com/Kotlin/kotlinx-datetime/issues/72#issuecomment-1092567580 that enums are supported and looks like it was).
But also all DTOs should be marked as `JsExport`. Here we get the next issue: [kotlinx-datetime doesn't export their classes
](https://github.com/Kotlin/kotlinx-datetime/issues/72)
+1: Long is not exportable for JS.

These problems look as manageable, but will take some time.

Benefits (except presented by Kotlin advocates): mapping Kotlin code to exceptions in browser does work, you can navigate to source code from stacktrace of exception

PR #1912, opened 2023-03-01
Migrate to Kotlin/JS IR compiler backend

nulls:
### What's done:
- added `optIn` for `kotlin.js.ExperimentalJsExport` on `build.gradle.kts` level
- added `JsExport` for `*Dto` classes and enum

It closes https://github.com/saveourtool/save-cloud/issues/780


sanyavertolet:
To my mind this PR should be split into several ones. We have different problems e.g. `withRouter`, `enum class`es etc. Suggestion to create an issue with dot list and fix those issues step-by-step. As you seem to know much more about migrating to new Kotlin/JS IR compiler backend, I would like you to create this issue. 🥺

sanyavertolet:
Should be done in several steps

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/benchmarks/BenchmarkEntity.kt
```diff
@@ -27,7 +30,7 @@ enum class BenchmarkCategoryEnum {
         private val postfixInRegex = values().joinToString("|") { it.name.lowercase() }
         override val nameOfTheHeadUrlSection = "archive"
         override val defaultTab: BenchmarkCategoryEnum = ALL
-        override val regexForUrlClassification: Regex = Regex("/$nameOfTheHeadUrlSection/[^/]+/($postfixInRegex)")
```
kgevorkyan, line 30: It will be cast/ or used as regex under hood?

kgevorkyan (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/benchmarks/BenchmarkEntity.kt
nulls: yes, when it's needed, regex will be created inside a method. Regex cannot be a part of public api

save-cloud-common/build.gradle.kts
```diff
@@ -36,7 +36,10 @@ kotlin {
 
     sourceSets {
         sourceSets.all {
-            languageSettings.optIn("kotlin.RequiresOptIn")
+            languageSettings.apply {
+                optIn("kotlin.RequiresOptIn")
+                optIn("kotlin.js.ExperimentalJsExport")
```
sanyavertolet, line 41: Guess it should be in `jsMain` only

nulls: actually not, `JsExport` is applicable for common and shared code. Like `Dto` classes for examle

PR #1919, opened 2023-03-02
Removed `AgentEnvName.AGENT_VERSION`

nulls:
### What's done:
- removed `AgentEnvName.AGENT_VERSION`
- added `generateVersionFileTaskProvider` to `save-agent`
- used `generated.AGENT_VERSION`

It's extracted from #1916


sanyavertolet (approved):

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-cli-configuration.gradle.kts
```diff
@@ -51,13 +51,11 @@ dependencies {
     )
 }
 
-// todo: this logic is duplicated between agent and frontend, can be moved to a shared plugin in gradle/plugins
```
kgevorkyan, line 54: now it's not duplicated or what?

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-cli-configuration.gradle.kts
```diff
@@ -68,7 +66,6 @@ val generateVersionFileTaskProvider = tasks.register("generateVersionFile") {
             package generated
 
             internal const val SAVE_CORE_VERSION = "$saveCliVersion"
-            internal const val SAVE_CLOUD_VERSION = "$version"
```
kgevorkyan, line 71: was unused?

save-agent/build.gradle.kts
```diff
@@ -30,13 +30,39 @@ kotlin {
         }
     }
 
+    // generate kotlin file with project version
+    val generateVersionFileTaskProvider = tasks.register("generateVersionFile") {
```
kgevorkyan, line 34: do already we have this logic in save-cloud somewhere? may be we can create some method, which will expect few arguments like `internal const val AGENT_VERSION = "$version"` etc

kgevorkyan (approved):

save-agent/build.gradle.kts
kgevorkyan: or we have something similar in save-cli?

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-cli-configuration.gradle.kts
nulls: yes, we used it in backend\sandbox before to pass only version of save-agent

save-agent/build.gradle.kts
nulls: > do already we have this logic in save-cloud somewhere? may be we can create some method, which will expect few arguments like internal const val AGENT_VERSION = "$version" etc

Yes, we have similar logic with frontend

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-cli-configuration.gradle.kts
nulls: _frontend_ needs only `SAVE_CLOUD_VERSION` (as `SAVE_VERSION`), _backend_ and _sandbox_ need `SAVE_CORE_VERSION`.
_agent_ needs `AGENT_VERSION` (it's `SAVE_CLOUD_VERSION`).
Now we have similar task between _frontend_ and _agent_ -- will try to combine

Cheshiriks (approved):

save-agent/build.gradle.kts
nulls: moved to a common place

gradle/plugins/src/main/kotlin/com/saveourtool/save/buildutils/save-cli-configuration.gradle.kts
nulls: done

0x6675636b796f75676974687562 (approved):

PR #1923, opened 2023-03-02
Demo deletion

### LINKED ISSUE #1922: Implement demo deletion
An endpoint for demo deletion is required. It should also be supported on frontend

sanyavertolet:
This PR closes #1922

### What's done:
 * Implemented demo deletion
 * Added frontend button that would delete demo
 * Added `ttlSecondsAfterFinished` to `s3-deployment.yaml` in order to clean the startup job (`minio-create-bucket`)

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
@@ -333,4 +333,44 @@ class DemoManagerController(
                     )
                 )
         }
+
+    @PostMapping("/{organizationName}/{projectName}/delete")
+    @RequiresAuthorizationSourceHeader
+    @PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")
+    @Parameters(
+        Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of saveourtool organization", required = true),
+        Parameter(name = "projectName", `in` = ParameterIn.PATH, description = "name of saveourtool project", required = true),
+    )
+    @Operation(
+        method = "POST",
+        summary = "Delete demo.",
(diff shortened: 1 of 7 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -333,4 +333,44 @@ class DemoManagerController(
                     )
                 )
         }
+
+    @PostMapping("/{organizationName}/{projectName}/delete")
+    @RequiresAuthorizationSourceHeader
+    @PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")
```
nulls, line 339: do we still need it taking into account that you check permission on project level?

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
```diff
@@ -104,4 +107,17 @@ class DemoService(
...
+    /**
+     * @param demo [Demo] entity
+     * @param version version of demo
+     * @return [Mono] of [Unit]
+     */
+    fun delete(demo: Demo, version: String): Mono<StringResponse> = stop(demo)
+        .let { dependencyStorage.list(demo, version) }
+        .concatMap { dependencyStorage.delete(it) }
+        .collectList()
+        .publishOn(Schedulers.boundedElastic())
```
nulls, line 120: why not `blockingToMono`?

sanyavertolet: Well, it is actually the same but without additional `Mono` creation

nulls: yeap, but it's more visual to show blocking operations. but up to you

PR #1941, opened 2023-03-06
downgrade spring-boot

nulls:
It is due to https://github.com/spring-projects/spring-data-jpa/issues/2812


kgevorkyan:
@nulls by the way, do we have any plans to migrate to spring boor 3x one day ? forget, discuss we this or not

nulls:
> @nulls by the way, do we have any plans to migrate to spring boor 3x one day ? forget, discuss we this or not

As I remember, we have discussed it and decided to no take such big changes. But we don't have even an issue for it. Created: https://github.com/saveourtool/save-cloud/issues/1943 for tracking

PR #1956, opened 2023-03-07
Get rid of type aliases

0x6675636b796f75676974687562:
- This converts `CloneAndProcessDirectoryAction`, `GitRepositoryProcessor`, and `CloneResult` into classes/interfaces.
 - Additionally, `fetch()` now returns a `Flux<TestSuiteValidationResult>` instead of `Mono<Unit>`.
 - Related: #1096


akuleshov7 (approved):
Lgtm, but basically diktat and detekt will ask you to use type alias anyway in case of complex generics, so you will have to use them anyway here (I guess)

0x6675636b796f75676974687562:
@akuleshov7 You're right indeed, but I needed a `typealias` with a type boundary (e.g.: both `Foo<Mono<T>` and `Foo<Flux<T>`), and `typealias`'es don't offer this yet.

PR #1960, opened 2023-03-09
save-demo-agent logging improvements

sanyavertolet:
### What's done:
 * Improved logging on save-demo-agent
 * Fixed URLs on projectDemoMenu
 * Fixed getDemoInfo in DemoManagerController
 * Fixed registerLiquibaseTask name (reqisterLiquibaseTask -> registerLiquibaseTask)
 * Removed PreAuthorization annotations from DemoManagerController

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
@@ -6,7 +6,6 @@ import com.saveourtool.save.backend.service.LnkProjectGithubService
 import com.saveourtool.save.backend.service.ProjectService
 import com.saveourtool.save.configs.RequiresAuthorizationSourceHeader
 import com.saveourtool.save.demo.DemoCreationRequest
-import com.saveourtool.save.demo.DemoDto
 import com.saveourtool.save.entities.FileDto
 import com.saveourtool.save.entities.Project
 import com.saveourtool.save.permission.Permission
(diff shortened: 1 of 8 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -53,7 +49,6 @@ class DemoManagerController(
 
     @PostMapping("/{organizationName}/{projectName}/add")
     @RequiresAuthorizationSourceHeader
-    @PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")
```
kgevorkyan, line 56: why do you remove it?

Cheshiriks (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -216,69 +208,8 @@ class DemoManagerController(
             .bodyToMono()
     }
 
-    @GetMapping("/{organizationName}/{projectName}/status")
```
kgevorkyan, line 219: and what's was wrong with these endpoints?

sanyavertolet: This endpoints forward requests to demo's api part. No Permission check is required for it so there is better to send requests to save-demo directly

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
sanyavertolet: It was because all the save-demo was in development. Not it is time to fix bugs so I decided to remove it

kgevorkyan (approved):

PR #1966, opened 2023-03-10
Rollback helm for Grafana\loki

### LINKED ISSUE #1947: Grafana loki doesn't work
Grafana cannot connect to `http://loki`

```
logger=context userId=1 orgId=1 uname=admin t=2023-03-07T11:06:11.879194976Z level=error msg="Failed to call resource" error="Get \"http://loki:3100/loki/api/v1/labels?start=1678186571442000000&end=1678187171442000000\": dial tcp: lookup loki on 10.247.3.10:53: no such host" traceID=
```

nulls:
Rollbacked to latest working version in our environment

It closes #1947


kgevorkyan (commented):
what's about `docker-compose.yaml`?

https://github.com/saveourtool/save-cloud/commit/376a50994d9ff8a58d10e6d17052a8345b92334a

nulls:
> what's about `docker-compose.yaml`?
> 
> [376a509](https://github.com/saveourtool/save-cloud/commit/376a50994d9ff8a58d10e6d17052a8345b92334a)

we actually don't use docker compose anymore (after migration to k8s).

Exclusion is the internal instance which we do not upgrade.

So, I'll rollback `docker-compose.yaml`, but won't be able to test it

save-cloud-charts/save-cloud/values.yaml
```diff
@@ -89,7 +89,7 @@ grafana:
   enabled: false
   image:
     repository: grafana/grafana
-    tag: 9.4.3
+    tag: 9.3.6
```
petertrr, line 92: I think these versions can be kept as is, because they shouldn't be that dependant on charts

petertrr (approved):

Issue #1967, opened 2023-03-10
Move local run to MiniKube


nulls:

sanyavertolet:
It seems that running everything in minikube is extremely slow as I found out when I was developing save-demo. The idea is to add configuration `fake-kubernetes` that would allow to use kubernetes orchestration when running spring service outside the cluster.

nulls:
> when running spring service outside the cluster.

It is desired solution: only mysql/s3 and save-agent/save-demo-agent should be run in minikube -- the rest components should be run outside

PR #1969, opened 2023-03-10
minikube profile for save-demo

sanyavertolet:
This PR is a part of #1967 

### What's done:
 * Partially added minikube support in order to run save-demo outside from kubernetes but with an ability to use KubernetesService and so on
 * Updated liquibase container version in order to support arm architecture
 * Added load balancer services for local minikube deployment - just in case


sanyavertolet:
@nulls take a look at what I have. Maybe you will have an idea. I get 403 when trying to download from Storage replacing `localhost` with `host.minikube.internal`. The point is that `netcat` says `host.minikube.internal:9000` is available from Minikube (checked with `minikube ssh`). Do you know if there are there any limitations from minio that are connected with hostname?

nulls:
> @nulls take a look at what I have. Maybe you will have an idea. I get 403 when trying to download from Storage replacing `localhost` with `host.minikube.internal`. The point is that `netcat` says `host.minikube.internal:9000` is available from Minikube (checked with `minikube ssh`). Do you know if there are there any limitations from minio that are connected with hostname?

Yes, it is: presigned url is used in generating an url. So replacement should be done on configuration level to use this url in presigned url

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -69,7 +69,7 @@ private fun Project.registerLiquibaseTask(projectName: String, relativeChangeLog
             "--rm",
             "--env", "INSTALL_MYSQL=true",
             "--network", "build_default",
-            "liquibase/liquibase:4.15",
+            "liquibase/liquibase:4.20",
```
nulls, line 72: what's about `liquibase/liquibase` in the rest deployments?

nulls (approved):

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
sanyavertolet: There is no problem with the other deployment. Will change if really required

sanyavertolet: Done

PR #1983, opened 2023-03-13
Tool should be uploaded to save-demo asynchronously

sanyavertolet:
### What's done:
 * Changed `asyncEffect` to `doOnSuccess` in order to initialize file download after the response is received

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
@@ -29,6 +29,7 @@ import org.springframework.web.reactive.function.client.*
 import org.springframework.web.server.ResponseStatusException
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
+import reactor.core.scheduler.Schedulers
 
 /**
  * Controller that allows adding tools to save-demo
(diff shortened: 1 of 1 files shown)


nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -100,16 +101,18 @@ class DemoManagerController(
                 }
                 .toBodilessEntity()
         }
-        .flatMap {
+        .map {
```
nulls, line 104: also suggestion to collect all required info to a single request and send it to demo (in `doOnSuccess` section) and here return a single `accepted`

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
nulls: the same as demo can fail when file started to upload.
but got you point and actually, Demo can return 202 on its side.
As I can see -- backend is just a proxy here (to check authorization and authentication).
So common suggestion to reduce logic on backend and move it to demo

sanyavertolet: ok, may be done in next PR

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/DemoManagerController.kt
```diff
@@ -100,16 +101,18 @@ class DemoManagerController(
...
+            StringResponse.ok("Successfully signed up ${demoCreationRequest.demoDto.projectCoordinates} demo.")
```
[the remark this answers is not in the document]
sanyavertolet: Done

PR #1984, opened 2023-03-14
Frontend doesn't show demo result

sanyavertolet:
### What's done:
 * Fixed frontend error when no demo result was shown
 * Fixed deserialization error on `demoView`
 * Fixed logging in `KubernetesService`

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
@@ -34,6 +34,7 @@ import react.dom.html.ReactHTML.strong
 import web.file.FileReader
 import web.html.InputType
 
+import kotlinx.browser.window
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 
@@ -42,6 +43,8 @@ private val defaultCode = """
     |
     |fun main() {
     |    val SAVEOUR = "tool"
+    |    
+    |    
     |}
 """.trimMargin()
 
@@ -56,8 +59,20 @@ private val defaultCode = """
 )
 val demoRunComponent: FC<DemoRunComponentProps> = FC { props ->
     val (demoRunRequest, setDemoRunRequest) = useState(props.emptyDemoRunRequest)
-    val (diktatResult, setDiktatResult) = useState(DemoResult.empty)
+    val (demoResult, setDemoResult) = useState(DemoResult.empty)
     val (codeLines, setCodeLines) = useState(defaultCode)
+
+    /*
+     * Temporary workaround.
+     *
+     * There is a problem with attempt to process different demos in the same view.
+     * In order to flush all the data corresponding to one demo, this [useEffect] is required
+     */
+    useEffect(window.location.href) {
+        setDemoResult(DemoResult.empty)
+        setCodeLines(defaultCode)
+        setDemoRunRequest(props.emptyDemoRunRequest)
+    }
 
     val sendRunRequest = useDeferredRequest {
         val result: DemoResult = post(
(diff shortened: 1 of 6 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
```diff
@@ -42,6 +43,8 @@ private val defaultCode = """
     |
     |fun main() {
     |    val SAVEOUR = "tool"
+    |
```
nulls, line 46: do we need it? Or is it to have a formatting issue in stub example?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
```diff
@@ -111,3 +89,23 @@ private fun demoView(): FC<DemoViewProps> = FC { props ->
         }
     }
 }
+
+/**
+ * [Props] for DemoView
+ */
+external interface DemoViewProps : Props {
```
nulls, line 96: suggestion to add a flag to disable custom-selector

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
```diff
@@ -53,35 +51,14 @@ val demoView: VFC = VFC {
...
-    var demoRunEndpoint: String
-
-    /**
-     * Optional config name for this demo
-     */
-    var configName: String?
-}
-
-private fun demoView(): FC<DemoViewProps> = FC { props ->
+private val commonDemoView: FC<DemoViewProps> = FC { props ->
```
nulls, line 61: so, `val` is better? not a `fun`?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/DemoRunComponent.kt
sanyavertolet: Yep, it is to make `ktlint` find some formatting issues

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
sanyavertolet: making `FC` be `val` allows to remove line 
```
val commonDemoView = demoView()
```

Actually there is no other difference

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
sanyavertolet: Should be done in PR with several run modes (if you are talking about disabling demoMode selector). Probably it is the task that should be done next

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/demo/DemoView.kt
nulls: > Actually there is no other difference

So it's only refactoring, isn't it?

sanyavertolet: We used to create a new demoView instance, which was recreated during each re-render (and that caused no result on frontend despite the fact that the result was received). Now we use commonDemoView that is created once.

nulls: so `val` is better :)

PR #1985, opened 2023-03-14
FossGraph. Added controller for save new vulnerabilities

### LINKED ISSUE #1980: FossGraph. Add controllers to backend for save new vulnerabilites
Need add controllers to backend for save new vulnerabilites

Cheshiriks:
* Added new controller for save vulnerabilities

Closes #1980

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
@@ -12,7 +12,6 @@ import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
 import io.swagger.v3.oas.annotations.tags.Tags
-import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.*
 import reactor.core.publisher.Mono
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
@@ -5,12 +5,15 @@ import com.saveourtool.save.configs.ApiSwaggerSupport
 import com.saveourtool.save.entities.vulnerabilities.Vulnerability
 import com.saveourtool.save.entities.vulnerability.VulnerabilityDto
 import com.saveourtool.save.filters.VulnerabilityFilter
+import com.saveourtool.save.utils.StringResponse
 import com.saveourtool.save.utils.blockingToMono
 import com.saveourtool.save.v1
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.responses.ApiResponse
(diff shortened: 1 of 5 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
```diff
@@ -55,4 +58,19 @@ class VulnerabilityController(
...
+        summary = "Save vulnerability.",
+        description = "Save vulnerability.",
+    )
+    @ApiResponse(responseCode = "200", description = "Successfully saved vulnerability")
+    fun save(
+        @RequestBody vulnerabilityDto: VulnerabilityDto,
+    ): Mono<StringResponse> = blockingToMono {
+        vulnerabilityService.save(vulnerabilityDto)
+    }.map {
+        ResponseEntity.status(HttpStatus.OK).body("Vulnerability was successfully saved")
```
sanyavertolet, line 74:
```suggestion
        ResponseEntity.ok("Vulnerability was successfully saved")
```

This should be better

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -46,4 +51,35 @@ class VulnerabilityService(
...
+     */
+    @Transactional
+    fun save(vulnerabilityDto: VulnerabilityDto) {
+        val number = vulnerabilityRepository.findMaxId() + 1
+        val name = "SOTV-${LocalDateTime.now().year}-$number"
+
+        val vulnerability = Vulnerability(
+            name = name,
+            progress = 0,
+            projects = emptyList()
```
sanyavertolet, line 66: Is this old by the way? Don't you want to save projects as well?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -46,4 +51,35 @@ class VulnerabilityService(
         val description = vulnerabilityDescriptionRepository.findByVulnerabilityName(name)?.description.orEmpty()
         return vulnerabilityRepository.findByName(name)?.toDtoWithDescription(description) ?: VulnerabilityDto.empty
     }
+
+    /**
+     * @param vulnerabilityDto dto of new vulnerability
+     */
+    @Transactional
+    fun save(vulnerabilityDto: VulnerabilityDto) {
+        val number = (vulnerabilityRepository.findMaxId() ?: 0) + 1
```
nulls, line 60: do we need to assign the id then?
or maybe save in db with empty\stub name and then update it using ID assigned by database?

Cheshiriks: We need a unique name. But it may not match the id in the database.

nulls: I guess there is a contain that name is unique and we will get an error in case of two requests at same time

Cheshiriks: Okay, I agree. I'll change that in the next PR.

PR #2001, opened 2023-03-18
Several run commands for single demo

### LINKED ISSUE #2000: Several run commands for a single demo
Right now we need to create several projects for a tool to demonstrate both `fix` and `warn` modes. We need to implement adding several run commands with a pretty UI.

sanyavertolet:
This PR closes #2000

### What's done:
 * Added `RunCommand` entity for storing run commands in database
 * Added `RunCommandRepository`
 * Replaces `runCommand` field of `Demo` entity with list of `RunCommand` entities
 * Split `ProjectDemoMenu.kt` file into several files with few functions/components per file
 * Supported several run commands on frontend
 * Removed hardcoded diktat demo from frontend
 * Refactored `demoView`
 * Updated the logic of `Demo` entity save
 * Removed default demo code lines

### How it looks:
https://user-images.githubusercontent.com/35039155/226077751-3c08e053-7b0e-4fb0-8232-7bdac94d0334.mov

### What's next:
 * Remove hardcoded diktat demo
 * Implement `Demo` update
 * Fix frontend issues - wrong SDK and File displaying (Seem to be set to defaults which is not true)
 * Improve stability
 * Add demos for popular static analysis tools

### DIFF
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/cli/DiktatCliRunner.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/cli/DiktatCliRunner.kt
@@ -47,7 +47,7 @@ class DiktatCliRunner(
         append(" --disabled_rules=diktat-ruleset:package-naming,standard")
 
         append(" --reporter=plain,output=$outputPath ")
-        if (demoRunRequest.mode == DemoMode.FIX) {
+        if (demoRunRequest.mode == DemoMode.FIX.name) {
             append(" --format ")
         }
         append(testPath)
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
@@ -104,9 +104,9 @@ class DemoService(
     fun saveIfNotPresent(demo: Demo, runCommands: RunCommandMap): Demo = demoRepository.findByOrganizationNameAndProjectName(demo.organizationName, demo.projectName)?.let {
         throw IllegalStateException("Demo for project ${demo.organizationName}/${demo.projectName} is already added.")
     } ?: run {
-        save(demo).also { savedDemo ->
-            runCommands.map { (mode, command) ->
-                RunCommand(savedDemo, mode, command)
+        save(demo).apply {
+            this.runCommands = runCommands.map { (mode, command) ->
+                RunCommand(this, mode, command)
             }.let { runCommandList ->
                 runCommandRepository.saveAll(runCommandList)
             }
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
@@ -2,8 +2,11 @@ package com.saveourtool.save.demo.service
 
 import com.saveourtool.save.demo.DemoAgentConfig
 import com.saveourtool.save.demo.DemoStatus
+import com.saveourtool.save.demo.RunCommandMap
 import com.saveourtool.save.demo.entity.Demo
+import com.saveourtool.save.demo.entity.RunCommand
 import com.saveourtool.save.demo.repository.DemoRepository
+import com.saveourtool.save.demo.repository.RunCommandRepository
 import com.saveourtool.save.demo.runners.RunnerFactory
 import com.saveourtool.save.demo.storage.DependencyStorage
 import com.saveourtool.save.utils.StringResponse
@@ -26,6 +29,7 @@ class DemoService(
     private val demoRepository: DemoRepository,
     private val kubernetesService: KubernetesService?,
     private val dependencyStorage: DependencyStorage,
+    private val runCommandRepository: RunCommandRepository,
 ) {
     /**
      * Get preferred [RunnerFactory.RunnerType] for demo runner.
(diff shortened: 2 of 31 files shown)


save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/cli/DiktatCliRunner.kt
```diff
@@ -47,7 +47,7 @@ class DiktatCliRunner(
         append(" --disabled_rules=diktat-ruleset:package-naming,standard")
 
         append(" --reporter=plain,output=$outputPath ")
-        if (demoRunRequest.mode == DemoMode.FIX) {
+        if (demoRunRequest.mode == DemoMode.FIX.name) {
```
nulls, line 50: since now `mode` is dynamic, maybe rename `DemoMode` to `DiktatMode`?

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
```diff
@@ -91,20 +95,30 @@ class DemoService(
...
-    fun saveIfNotPresent(demo: Demo): Demo = demoRepository.findByOrganizationNameAndProjectName(demo.organizationName, demo.projectName)?.let {
+    fun saveIfNotPresent(demo: Demo, runCommands: RunCommandMap): Demo = demoRepository.findByOrganizationNameAndProjectName(demo.organizationName, demo.projectName)?.let {
         throw IllegalStateException("Demo for project ${demo.organizationName}/${demo.projectName} is already added.")
-    } ?: save(demo)
+    } ?: run {
+        save(demo).also { savedDemo ->
+            runCommands.map { (mode, command) ->
+                RunCommand(savedDemo, mode, command)
+            }.let { runCommandList ->
+                runCommandRepository.saveAll(runCommandList)
```
nulls, line 111: after saving `RunCommand`, do we need update `savedDemo` which `saveIfNotPresent` returns ?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
```diff
@@ -152,6 +158,10 @@ val projectDemoMenu: FC<ProjectDemoMenuProps> = FC { props ->
         getDemoStatus()
     }
 
+    val (selectedModeCommand, setSelectedModeCommand) = useState<RunCommandPair?>(null)
```
nulls, line 161: I thought you don't like to use `null` in state? 😄

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/BasicRouting.kt
```diff
@@ -92,34 +92,43 @@ val basicRouting: FC<AppProps> = FC { props ->
...
-    val contestGlobalRatingView: FC<Props> = withRouter { location, _ ->
+    val contestGlobalRatingView: VFC = withRouter { location, _ ->
         ContestGlobalRatingView::class.react {
             organizationName = URLSearchParams(location.search).get("organizationName")
             projectName = URLSearchParams(location.search).get("projectName")
             this.location = location
         }
     }
 
+    val demoView: FC<Props> = withRouter { _, params ->
```
nulls, line 123: jFYI: I failed to migrate `withRouter` method to the new version of kotlin wrapper.

nulls (approved):

save-demo/src/main/kotlin/com/saveourtool/save/demo/runners/cli/DiktatCliRunner.kt
sanyavertolet: I guess it will be removed very soon so I see no need in refactoring it right now

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
sanyavertolet: Why do we need to update it? We do not store `RunCommand` ids in `demo` table. AFAIK, each time we fetch `Demo` entity, a separate request is done to fetch `RunCommand` by `demo_id`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/routing/BasicRouting.kt
sanyavertolet: oh sugar, it's gonna be a problem

Seems that we will need to rework existing `withRouter`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/projects/ProjectDemoMenu.kt
sanyavertolet: Here it is done in order to use `selectedModeCommand` as `WindowOpenness`. If it is null, the modal is hidden

I guess it is ok in this way

> Times they are a-changin'

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
nulls: `saveIfNotPresent` return `Demo` which is not valid -- it doesn't contains saved `RunCommand`
You didn't change `@PostMapping("/add")` and result of `saveIfNotPresent` is used as result of this endpoint

sanyavertolet: Oh, got the point

sanyavertolet: Done

PR #2033, opened 2023-03-23
Generate OpenAPI docs by schedule

kgevorkyan:
### What's done:
* Generate OpenAPI docs by schedule: update api spec at 5.30 every day and create PR


.github/workflows/backend-api-spec-pr.yml
```diff
@@ -0,0 +1,28 @@
+name: Create PR for new version of backend api spec
+on:
+  schedule:
+    # triggers the workflow every day at 5:45 UTC
+    - cron: '45 5 * * *'
+
+# FixMe: Properly should be:
+#  push:
+#    branches:
+#      - 'infra/update-backend-api-spec'
```
kgevorkyan, line 10: For some reasons, when bot update this branch via `backend-api-spec-update.yml` (another action), current job weren't trigger
its triggers only after manual push

so in this file I used schedule too 😐

nulls: maybe combine these two action to one?

kgevorkyan: good point, yes. probably we will not soon investigate this problem, so better solution will be to combine them

kgevorkyan: Fixed

Synthetically triggered action created PR successfully https://github.com/saveourtool/save-cloud/pull/2038

.github/workflows/backend-api-spec-update.yml
```diff
@@ -1,18 +1,23 @@
...
 
 jobs:
   update_backend_api_spec:
     runs-on: ubuntu-latest
     steps:
+      - uses: peterjgrainger/action-create-branch@v2.2.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          branch: 'infra/update-backend-api-spec'
```
nulls, line 16: what's about:
```
branch
Optional The name of the branch to create. Default "release-candidate". If your branch conains forward slashes (/) use the full branch reference. Instead of /long/branch/name use refs/heads/long/branch/name. It's an issue with the GitHub API https://gist.github.com/jasonrudolph/10727108
```?

kgevorkyan: I saw it, but it's successfully created without /refs/heads

kgevorkyan: However. I can add it, nevermind

nulls (approved):

PR #2046, opened 2023-03-27
FossGraph. Added progress in createVulnerabilityView

Cheshiriks:
* added progress in createVulnerabilityView
* added modal window for deleting vulnerability

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
@@ -9,6 +9,8 @@ package com.saveourtool.save.frontend.components.views.fossgraph
 import com.saveourtool.save.domain.Role
 import com.saveourtool.save.entities.vulnerability.VulnerabilityDto
 import com.saveourtool.save.entities.vulnerability.VulnerabilityProjectDto
+import com.saveourtool.save.frontend.components.modal.displayModal
+import com.saveourtool.save.frontend.components.modal.mediumTransparentModalStyle
 import com.saveourtool.save.frontend.components.tables.TableProps
 import com.saveourtool.save.frontend.components.tables.columns
 import com.saveourtool.save.frontend.components.tables.tableComponent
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
```diff
@@ -208,8 +227,10 @@ val fossGraph: FC<FossGraphViewProps> = FC { props ->
                 div {
                     className = ClassName("col-xl col-md-6 mb-4")
                     val progress = vulnerability.progress
-                    val color = if (progress < 51) {
+                    val color = if (progress < 34) {
                         Color.GREEN.hexColor
+                    } else if (progress < 67) {
```
sanyavertolet, line 232: Suggestion to move these constants to `private const val`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/progressbar/ReactCircleBuilder.kt
```diff
@@ -11,6 +11,7 @@ import react.react
 enum class Color(val hexColor: String) {
```
sanyavertolet, line 11: Suggestion to make `Color` private if it is used only in this file as we have many similar classes (if I am not mistaken)

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
Cheshiriks: This variable depends on the class variable, it cannot be made into a constant.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/progressbar/ReactCircleBuilder.kt
Cheshiriks: There is no such class. I just think that it can be used in other places in the future

sanyavertolet: Oh ok

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
sanyavertolet: I mean constants like `34` and `67`

PR #2058, opened 2023-03-30
Demo search on frontend

### LINKED ISSUE #2057: Implement demo search on frontend
We need to support an ability to search the demo by organization and project as a part of `DemoMainView`

sanyavertolet:
This PR closes #2057

### What's done:
* Replaced stub `demoList` with functional one
* Added demo filtering
* Implemented `useDebouncedDeferredRequest`

### How it looks:
https://user-images.githubusercontent.com/35039155/228872066-a0409284-e4d6-403e-9c6a-ea9184bc7b6a.mov

### DIFF
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
@@ -19,6 +19,7 @@ import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
+import reactor.kotlin.core.publisher.toMono
 
 import reactor.kotlin.core.util.function.component1
 import reactor.kotlin.core.util.function.component2
@@ -43,14 +44,15 @@ class DemoController(
      */
     @PostMapping("/demo-list")
     fun getFilteredDemoList(
-        @RequestBody(required = false) filter: DemoFilter = DemoFilter.any,
+        @RequestBody(required = false) filter: DemoFilter?,
         @RequestParam(required = false, defaultValue = DemoDto.DEFAULT_FETCH_NUMBER.toString())
         demoAmount: Int = DemoDto.DEFAULT_FETCH_NUMBER,
-    ): Flux<DemoDto> = blockingToFlux {
-        demoService.getFiltered(filter, demoAmount).toList().map { it to demoService.getStatus(it).block() }
-    }
-        .filter { (_, status) -> status in filter.statuses }
-        .map { it.first.toDto() }
+    ): Flux<DemoDto> = filter.toMono()
+        .switchIfEmpty(DemoFilter.any.toMono())
+        .flatMapMany { demoFilter ->
+            blockingToFlux { demoService.getFiltered(demoFilter, demoAmount) }
+        }
+        .map { it.toDto() }
 
     /**
      * @param organizationName saveourtool organization name
--- a/save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
+++ b/save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
@@ -4,9 +4,9 @@ import com.saveourtool.save.configs.ApiSwaggerSupport
 import com.saveourtool.save.demo.DemoDto
 import com.saveourtool.save.demo.DemoResult
 import com.saveourtool.save.demo.DemoRunRequest
-import com.saveourtool.save.demo.DemoStatus
 import com.saveourtool.save.demo.runners.RunnerFactory
 import com.saveourtool.save.demo.service.DemoService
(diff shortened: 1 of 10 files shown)


save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
```diff
@@ -88,9 +91,29 @@ class DemoService(
...
-     * @return list of [Demo]s that are stored in database
+     * @param demoFilter [DemoFilter]
+     * @param pageSize amount of [Demo]s that should be fetched
+     * @return list of [Demo]s that match [DemoFilter]
      */
-    fun getAllDemos(): List<Demo> = demoRepository.findAll()
+    fun getFiltered(demoFilter: DemoFilter, pageSize: Int): Page<Demo> = demoRepository.findAll({ root, _, cb ->
+        with(demoFilter) {
+            val organizationNamePredicate = if (organizationName.isBlank()) {
+                cb.and()
```
nulls, line 101: what's about status?

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
```diff
@@ -131,7 +154,6 @@ class DemoService(
      * @param projectName saveourtool project name
      * @return [Demo] connected with project [organizationName]/[projectName] or null if not present
      */
-    @Transactional
```
nulls, line 134: or `@Transactional(readonly=true)`?

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
```diff
@@ -36,13 +37,19 @@ class DemoController(
...
+     * @param demoAmount number of [DemoDto]s that should be fetched, [DemoDto.DEFAULT_FETCH_NUMBER] by default
+     * @return all [DemoDto]s matching [filter] as [Flux]
      */
-    @GetMapping("/active")
-    fun active(): Flux<DemoDto> = blockingToFlux {
-        demoService.getAllDemos().map { it to demoService.getStatus(it).block() }
+    @PostMapping("/demo-list")
+    fun getFilteredDemoList(
+        @RequestBody(required = false) filter: DemoFilter = DemoFilter.any,
+        @RequestParam(required = false, defaultValue = DemoDto.DEFAULT_FETCH_NUMBER.toString())
```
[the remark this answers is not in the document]
sanyavertolet: It seems that the [issue](https://github.com/spring-projects/spring-framework/issues/21139) is still opened so will make `filter` nullable and perform null-check

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
sanyavertolet: Right now I check statuses after the filtered demo is fetched as I do not store the status in database due to no sync solution (right now there is no way to keep data up to date, need to implement some kind of service that does this task repeatedly over time). If you find it better, I could move status filtering to this method, but the point is that status check right now is an expensive operation

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
sanyavertolet: Probably 😃

sanyavertolet: The point is that Idea gives me an error:
```
@Transactional self-invocation (in effect, a method within the target object calling another method of the target object) does not lead to an actual transaction at runtime 
```
 on `findBySaveourtoolProject` invocation in method:
```Kotlin
fun findBySaveourtoolProjectOrNotFound(
    organizationName: String,
    projectName: String,
    lazyMessage: () -> String,
): Mono<Demo> = blockingToMono {
    findBySaveourtoolProject(organizationName, projectName)
}.switchIfEmptyToNotFound(lazyMessage)
```

Cheshiriks: You cannot call a transactional method from another method in the same class

sanyavertolet: Discussed offline - decided to ignore the warning 😁

save-demo/src/main/kotlin/com/saveourtool/save/demo/service/DemoService.kt
sanyavertolet: Will fix that in the next PR.

save-demo/src/main/kotlin/com/saveourtool/save/demo/controller/DemoController.kt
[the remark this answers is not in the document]
sanyavertolet: Done

nulls (approved):

PR #2070, opened 2023-04-07
Fixed avatar in topBar

### LINKED ISSUE #2048: Profile avatar is not displayed in toolbar
Avatar is not displayed on the top of the screen, near the login, while it is expected to be.
screenshot

Cheshiriks:
Closes #2048

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarUserField.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarUserField.kt
@@ -5,6 +5,7 @@ package com.saveourtool.save.frontend.components.topbar
 import com.saveourtool.save.domain.Role
 import com.saveourtool.save.frontend.components.modal.logoutModal
 import com.saveourtool.save.frontend.externals.fontawesome.*
+import com.saveourtool.save.frontend.utils.*
 import com.saveourtool.save.frontend.utils.AVATAR_PLACEHOLDER
 import com.saveourtool.save.info.UserInfo
 import com.saveourtool.save.v1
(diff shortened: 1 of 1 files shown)


nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarUserField.kt
```diff
@@ -97,7 +98,7 @@ private fun topBarUserField() = FC<TopBarUserFieldProps> { props ->
                             }
                         }
                     }
-                    avatar?.let { avatar ->
+                    props.userInfo?.avatar?.let {
```
sanyavertolet, line 101: Here you use avatar from `props`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarUserField.kt
```diff
@@ -53,7 +54,7 @@ private fun topBarUserField() = FC<TopBarUserFieldProps> { props ->
     val navigate = useNavigate()
     var isLogoutModalOpen by useState(false)
     var isAriaExpanded by useState(false)
-    val (avatar, setAvatar) = useState(props.userInfo?.avatar?.let { "/api/$v1/avatar$it" })
+    val (avatar, setAvatar) = useStateFromProps("/api/$v1/avatar${props.userInfo?.avatar}")
```
sanyavertolet, line 57: Here you use avatar from state

sanyavertolet (commented):
Why?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarUserField.kt
Cheshiriks: No. I use avatar from state, but state is nullable and contains the full link. here I just check that the avatar exists

sanyavertolet: Oh sure yes, by bad

sanyavertolet (approved):

PR #2090, opened 2023-04-17
Support latest kotlin wrapper

nulls:


akuleshov7 (approved):
I don’t know if it was really needed, but looks good

PR #2093, opened 2023-04-19
FossGraph. Added the ability to comment

### LINKED ISSUE #2092: FossGraph. Add the ability to comment
Add the ability to comment

Cheshiriks:
Added the ability to comment

comment

Closes #2092


db/v-2/tables/comments.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="comments-1" author="frolov" context="dev or prod">
```
kgevorkyan, line 8: why `comments-1`? do you plan to add something else?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/CommentDto.kt
```diff
@@ -0,0 +1,25 @@
...
+ * @property userName
+ * @property userAvatar
+ * @property section
+ */
+@Serializable
+data class CommentDto(
+    val message: String,
+    val userName: String,
+    val userAvatar: String?,
+    val section: String = "",
```
kgevorkyan, line 16: what section will be used by default? If here we set ""?

kgevorkyan: probably its better to document it in kdoc

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Card.kt
```diff
@@ -31,21 +34,29 @@ external interface CardProps : PropsWithChildren {
...
 ) = FC<CardProps> { props ->
     val boarder = if (isBordered) "border-secondary" else ""
     val card = if (hasBg) "card" else ""
     val pb = if (isPaddingBottomNull) "pb-0" else ""
     val paddingInside = if (isNoPadding) "pt-0 pr-0 pl-0" else ""
     div {
         className = ClassName("$card card-body mt-0 $paddingInside $pb $boarder")
+        if (isFilling) {
+            style = jso {
+                width = "100%".unsafeCast<Width>()
```
kgevorkyan, line 56: isn't this the default value? 100%

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/Comment.kt
```diff
@@ -0,0 +1,32 @@
...
+    var section: String,
+
+    @ManyToOne
+    @JoinColumn(name = "user_id")
+    var user: User,
+
+) : BaseEntityWithDateAndDto<CommentDto>() {
+    override fun toDto() = CommentDto(
+        message = message,
+        userName = user.name ?: "Unknown",
```
kgevorkyan, line 29: how we can get user.name == null?

Cheshiriks: The _name_ field in the _User_ entity is nullable. This situation should not happen, but I think it's better to do a check than to put the operator !!

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Card.kt
Cheshiriks: The width is default, but the height is not. Here I explicitly registered it, so that it is clear.

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/CommentDto.kt
Cheshiriks: The backend will check that the Section field is filled. This field in the DTO is only needed to save a comment

db/v-2/tables/comments.xml
Cheshiriks: I always add a number in the ID Change Sets in case there are new changes in the future

kgevorkyan (approved):

PR #2097, opened 2023-04-21
Creating Role and RoleBinding in agent's namespace

nulls:
### What's done:
- create a single role for orchestrator, sandbox and demo
- added namespace specification for role binding
- added network-policy from `save-agent` to `backend`


save-cloud-charts/save-cloud/templates/orchestrator-service-account.yaml
```diff
@@ -9,7 +9,13 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: jobs-executor
+  {{ if .Values.agentNamespace }}
+  namespace: {{ .Values.agentNamespace }}
```
sanyavertolet, line 13: Does this namespace mean that pods matching selector have a role in another namespace or that pods from this namespace matching selector have a role on cluster level?

save-cloud-charts/save-cloud/templates/orchestrator-service-account.yaml
```diff
@@ -18,6 +24,9 @@ rules:
     verbs: [list, get, delete, create]
   # this rule is required for sprint-cloud-kubernetes-fabric8-config
   - apiGroups: [""]  # "" indicates the core API group
+    resources: [pods/portforward]
```
sanyavertolet, line 27: Guess it can be removed as it is not used anymore

nulls: sure

save-cloud-charts/save-cloud/templates/orchestrator-service-account.yaml
nulls: > matching selector have a role in another namespace
this one

Role and RoleBinding should be in the same namespace, since it did work previously when role and agent's pod were in the same namespace, I moved role to agent's namespace

sanyavertolet: oh ok

PR #2106, opened 2023-04-25
Fix local run for orchestrator

nulls:
### What's done:
- added a new AgentEnvName.KUBERNETES
- set as true for KubernetesManager and false for DockerContainerRunner (local mode)

### DIFF
--- a/save-agent/src/commonMain/kotlin/com/saveourtool/save/agent/AgentConfiguration.kt
+++ b/save-agent/src/commonMain/kotlin/com/saveourtool/save/agent/AgentConfiguration.kt
@@ -9,6 +9,7 @@ import com.saveourtool.save.agent.utils.TEST_SUITES_DIR_NAME
 import com.saveourtool.save.core.config.LogType
 import com.saveourtool.save.core.config.OutputStreamType
 import com.saveourtool.save.core.config.ReportType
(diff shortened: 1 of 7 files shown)


sanyavertolet (approved):

.github/workflows/build_and_test.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         with:
           path: ~/.konan
           key: ${{ runner.os }}-gradle-konan-${{ env.KOTLIN_VERSION }}
-      - name: Build all (excluding save-orchestrator-common, save-orchestrator and save-sandbox)
+      - name: Build all (excluding save-orchestrator-common, save-orchestrator, save-sandbox, save-backend and save-frontend)
```
kgevorkyan, line 43: may be now it's easy to say what do we build exactly, instead of `all excluding a,b,c,d,e,f,g,...` 😁

kgevorkyan (approved):

.github/workflows/build_and_test.yml
nulls: but still, it doesn't help. I see now only that frontend fails

nulls: refactored to build some specific modules at first and left at the end

PR #2112, opened 2023-04-26
Dependency in building

nulls:
### What's done:
- build save-cloud-common and save-orchestrator-common at the first step
- push codecov as a single artifact


.github/workflows/build_and_test.yml
```diff
@@ -14,19 +14,44 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build_and_test_with_code_coverage_module:
-    name: Build, test and upload code coverage by module
-    runs-on: ubuntu-latest
+  build_save-cloud-common:
```
kgevorkyan, line 17: could it be more compact via some matrix?

where matrix is `[save-cloud-common, save-orchestrator-common, ...]`

kgevorkyan (approved):

.github/workflows/build_and_test.yml
nulls: I need to check input.<module_name> to run building of a specific module.
I'm not sure GitHub allows to dynamic variable to get some value from `inputs`, but I can construct a martix dynamically.
Will try to refactor

But still -- backend and sandbox will be separated because I need to specify a dependency to save-cli artifact

nulls: done

nulls: oh..missed that you asked about `build_and_test` -- common modules should be built one by one (each next module takes cache from previous one and updates it). It needs to allow to use caches in rest steps.

I added a similar approach with matrix (like for left modules) in deploy_images.yml: There backend and sandbox are built with dependency to save-cli and rest are built independently

Cheshiriks (approved):

PR #2123, opened 2023-04-27
Fix backend-api-spec-update.yml

nulls:
### What's done:
- move pull master before generating OpenAPI
- removed unused parameter

### DIFF
--- a/.github/workflows/backend-api-spec-update.yml
+++ b/.github/workflows/backend-api-spec-update.yml
@@ -5,12 +5,6 @@ on:
     # triggers the workflow every day at 5:30 UTC
     - cron: '30 5 * * *'
   workflow_dispatch:
-    inputs:
-      branch:
-        type: string
-        default: master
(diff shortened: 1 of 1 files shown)


nulls:
Testing; https://github.com/saveourtool/save-cloud/actions/runs/4820662913

.github/workflows/backend-api-spec-update.yml
```diff
@@ -5,12 +5,6 @@ on:
     # triggers the workflow every day at 5:30 UTC
     - cron: '30 5 * * *'
   workflow_dispatch:
-    inputs:
```
kgevorkyan, line 8: Hm, I thought it's required, good

kgevorkyan (approved):

.github/workflows/backend-api-spec-update.yml
nulls: tested, it does work without inputs

PR #2127, opened 2023-04-28
Fix `backend-api-spec-update.yml`

nulls:
### What's done:
- set fetch-depth: 0 <-- this fixes pull issue
- moved git configuration to earlier step <-- author is required to rebase\merge changes from master

---
Not tested case:
1. we have opened PR
2. we changes some endpoint
3. the workflow regenerates `backend-api-docs.json`
4. the workflow tries to create PR but it's already existed...

### DIFF
--- a/.github/workflows/backend-api-spec-update.yml
+++ b/.github/workflows/backend-api-spec-update.yml
@@ -19,6 +19,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: 'infra/update-backend-api-spec'
+          fetch-depth: 0
 
       - name: Set up Java
         uses: actions/setup-java@v3
--- a/.github/workflows/backend-api-spec-update.yml
+++ b/.github/workflows/backend-api-spec-update.yml
@@ -26,15 +26,20 @@ jobs:
         with:
           java-version: 17
           distribution: temurin
+
+      - name: Set up Git
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+
       - name: Install system packages
         # libcurl is needed for ktor-client-curl
         run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
 
       - name: Update branch from master
         run: |
-          git log
-          git status
           git pull origin master --no-edit --rebase
+
       - name: Generate open api doc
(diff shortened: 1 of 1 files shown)


.github/workflows/backend-api-spec-update.yml
```diff
@@ -19,21 +19,27 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: 'infra/update-backend-api-spec'
+          fetch-depth: 0
```
kgevorkyan, line 22: could you add comment about this code

.github/workflows/backend-api-spec-update.yml
```diff
@@ -19,21 +19,27 @@ jobs:
...
 
       - name: Set up Java
         uses: actions/setup-java@v3
         with:
           java-version: 17
           distribution: temurin
+
+      - name: Set up Git
```
kgevorkyan, line 30: how did it worked previously without this 🤦‍♂️

kgevorkyan (approved):

kgevorkyan:
> 4. the workflow tries to create PR but it's already existed...

I think in worst case, this step will just fail, but ``git pull` is first, so all should be okay at the end

.github/workflows/backend-api-spec-update.yml
nulls: I'm not sure, maybe `infra/update-backend-api-spec` doesn't have any changes\commits during git pull

nulls:
> > 4. the workflow tries to create PR but it's already existed...
> 
> I think in worst case, this step will just fail, but `git pull` is first, so all should be okay at the end

yeap, but it breaks automatization, but will return to this question if it will gets more problems.

Note: automatization is still broken, since PR and commits are done by bot user and builds are not triggered automatically

.github/workflows/backend-api-spec-update.yml
nulls, line 22:
```suggestion
          fetch-depth: 0 # we need a full history to make rebase with master
```

nulls: done

kgevorkyan: good idea in aim to do fast commit via github 😀

PR #2149, opened 2023-05-15
For non-authorised users OrganizationView not working

### LINKED ISSUE #2148: For non-authorised users OrganizationView is absolutely not working
1) Tools are not loading
2) Tests are not loading
3) Users are not shown, avatar is not shown

https://github.com/saveourtool/save-cloud/assets/58667063/f98b6de4-5851-4038-bd29-18cee9e593cb

Cheshiriks:
For non-authorised users OrganizationView is absolutely not working

Closes #2148


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -109,16 +107,18 @@ class LnkUserOrganizationController(
...
+        getUserAndOrganizationWithPermissions(
+            userName ?: authentication.toUser().name!!,
+            organizationName,
+            Permission.READ,
+            authentication,
+        )
+            .map { (user, organization) ->
+                lnkUserOrganizationService.getRole(user, organization)
+            }
+    } ?: Role.NONE.toMono()
```
kgevorkyan, line 121: Do we have role viewer?

Cheshiriks: Viewer role for authorized users who are not members of this organization.

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -109,16 +107,18 @@ class LnkUserOrganizationController(
...
-        Permission.READ,
-        authentication,
-    )
-        .map { (user, organization) ->
-            lnkUserOrganizationService.getRole(user, organization)
-        }
+        authentication: Authentication?,
+    ): Mono<Role> = authentication?.let {
+        getUserAndOrganizationWithPermissions(
+            userName ?: authentication.toUser().name!!,
```
nulls, line 113: `userName` can be provided but `authentication` can be `null` and it still leads to `Role.NONE`
```suggestion
        authentication: Authentication?,
    ): Mono<Role> = (userName ?: authentication?.toUser()?.name)?.let { _userName ->
        getUserAndOrganizationWithPermissions(
            _userName,
            organizationName,
            Permission.READ,
            authentication,
        )
            .map { (user, organization) ->
                lnkUserOrganizationService.getRole(user, organization)
            }
    } ?: Role.NONE.toMono()
```

Cheshiriks: Can we have a UserName for non-authenticated users? After all, we take this name just from the database after authentication

Cheshiriks: userName param is _required = false_. Somewhere in the code we use userName to call the controller, but somewhere we don’t pass anything and use authentication. I don't know why it was implemented like this. But if you remove _authentication_, then the logic will break.

Cheshiriks: Removed UserName

nulls (approved):

PR #2155, opened 2023-05-19
Added a view to create templates for future contests

### LINKED ISSUE #2154: Add a view to create templates for future contests
Users should be able to create their own custom contests.

Cheshiriks:
- added Create contest template view
- added button to go to the view

Closes #2154

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/contest/ContestSampleDto.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/contest/ContestSampleDto.kt
@@ -5,11 +5,13 @@ import kotlinx.serialization.Serializable
 /**
  * @property name name of contest sample
  * @property description description of contest sample
+ * @property fields list of fields
  */
 @Serializable
 data class ContestSampleDto(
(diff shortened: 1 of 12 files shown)


kgevorkyan (commented):
Why do we need to show the types of name, description and so on, on view?

Cheshiriks:
> Why do we need to show the types of name, description and so on, on view?

On this view, the user specifies which fields to fill in future contests and chooses what type they should be.

kgevorkyan:
> Member

I think
`name` = `str`
`descr` = `text`
should be enough, isn't it?

if not, may be we can create aliases on frontend? pure types like `MULTI_STRING` are very ugly

and by the way, why is critial = `num`? shouldn't it be `bool`?

Cheshiriks:
> > Member
> 
> I think `name` = `str` `descr` = `text` should be enough, isn't it?
> 
> if not, may be we can create aliases on frontend? pure types like `MULTI_STRING` are very ugly
> 
> and by the way, why is critial = `num`? shouldn't it be `bool`?

Added pretty name

kgevorkyan (approved):

PR #2203, opened 2023-06-07
OrganizationView unauthorized fix

sanyavertolet:
### What's done:
* Removed `authentication` parameter from `/api/$v1/organizations/{organizationName}/users` endpoint

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
@@ -60,8 +60,6 @@ class LnkUserOrganizationController(
     private val organizationPermissionEvaluator: OrganizationPermissionEvaluator,
 ) {
     @GetMapping("/{organizationName}/users")
-    @RequiresAuthorizationSourceHeader
-    @PreAuthorize("permitAll()")
     @Operation(
         method = "GET",
         summary = "Get list of users that are connected with given organization.",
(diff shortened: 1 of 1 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -74,7 +72,7 @@ class LnkUserOrganizationController(
     @ApiResponse(responseCode = "404", description = "Contest with such name was not found.")
     fun getAllUsersByOrganizationName(
         @PathVariable organizationName: String,
-        authentication: Authentication,
+        authentication: Authentication?,
```
nulls, line 75: do we need this property in this method?

sanyavertolet: Actually no, we don't. Will remove it

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
```diff
@@ -74,7 +72,6 @@ class LnkUserOrganizationController(
     @ApiResponse(responseCode = "404", description = "Contest with such name was not found.")
     fun getAllUsersByOrganizationName(
         @PathVariable organizationName: String,
-        authentication: Authentication,
```
Cheshiriks, line 77: Won't an error be thrown when a person enters not authorized?

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/LnkUserOrganizationController.kt
Cheshiriks: already fixed :)

Cheshiriks (approved):

PR #2214, opened 2023-06-14
Downgraded kotlinx-coroutines version

nulls:
### What's done:
- downgraded kotlinx-coroutines version to 1.6.4 is due to https://github.com/ktorio/ktor/pull/3609 is not released yet
- upgraded ktor version to 2.3.1


Cheshiriks (approved):

.github/workflows/build_and_test.yml
```diff
@@ -41,6 +41,26 @@ jobs:
...
+      - id: save-save-cli-version
+        name: Store save-cli version in output
+        run: |
+          sv=$(cat gradle/libs.versions.toml | grep -m1 '^save-cli =' | awk -F'[=]' '{print $2}' | tr -d '" ')
+          echo "Detected version $sv"
+          echo version=$sv >> $GITHUB_OUTPUT
+  build_cli:
+    name: Build save-cli
+    needs: [ retrieve_save_cli_version ]
+    uses: ./.github/workflows/build_save-cli_reusable.yml
```
kgevorkyan, line 61: do we really need the first `.`?

.github/workflows/build_and_test.yml
```diff
@@ -41,6 +41,26 @@ jobs:
...
       gradle-cache-read-only: ${{ github.ref != 'refs/heads/master' && github.event_name != 'pull_request' }}
+  retrieve_save_cli_version:
+    name: Retrieve save-cli version from save-cloud
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.save-save-cli-version.outputs.version }}
+    steps:
+      - name: checkout save-cloud
+        uses: actions/checkout@v3
+      - id: save-save-cli-version
```
kgevorkyan, line 52:
```suggestion
      - id: save-cli-version
```

save-demo-cpg/build.gradle.kts
```diff
@@ -19,6 +19,12 @@ repositories {
             artifact("/[organisation].[module]_[revision].[ext]")
         }
     }
+    maven {
+        url = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
kgevorkyan, line 23: its not clear for me
we didn't used snapshot version prevoisly?

.github/workflows/build_and_test.yml
```diff
@@ -41,6 +41,26 @@ jobs:
     with:
       module: save-demo-agent
       gradle-cache-read-only: ${{ github.ref != 'refs/heads/master' && github.event_name != 'pull_request' }}
+  retrieve_save_cli_version:
```
kgevorkyan, line 44: do we removed all previous logic, corresponding for save-cli snapshots?

gradle/libs.versions.toml
```diff
@@ -2,11 +2,11 @@
 kotlin = "1.8.22"
 jetbrains-annotations = "24.0.1"
 save-cli = "0.4.0-SNAPSHOT"
-ktor = "2.3.0"
+ktor = "2.3.1"
 okio = "3.3.0"
 serialization = "1.5.1"
 kotlinx-datetime = "0.4.0"
-kotlinx-coroutines = "1.7.1"
+kotlinx-coroutines = "1.6.4"
```
sanyavertolet, line 9: And for slow ones - why do we. downgrade it?

sanyavertolet (approved):

gradle/libs.versions.toml
sanyavertolet: oh, got it. Main PR is merged, but not released

save-demo-cpg/build.gradle.kts
nulls: we used, but removed when migrated to release version. also `save-demo-cpg` has overrides for repositories. that why we need to duplicate configuration here

.github/workflows/build_and_test.yml
nulls: yes, we refactored it during splitting save-cloud build: we had problem with long running jib on GitHub and split them to small jobs. We still have logic for save-cli in building images, but removed in regular build

.github/workflows/build_and_test.yml
nulls: taken from [documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow): 
```
You reference reusable workflow files using one of the following syntaxes:

- {owner}/{repo}/.github/workflows/{filename}@{ref} for reusable workflows in public and private repositories.
- ./.github/workflows/{filename} for reusable workflows in the same repository
```

But, I didn't test without leading `.`

.github/workflows/build_and_test.yml
nulls: done

.github/workflows/build_and_test_reusable.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         uses: gradle/gradle-build-action@v2
         env:
           SKIP_SAVE_AGENT_DEPENDENCY: true
+          SKIP_SAVE_CLI_DEPENDENCY: true
```
nulls, line 45: actually, we don't need to have save-cli kexe to build and run tests

PR #2219, opened 2023-06-15
cytoscape.js

sanyavertolet:
### What's done:
 * Added `cytoscape` wrapper
 * Reused `cytoscape` wrapper with `cytoscapeVisualizer` in order to replace `graphVizVisualizer`

### TODO:
 * Fix visuals
 * Make all selectors on `DemoCpgView` work

### How it looks:
screenshot


kgevorkyan:
looks ugly - Andrey K

akuleshov7:
is there any chance for us to have the same as NEO4J has? Where you can move nodes with a mouse?

sanyavertolet:
> is there any chance for us to have the same as NEO4J has?

I hope yes. What we have now is just a first version. Styles are not properly set up, some wanted features are missing, but at least it works. There will be PRs to fix that and make it prettier, but I cannot guarantee that it will be as good as `Neo4j`.

> Where you can move nodes with a mouse?

If I am not mistaken you already are able to move nodes with a mouse.

PR #2220, opened 2023-06-15
Added a top rating view

Cheshiriks:
- added a top rating view

Rating

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/OrganizationRatingTab.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/OrganizationRatingTab.kt
@@ -2,33 +2,41 @@
 
 package com.saveourtool.save.frontend.components.views.toprating
 
+import com.saveourtool.save.entities.OrganizationDto
+import com.saveourtool.save.filters.OrganizationFilter
+import com.saveourtool.save.frontend.components.basic.nameFiltersRow
 import com.saveourtool.save.frontend.components.tables.*
 import com.saveourtool.save.frontend.externals.fontawesome.faTrophy
 import com.saveourtool.save.frontend.externals.fontawesome.fontAwesomeIcon
 import com.saveourtool.save.frontend.utils.*
 import com.saveourtool.save.frontend.utils.noopLoadingHandler
 import com.saveourtool.save.frontend.utils.noopResponseHandler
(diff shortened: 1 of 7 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/OrganizationRatingTab.kt
```diff
@@ -0,0 +1,151 @@
...
+import react.dom.html.ReactHTML.th
+import react.dom.html.ReactHTML.tr
+import react.router.dom.Link
+import web.cssom.*
+
+import kotlinx.browser.window
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+val renderOrganizationRatingTab: FC<Props> = FC { _ ->
```
sanyavertolet, line 30: To my mind, `render` here is odd. Should be renamed to `organizationRatingTab`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/TopRatingView.kt
```diff
@@ -0,0 +1,61 @@
...
+            title(" Top Rating", faTrophy)
+        }
+    }
+
+    tab(selectedMenu.name, TopRatingTab.values().map { it.name }, "nav nav-tabs mt-3") { value ->
+        setSelectedMenu(TopRatingTab.valueOf(value))
+    }
+
+    when (selectedMenu) {
+        TopRatingTab.USERS -> renderUserRatingTable()
```
sanyavertolet, line 40: Here also there is no need to add `render` as it is not a function that renders, but a `val` that is a tab

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/OrganizationRatingTab.kt
Cheshiriks: Ok. Renamed it :)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/toprating/TopRatingView.kt
Cheshiriks: Fixed

PR #2226, opened 2023-06-20
Fixed user profile props and size of Vuln archive in top bar

Cheshiriks:
- fixed user profile props
- fixed size of Vuln archive in top bar


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
```diff
@@ -57,7 +57,7 @@ private fun topBarLinks() = FC<TopBarLinksProps> { props ->
         sequenceOf(
             TopBarLink(hrefAnchor = FrontendRoutes.DEMO.path, width = 4.rem, text = "Demo"),
             TopBarLink(hrefAnchor = "${FrontendRoutes.DEMO.path}/cpg", width = 3.rem, text = "CPG"),
-            TopBarLink(hrefAnchor = FrontendRoutes.VULNERABILITIES.path, width = 6.rem, text = "Vuln. Archive"),
```
sanyavertolet, line 60: Our topbar becomes too huge...

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/topbar/TopBarLinks.kt
Cheshiriks: Yes. It will be necessary to think about how it can be changed.

PR #2233, opened 2023-06-22
Added vulnerability identifier

Cheshiriks:
- added field Identifier in vulnerability
- added field Identifier on vulnerability view
- added check that a vulnerability with this ID does not yet exist

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
@@ -357,11 +357,7 @@ val fossGraph: FC<FossGraphViewProps> = FC { props ->
                         div {
                             className = ClassName("row")
                             h6 {
-                                className = ClassName("m-0 font-weight-bold text-primary")
-                                style = jso {
-                                    display = Display.flex
-                                    alignItems = AlignItems.center
-                                }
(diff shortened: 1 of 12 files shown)


akuleshov7:
where is a field for CVE? Can you pls add a full screenshot?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
```diff
@@ -352,7 +352,26 @@ val fossGraph: FC<FossGraphViewProps> = FC { props ->
...
-
+                    div {
+                        className = ClassName("card-header py-3")
+                        div {
+                            className = ClassName("row")
+                            h6 {
+                                className = ClassName("m-0 font-weight-bold text-primary")
+                                style = jso {
+                                    display = Display.flex
+                                    alignItems = AlignItems.center
```
sanyavertolet, line 363: You can use classnames for this purpose

sanyavertolet (approved):

Cheshiriks:
> where is a field for CVE? Can you pls add a full screenshot?

Added

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/FossGraphView.kt
Cheshiriks: Fixed

sanyavertolet:
Why is input with CVE displayed as `success` instead of `danger`?

PR #2235, opened 2023-06-22
Making flexible size of the text

akuleshov7:
### What's done:
- very small change to make views more flexible
- it appeared that we can be very flexible on all windows just with a small css change (font size in "vw")

https://github.com/saveourtool/save-cloud/assets/58667063/03634c9a-6149-4780-b8ac-70d31daf5973


sanyavertolet (approved):
It is very painful to see that the heights of two cards differ

Cheshiriks (approved):

PR #2247, opened 2023-06-28
Added vulnerability comments

Cheshiriks:
- added tab for vulnerability comments

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Forum.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Forum.kt
@@ -1,4 +1,4 @@
-@file:Suppress("FILE_NAME_MATCH_CLASS")
+@file:Suppress("FILE_NAME_MATCH_CLASS", "HEADER_MISSING_IN_NON_SINGLE_CLASS_FILE")
 
 package com.saveourtool.save.frontend.components.basic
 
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Forum.kt
```diff
@@ -36,10 +36,20 @@ external interface CommentWindowProps : PropsWithChildren {
...
+     * Callback invoked when added new comment
+     */
+    var addComment: () -> Unit
+}
+
 /**
  * @return a function component
  */
-fun newCommentWindow() = FC<PropsWithChildren> {
+fun newCommentWindow() = FC<NewCommentWindowProps> { props ->
```
sanyavertolet, line 52: should be `val` instead of `fun`

sanyavertolet (approved):
Visuals should be reworked. Let's leave it as is for this PR in order to add functionality, but the next PR should improve styles

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Forum.kt
Cheshiriks: Fixed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/Forum.kt
```diff
@@ -36,10 +36,20 @@ external interface CommentWindowProps : PropsWithChildren {
...
+     * Callback invoked when added new comment
+     */
+    var addComment: () -> Unit
+}
+
 /**
  * @return a function component
  */
-fun newCommentWindow() = FC<PropsWithChildren> {
+val newCommentWindow = FC<NewCommentWindowProps> { props ->
```
sanyavertolet, line 52:
```suggestion
val newCommentWindow: FC<NewCommentWindowProps> = FC { props ->
```

Cheshiriks: Fixed

akuleshov7 (approved):

kgevorkyan:
> Visuals should be reworked. Let's leave it as is for this PR in order to add functionality, but the next PR should improve styles

create the issue please

PR #2252, opened 2023-06-28
Added rating in user comment

Cheshiriks:
- added rating in user comment

11Снимок


sanyavertolet (approved):
To my mind, rating label should not be bigger than the message itself

PR #2268, opened 2023-07-03
Added link users to vulnerability

Cheshiriks:
- added entity for LnkVulnerabilityUser
- added controller for users to vulnerability


db/v-2/tables/lnk-vulnerability-user.xml
```diff
@@ -0,0 +1,26 @@
...
+                <constraints foreignKeyName="fk_vulnerability_user_vulnerability" references="vulnerability(id)" nullable="false" deleteCascade="true"/>
+            </column>
+            <column name="user_id" type="bigint">
+                <constraints foreignKeyName="fk_lnk_vulnerability_user_user" references="user(id)" nullable="false" deleteCascade="true"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet id="lnk-vulnerability-user-2" author="frolov">
+        <addUniqueConstraint tableName="lnk_vulnerability_user" columnNames="user_id, vulnerability_id"/>
```
nulls, line 23: can it be moved to `changeSet id="lnk-vulnerability-user-1"`?

Cheshiriks: Yes. changed it

sanyavertolet (approved):

nulls (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/vulnerabilities/LnkVulnerabilityUser.kt
```diff
@@ -0,0 +1,20 @@
...
+import javax.persistence.*
+
+/**
+ * @property vulnerabilityId
+ * @property user in vulnerability with [vulnerabilityId]
+ */
+@Entity
+class LnkVulnerabilityUser(
+
+    var vulnerabilityId: Long,
```
nulls, line 14: why it was refactored to a single id?

Cheshiriks: Both the Vulnerability and the User entities contain fields with the ManyToOne type. The entity LinkVulnerabilityUser will then contain two such fields, but in Hibernate this is prohibited and throw an error.

Cheshiriks: Yes
_Fetching two or more Bags at the same time on an Entity could form a Cartesian Product. Since a Bag doesn't have an order, Hibernate would not be able to map the right columns to the right entities. Hence, in this case, it throws a MultipleBagFetchException._

https://www.baeldung.com/java-hibernate-multiplebagfetchexception#cause-of-multiplebagfetchexception

nulls: I meant in code :)

Cheshiriks: Ok, adding now in the next pr )

save-backend/src/test/kotlin/com/saveourtool/save/backend/controllers/RunExecutionControllerTest.kt
```diff
@@ -49,6 +51,7 @@ class RunExecutionControllerTest(
     @Autowired private lateinit var testExecutionRepository: TestExecutionRepository
     @Autowired private lateinit var lnkExecutionFileRepository: LnkExecutionFileRepository
     @Autowired private lateinit var lnkExecutionTestSuiteRepository: LnkExecutionTestSuiteRepository
+    @MockBean private lateinit var lnkVulnerabilityUserRepository: LnkVulnerabilityUserRepository
```
nulls, line 54: if you don't use this property, please move this bean to annotation section only:
```
@MockBeans(
    MockBean(LnkVulnerabilityUserRepository::class),
)
```

PR #2269, opened 2023-07-03
vulnerabilityInfoTab should rely on currentUserInfo

sanyavertolet:
### What's done:
 * Implemented passing currentUserInfo to vulnerabilityInfoTab using props
 * Hid several buttons if user is unauthorized

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -255,8 +255,8 @@ class VulnerabilityService(
     @Transactional
     fun saveAllProjects(
         vulnerabilityProjectDtos: List<VulnerabilityProjectDto>
-    ): VulnerabilityDto? = vulnerabilityRepository.findByName(vulnerabilityProjectDtos.first().vulnerabilityName)?.let { vulnerability ->
-        vulnerabilityProjectRepository.saveAll(
+    ): List<VulnerabilityProjectDto>? = vulnerabilityRepository.findByName(vulnerabilityProjectDtos.first().vulnerabilityName)?.let { vulnerability ->
+        return@let vulnerabilityProjectRepository.saveAll(
             vulnerabilityProjectDtos.map { dto ->
                 VulnerabilityProject(
                     name = dto.name,
@@ -268,9 +268,6 @@ class VulnerabilityService(
             }
         )
             .map { it.toDto() }
-            .let { vulnerabilityProjects ->
-                vulnerability.toDtoWithDescription().copy(projects = vulnerabilityProjects)
-            }
     }
 
     /**
@@ -306,7 +303,6 @@ class VulnerabilityService(
      */
     @Transactional
     fun deleteProject(name: String, vulnerabilityName: String) = vulnerabilityProjectRepository.deleteByNameAndVulnerabilityName(name, vulnerabilityName)
-        .let { vulnerabilityRepository.findByName(name) }
 
     companion object {
         private const val VULNERABILITY_ORGANIZATION_RATING = 10
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -249,23 +249,28 @@ class VulnerabilityService(
     }
 
     /**
-     * @param vulnerabilityDtos dto list of vulnerability projects
+     * @param vulnerabilityProjectDtos dto list of vulnerability projects
+     * @return [VulnerabilityDto] with saved [vulnerabilityProjectDtos]
      */
     @Transactional
-    fun saveAllProjects(vulnerabilityDtos: List<VulnerabilityProjectDto>) {
-        vulnerabilityRepository.findByName(vulnerabilityDtos.first().vulnerabilityName)?.let { vulnerability ->
-            vulnerabilityProjectRepository.saveAll(
-                vulnerabilityDtos.map { dto ->
-                    VulnerabilityProject(
-                        name = dto.name,
-                        url = dto.url,
-                        versions = dto.versions,
-                        type = dto.type,
-                        vulnerability = vulnerability,
-                    )
-                }
-            )
-        }
+    fun saveAllProjects(
+        vulnerabilityProjectDtos: List<VulnerabilityProjectDto>
+    ): VulnerabilityDto? = vulnerabilityRepository.findByName(vulnerabilityProjectDtos.first().vulnerabilityName)?.let { vulnerability ->
+        vulnerabilityProjectRepository.saveAll(
+            vulnerabilityProjectDtos.map { dto ->
+                VulnerabilityProject(
+                    name = dto.name,
+                    url = dto.url,
+                    versions = dto.versions,
+                    type = dto.type,
+                    vulnerability = vulnerability,
+                )
+            }
+        )
+            .map { it.toDto() }
+            .let { vulnerabilityProjects ->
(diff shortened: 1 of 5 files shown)


nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -249,23 +249,28 @@ class VulnerabilityService(
...
+                    url = dto.url,
+                    versions = dto.versions,
+                    type = dto.type,
+                    vulnerability = vulnerability,
+                )
+            }
+        )
+            .map { it.toDto() }
+            .let { vulnerabilityProjects ->
+                vulnerability.toDtoWithDescription().copy(projects = vulnerabilityProjects)
```
Cheshiriks, line 272: I think it is a bad idea.
The SaveAll method returns the saved list back to you. I think it should be transferred to the controller and from there, from the first project, get the vulnerability and already transfer it in the response to the front

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: _@Transactional
    fun saveAllProjects(vulnerabilityDtos: List<VulnerabilityProjectDto>): List<VulnerabilityProject> {
        vulnerabilityRepository.findByName(vulnerabilityDtos.first().vulnerabilityName)?.let { vulnerability ->
            return  vulnerabilityProjectRepository.saveAll(...)
}_

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -297,10 +302,11 @@ class VulnerabilityService(
     /**
      * @param name name of project
      * @param vulnerabilityName name of vulnerability
-     * @return Unit
+     * @return updated [Vulnerability]
      */
     @Transactional
     fun deleteProject(name: String, vulnerabilityName: String) = vulnerabilityProjectRepository.deleteByNameAndVulnerabilityName(name, vulnerabilityName)
+        .let { vulnerabilityRepository.findByName(name) }
```
Cheshiriks, line 309: Here method vulnerabilityRepository.findByName(name)
will be executed before the delete method. It's better to get a vulnerability in the controller itself

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Why will it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Will fix it

sanyavertolet: Fixed as you suggested

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Removed this line.

It might be nice it you continue these in PR with `save` button removal

Cheshiriks: The data will only be committed after the transaction is closed

sanyavertolet: Oh, got it, thanks

Cheshiriks (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
nulls: > The data will only be committed in db after the transaction is closed

yeap, but we select inside the same transaction -- it should take into account changes from this transaction

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -249,23 +249,25 @@ class VulnerabilityService(
...
-                        type = dto.type,
-                        vulnerability = vulnerability,
-                    )
-                }
-            )
-        }
+    fun saveAllProjects(
+        vulnerabilityProjectDtos: List<VulnerabilityProjectDto>
+    ): List<VulnerabilityProjectDto>? = vulnerabilityRepository.findByName(vulnerabilityProjectDtos.first().vulnerabilityName)?.let { vulnerability ->
+        return@let vulnerabilityProjectRepository.saveAll(
```
nulls, line 259: do we need this `return@let`?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -249,23 +249,25 @@ class VulnerabilityService(
...
-                        versions = dto.versions,
-                        type = dto.type,
-                        vulnerability = vulnerability,
-                    )
-                }
-            )
-        }
+    fun saveAllProjects(
+        vulnerabilityProjectDtos: List<VulnerabilityProjectDto>
+    ): List<VulnerabilityProjectDto>? = vulnerabilityRepository.findByName(vulnerabilityProjectDtos.first().vulnerabilityName)?.let { vulnerability ->
```
nulls, line 258: you changed the return type, but it doesn't lead to change anywhere. Do we need this return value?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Nope, will remove it

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: It will be used it @Cheshiriks's PR

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Removed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: > > The data will only be committed in db after the transaction is closed
> 
> yeap, but we select inside the same transaction -- it should take into account changes from this transaction

Yes, but in order for these changes to be visible in the transaction itself, you must call the forced  SaveAndFlush or DeleteAndFlush methods

nulls: looks like you re right in general, but it should be visible by default is due to `auto flush` according to https://stackoverflow.com/a/56816364.

so yes, better to use `saveAndFlush` in such cases

PR #2282, opened 2023-07-05
Vulnerability dates should be able to be deleted

sanyavertolet:
### What's done:
 * Implemented vulnerability date deletion on backend
 * Supported vulnerability date deletion on frontend
 * Fixed an issue when timeline labels acted the same as timeline nodes (now labels are not clickable and `onHover` is not triggered on them)
 * Added unique constraint on dates in order to prevent saving duplicates
 * Added `409 CONFLICT` as possible API response code for `/save-date` in order to handle it on frontend

### How it looks:

https://github.com/saveourtool/save-cloud/assets/35039155/393d64f4-a051-4057-b416-350ec84aa2a6

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
@@ -11,10 +11,7 @@ import com.saveourtool.save.entities.vulnerability.VulnerabilityDto
 import com.saveourtool.save.entities.vulnerability.VulnerabilityProjectDto
 import com.saveourtool.save.filters.VulnerabilityFilter
 import com.saveourtool.save.info.UserInfo
-import com.saveourtool.save.utils.StringResponse
-import com.saveourtool.save.utils.blockingToFlux
-import com.saveourtool.save.utils.blockingToMono
-import com.saveourtool.save.utils.switchIfEmptyToNotFound
+import com.saveourtool.save.utils.*
 import com.saveourtool.save.v1
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.responses.ApiResponse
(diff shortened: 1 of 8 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
```diff
@@ -247,15 +245,38 @@ class VulnerabilityController(
...
+        method = "POST",
+        summary = "Delete date.",
+        description = "Delete date.",
+    )
+    @PreAuthorize("permitAll()")
+    @ApiResponse(responseCode = "200", description = "Successfully deleted date in vulnerability")
+    @ApiResponse(responseCode = "404", description = "Requested date is not found")
+    @Suppress("UnusedParameter")
+    fun deleteDate(
+        @RequestBody vulnerabilityDateDto: VulnerabilityDateDto,
```
nulls, line 274: I guess do you use `post` because of `request body`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/VulnerabilityDateModal.kt
```diff
@@ -25,17 +25,20 @@ import kotlinx.serialization.json.Json
  */
 val vulnerabilityDateModal: FC<VulnerabilityDateModalProps> = FC { props ->
     val (vulnerabilityDate, setVulnerabilityDate) = useState(VulnerabilityDateDto.empty)
-
+    val (conflictMessage, setConflictMessage) = useState<String?>(null)
```
nulls, line 28: as I remember, you do prefer `useState<T?>(null)` instead of `useState<T>()`, don't you?
In case it's not -- `useState<T>() == useState<T?>(null)`

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
sanyavertolet: Yes

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/VulnerabilityDateModal.kt
sanyavertolet: I know it is basically the same (except `useState<T>()` __might__ be `undefined`), but yes, I do prefer the first variant

PR #2292, opened 2023-07-05
Added participants in vulnerability dto

Cheshiriks:
Added participants in vulnerability dto

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -145,9 +145,16 @@ class VulnerabilityService(
      * @return vulnerability dto by name with description
      */
     fun getVulnerabilityWithDescriptionByName(name: String): VulnerabilityDto? {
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -144,10 +144,18 @@ class VulnerabilityService(
      * @param name name of vulnerability
      * @return vulnerability dto by name with description
      */
+    @Transactional
     fun getVulnerabilityWithDescriptionByName(name: String): VulnerabilityDto? {
-        val vulnerability = vulnerabilityRepository.findByName(name)?.toDtoWithDescription()
+        val vulnerability = vulnerabilityRepository.findByName(name).orNotFound {
+            "Not found vulnerability by name = $name"
```
sanyavertolet, line 150:
```suggestion
            "Not found vulnerability by name $name"
```

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: fixed

PR #2293, opened 2023-07-05
Make `SaveCloudClientEx` compatible with the current state of the back-end

0x6675636b796f75676974687562:
### What's done:

 - Reflected the changes made to the back-end over the last 6 months.

### DIFF
--- a/save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
+++ b/save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
@@ -6,7 +6,7 @@ import com.saveourtool.save.agent.TestExecutionExtDto
 import com.saveourtool.save.api.errors.SaveCloudError
 import com.saveourtool.save.api.impl.DefaultSaveCloudClient
 import com.saveourtool.save.entities.FileDto
-import com.saveourtool.save.entities.Organization
+import com.saveourtool.save.entities.OrganizationDto
 import com.saveourtool.save.entities.ProjectDto
 import com.saveourtool.save.entities.ProjectStatus.CREATED
 import com.saveourtool.save.entities.contest.ContestDto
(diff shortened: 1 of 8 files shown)


save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
```diff
@@ -203,7 +203,7 @@ interface SaveCloudClientEx {
      * @throws IllegalArgumentException if [file] is not a regular file.
      * @see SaveCloudClientEx.uploadFile
      */
-    suspend fun Organization.uploadFile(
+    suspend fun OrganizationDto.uploadFile(
```
akuleshov7, line 206: what is the change? Only this move to DTO?

akuleshov7 (approved):

save-api/src/main/kotlin/com/saveourtool/save/api/SaveCloudClientEx.kt
0x6675636b796f75676974687562: Mostly, yes, because now `/organizations/get/list` returns the list of DTOs.

Additionally, file size specified via `Content-Length-Custom` is now required when uploading a file.

save-api/src/main/kotlin/com/saveourtool/save/api/impl/DefaultSaveCloudClient.kt
```diff
@@ -162,7 +165,9 @@ internal class DefaultSaveCloudClient(
                         headers
                     )
                 })
-            ).getOrElse { error ->
+            ) {
+                this[CONTENT_LENGTH_CUSTOM] = file.fileSize().toString()
```
0x6675636b796f75676974687562, line 169: @akuleshov7 Here's the 2nd change.

akuleshov7: missed that, sounds like we made not so many in 6 months 😆

0x6675636b796f75676974687562: There's also the change in the way TS version is now reported if the version is a Git branch name.

Used to be just the branch name, like `master`. Now it also includes the short commit hash, like `master (deadbeef)`.

PR #2299, opened 2023-07-06
Polishing index view animation

akuleshov7:
### What's done:
- changed default redirection
- adding perfect animation to Index view

https://github.com/saveourtool/save-cloud/assets/58667063/9c1f502a-fe0c-48aa-80ea-da6001b00d72


sanyavertolet (approved):
Looks ultra good, but the only thing I would suggest it to change the frequency of neon titles blinking

akuleshov7:
> Looks ultra good, but the only thing I would suggest it to change the frequency of neon titles blinking

I made it less frequent and also removed animation on a border, but yes, can do it less frequent

Cheshiriks (approved):

akuleshov7:
> Looks ultra good, but the only thing I would suggest it to change the frequency of neon titles blinking

done (reduced 2 times)

PR #2303, opened 2023-07-07
Added vulnerability permission

Cheshiriks:
Added vulnerability permission

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
@@ -1,5 +1,6 @@
 package com.saveourtool.save.backend.controllers.vulnerability
 
+import com.saveourtool.save.backend.security.VulnerabilityPermissionEvaluator
 import com.saveourtool.save.backend.service.vulnerability.VulnerabilityService
 import com.saveourtool.save.backend.utils.hasRole
 import com.saveourtool.save.configs.ApiSwaggerSupport
@@ -12,6 +13,7 @@ import com.saveourtool.save.entities.vulnerability.VulnerabilityProjectDto
 import com.saveourtool.save.entities.vulnerability.VulnerabilityStatus
 import com.saveourtool.save.filters.VulnerabilityFilter
 import com.saveourtool.save.info.UserInfo
+import com.saveourtool.save.permission.Permission
 import com.saveourtool.save.utils.*
 import com.saveourtool.save.v1
 import io.swagger.v3.oas.annotations.Operation
@@ -41,6 +43,7 @@ typealias VulnerabilityDtoList = List<VulnerabilityDto>
 @RequestMapping(path = ["/api/$v1/vulnerabilities"])
 class VulnerabilityController(
     private val vulnerabilityService: VulnerabilityService,
+    private val vulnerabilityPermissionEvaluator: VulnerabilityPermissionEvaluator,
 ) {
     @PostMapping("/by-filters")
     @Operation(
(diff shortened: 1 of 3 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/security/VulnerabilityPermissionEvaluator.kt
```diff
@@ -0,0 +1,56 @@
...
+) {
+    /**
+     * Check permission for user to read, write and delete vulnerabilities by its [vulnerabilityName]
+     *
+     * @param authentication
+     * @param vulnerabilityName
+     * @param permission
+     * @return true if user with [authentication] has [permission] for [vulnerabilityName]
+     */
+    fun hasPermission(
```
akuleshov7, line 28: change the name to something more normal: `hasPermissionToEdit`

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/VulnerabilityPermissionEvaluator.kt
```diff
@@ -0,0 +1,56 @@
...
+        authentication: Authentication?,
+        vulnerabilityName: String,
+        permission: Permission,
+    ): Boolean {
+        authentication ?: return false
+
+        if (authentication.hasRole(Role.SUPER_ADMIN)) {
```
akuleshov7, line 35: create a common `when` statement 

when {
authentication.hasRole(Role.SUPER_ADMIN) -> 
permission == Permission.READ
else ->
}

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
```diff
@@ -253,9 +256,12 @@ class VulnerabilityController(
     fun saveDate(
         @RequestBody vulnerabilityDateDto: VulnerabilityDateDto,
         authentication: Authentication,
-    ): Mono<StringResponse> = blockingToMono {
-        vulnerabilityService.saveDate(vulnerabilityDateDto, authentication)
-    }
+    ): Mono<StringResponse> = vulnerabilityDateDto.vulnerabilityName.toMono()
+        .filter {
+            vulnerabilityPermissionEvaluator.hasPermission(authentication, it, Permission.DELETE)
```
akuleshov7, line 261: I thought you will have permissions like user/admin/etc., but you have it as attributes: Delete, Edit, etc.
Don't know which is the best one

akuleshov7 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/VulnerabilityController.kt
sanyavertolet: As we already have several `PermissionEvaluator`s, I would support existing attribute system

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/VulnerabilityPermissionEvaluator.kt
sanyavertolet: I would leave as is because of `permission` param

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -364,6 +370,13 @@ class VulnerabilityService(
         lnkVulnerabilityUserRepository.deleteByUserNameAndVulnerabilityId(userName, vulnerability.requiredId())
     }
 
+    /**
+     * @param vulnerabilityId id of vulnerability
+     * @return list of LnkVulnerabilityUser link to vulnerability
+     */
+    @Transactional
```
nulls, line 377: maybe add reaonly=true?

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: Yes, I removed that annotation. The class itself is already marked ReadOnly

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/VulnerabilityPermissionEvaluator.kt
Cheshiriks: Fixed

PR #2304, opened 2023-07-07
Authorization with other source providers

akuleshov7:
### What's done:
- Fix for registration of new users
- Polishing formating of auth providers
- Dev configuration for Gitee OAUTH (avatars and production auth keys will be updated later in next PRs)

https://github.com/saveourtool/save-cloud/assets/58667063/a899650b-a00e-4ef2-a5fa-498815906932

### DIFF
--- a/api-gateway/src/main/resources/application-dev.yml
+++ b/api-gateway/src/main/resources/application-dev.yml
@@ -16,6 +16,10 @@ spring:
       client:
         provider:
           # example: https://github.com/wearearima/spring-boot-dex/blob/master/src/main/resources/application.properties
+          gitee:
+            authorization-uri: https://gitee.com/oauth/authorize
+            token-uri: https://gitee.com/oauth/token
+            user-info-uri: https://gitee.com/api/v5/user
           dex:
             authorization-uri: http://localhost:5556/dex/auth
             token-uri: http://localhost:5556/dex/token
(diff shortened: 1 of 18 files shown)


api-gateway/src/main/resources/application-dev.yml
```diff
@@ -31,6 +35,15 @@ spring:
             # for more details.
             user-name-attribute: login
         registration:
+          gitee:
+            client-id: 6b7fc07ecdf7a12d9aa8e2aaf034743baa6f77e036dc22fe6455e02fcf51a851
```
akuleshov7, line 39: fyi: this is "test" config for gitee

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/App.kt
```diff
@@ -112,17 +112,10 @@ class App : ComponentWithScope<PropsWithChildren, AppState>() {
             requestModalHandler {
                 userInfo = state.userInfo
 
-                withRouter<Props> { location, _ ->
```
akuleshov7, line 115: here was a bug with redirection (after some of updates such navigation does not work). And actually that was an incorrect  one from the very beginning

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -150,6 +158,9 @@ class RegistrationView : AbstractView<RegistrationProps, RegistrationViewState>(
                         className = ClassName("col-sm-4")
                         div {
                             className = ClassName("container card o-hidden border-0 shadow-lg my-2 card-body p-0")
+                            style = jso {
+                                zIndex = "900".unsafeCast<ZIndex>()
```
sanyavertolet, line 162: Need to check that modals have a bigger `z-index`

sanyavertolet (approved):

nulls (approved):

nulls (approved):

PR #2309, opened 2023-07-10
Add users to vulnerability

sanyavertolet:
### What's done:
 * Implemented `addUserComponent` based on `inputWithDebounce`
 * Supported option fetch from backend
 * Fixed `createVulnerabilityView` paddings

### How it looks:

https://github.com/saveourtool/save-cloud/assets/35039155/9162b577-eb38-4d96-96ca-613df85cc732

#### upd:
screenshot

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/CreateVulnerabilityView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/CreateVulnerabilityView.kt
@@ -29,6 +29,8 @@ import react.dom.html.ReactHTML.textarea
 import react.router.useNavigate
 import react.useState
 import web.cssom.ClassName
+import web.cssom.Height
+import web.cssom.Width
 import web.html.InputType
 
 import kotlinx.browser.window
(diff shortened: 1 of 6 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/CreateVulnerabilityView.kt
```diff
@@ -314,58 +290,98 @@ val createVulnerabilityView: VFC = VFC {
...
-                                        buttonBuilder(faPlus, isOutline = true, classes = "btn-sm rounded-circle") {
-                                            projectWindowOpenness.openWindow()
+                                        vulnerability.participants.forEach { user ->
+                                            div {
+                                                className = ClassName("mt-2 d-flex align-items-center")
+                                                div {
+                                                    className = ClassName("col-2 d-flex justify-content-center align-items-center")
+                                                    img {
+                                                        className = ClassName("img-fluid")
+                                                        src = user.avatar ?: "img/undraw_profile.svg"
```
Cheshiriks, line 313: I would reduce the size of the avatars by 1.5 or 2 times

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/fossgraph/CreateVulnerabilityView.kt
sanyavertolet: makes sense

sanyavertolet: done

PR #2329, opened 2023-07-13
Proper support for users that are coming from Google

akuleshov7:
### What's done:
- prod configuration for Google
- as legacy Google does not have 'login', so we use email as a name and trim @gmail from it


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -221,13 +221,16 @@ class RegistrationView : AbstractView<RegistrationProps, RegistrationViewState>(
         "PARAMETER_NAME_IN_OUTER_LAMBDA",
     )
     private fun ChildrenBuilder.renderInputForm() {
+        // google does not provide us login, only e-mail, so need to trim everything after "@gmail"
+        val rawInput = state.fieldsMap[InputTypes.USER_NAME] ?: ""
+        val atIndex = rawInput.indexOf('@')
```
kgevorkyan, line 226: do we forbid manual creating the name with `@`?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
```diff
@@ -221,13 +221,16 @@ class RegistrationView : AbstractView<RegistrationProps, RegistrationViewState>(
         "PARAMETER_NAME_IN_OUTER_LAMBDA",
     )
     private fun ChildrenBuilder.renderInputForm() {
+        // google does not provide us login, only e-mail, so need to trim everything after "@gmail"
+        val rawInput = state.fieldsMap[InputTypes.USER_NAME] ?: ""
```
kgevorkyan, line 225: "" means guest? do we have some data class with such field (`GUEST`)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
akuleshov7: yes, moreover - you even cannot input "@" into the input field

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
sanyavertolet: why not `.orEmpty()`?

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/RegistrationView.kt
akuleshov7: very good suggestion, please create an issue for diktat @sanyavertolet 
We have a thousand of places where it is in save, will change everywhere

akuleshov7: > very good suggestion, please create an issue for diktat @sanyavertolet
> We have a thousand of places where it is in save, will change everywhere

@sanyavertolet I will create

akuleshov7: https://github.com/saveourtool/diktat/issues/1710

PR #2339, opened 2023-07-16
Changing filters for vulnerability view

akuleshov7:
STEP 1: We will need to add A LOT of filters on this page, but now we should at least refactor and split filters for it

### What's done:
- updating filters
- polishing vuln. collection view

image

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
@@ -1,12 +1,10 @@
+@file:Suppress("FILE_NAME_MATCH_CLASS")
+
 package com.saveourtool.save.frontend.components.basic.table.filters
 
 import com.saveourtool.save.frontend.externals.fontawesome.*
-import com.saveourtool.save.frontend.utils.buttonBuilder
-import com.saveourtool.save.frontend.utils.withNavigate
-import com.saveourtool.save.validation.FrontendRoutes
 import react.FC
 import react.Props
-import react.dom.html.ReactHTML
 import react.dom.html.ReactHTML.button
 import react.dom.html.ReactHTML.div
 import react.dom.html.ReactHTML.input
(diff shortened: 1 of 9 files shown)


akuleshov7:
We need to have a possibility to filter criticality: https://mdbootstrap.com/docs/b4/jquery/plugins/multi-range/

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+            button {
+                type = ButtonType.button
+                className = ClassName("btn btn-secondary mr-3")
+                fontAwesomeIcon(icon = faSearch)
+                onClick = {
+                    props.onChangeFilters(filtersName)
+                }
+            }
```
sanyavertolet, line 60:
```suggestion
            buttonBuilder(faSearch, style = "secondary", classes = "mr-3") {
                props.onChangeFilters(filtersName)
            }
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
```diff
@@ -0,0 +1,87 @@
...
+            button {
```
sanyavertolet, line 61: same here

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
akuleshov7: If you can see - I have updated new filters in new code to button builder, this is old one...
I will upgrade ALL buttons to buttonBuilder a little bit later in one PR

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
akuleshov7: yep, same in TestExecutionFilters

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/NameFiltersRow.kt
sanyavertolet: oh ok got it

0x6675636b796f75676974687562 (approved):

Cheshiriks (approved):

PR #2342, opened 2023-07-17
vulnWelcomeView

sanyavertolet:
### What's done:
 * Added `vulnWelcomeView`
 * Added vulnerability info cards

### How it looks:
screenshot


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/AboutUsView.kt
```diff
@@ -45,7 +45,7 @@ open class AboutUsView : AbstractView<AboutUsViewProps, AboutUsViewState>() {
         Developer("Vlad", "Frolov", "Cheshiriks", "Fullstack"),
         Developer("Peter", "Trifanov", "petertrr", "Arch"),
         Developer("Andrey", "Shcheglov", "0x6675636b796f75676974687562", "Backend"),
-        Developer("Sasha", "Frolov", "sanyavertolet", "Frontend"),
+        Developer("Sasha", "Frolov", "sanyavertolet", "Fullstack"),
```
nulls, line 48: such change during changing in frontend looks funny)

sanyavertolet: it is, but [still](https://saveourtool.com/#/demo)

PR #2362, opened 2023-07-21
Removed link to deleted user

Cheshiriks:
Снимок


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/AvatarRenderers.kt
```diff
@@ -47,7 +49,40 @@ fun ChildrenBuilder.renderAvatar(
     classes: String = "",
     link: String? = null,
     styleBuilder: CSSProperties.() -> Unit = {},
-) = renderAvatar(userInfo.avatar ?: AVATAR_PROFILE_PLACEHOLDER, classes, link, styleBuilder)
+) {
+    val newLink = if (userInfo.status != UserStatus.DELETED) {
```
sanyavertolet, line 53: `.takeIf { }`?

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/AvatarRenderers.kt
Cheshiriks: Fixed

PR #2367, opened 2023-07-24
Hotfixes for avatar rendering

akuleshov7:
### What's done:
- added hotfix for avatars sizes


sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/AvatarRenderers.kt
```diff
@@ -85,7 +92,7 @@ private fun ChildrenBuilder.renderAvatar(
     avatarLink: String,
     classes: String,
     link: String?,
-    styleBuilder: CSSProperties.() -> Unit,
+    styleBuilder: CSSProperties.() -> Unit
```
nulls, line 95: why `,` at the end was removed?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/AvatarRenderers.kt
akuleshov7: hm, strange, diktat

PR #2373, opened 2023-07-26
Unifying tags

sanyavertolet:
### What's done:
 * Added tag table
 * Added lnk_vulnerability_tag
 * Added general Tag entity - this entity should be linked with another taggable entity
 * Added TagRepository and TagService
 * Supported Tag entity in a field of Vulnerability entity - created LnkVulnerabilityTag
 * Dropped tags column in vulnerability table
 
 ### TODO:
  * Implement filtering by tag and double-test it


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -229,11 +280,15 @@ class VulnerabilityService(
             status = VulnerabilityStatus.CREATED,
             userId = user.requiredId(),
             organization = organizationNew,
-            tags = vulnerabilityDto.tags.joinToString(DATABASE_DELIMITER),
         )
         val vulnerabilityNew = vulnerabilityRepository.saveAndFlush(vulnerability)
         val newName = "SOTV-${LocalDateTime.now().year}-${vulnerabilityNew.id}"
 
+        val existedTags = tagRepository.findByNameIn(vulnerabilityDto.tags)
```
kgevorkyan, line 287: do we need this check? `spring-data-jpa` will just update the existing record, if it is the same
and in our case, since we didnt change anything, the record wont change, isnt it?

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -271,6 +326,9 @@ class VulnerabilityService(
                 )
             }
         )
+
+        val vulnTagLinks = existedTags.plus(newTags).map { LnkVulnerabilityTag(vulnerability, it) }
```
kgevorkyan, line 330: `existedTags.plus(newTags)` == `vulnerabilityDto.tags`, I guess?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: The point is that primary key is `id`, not name. Thus, we do not save duplicates

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: `existedTags` and `newTags` are `Tag` entities, while `VulnerabilityDto.tags` are just `String`s

db/v-2/tables/tags.xml
```diff
@@ -0,0 +1,19 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="tag-1" author="sanyavertolet" context="dev or prod">
+        <createTable tableName="tag">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="name" type="varchar(16)">
```
nulls, line 13: I guess we can add a unique constraint for name too

sanyavertolet: Sure, will do it

PR #2375, opened 2023-07-26
Fixing small FE problems

akuleshov7:
### What's done:
- alignment of avatars
- fixing 404 page


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/FallbackView.kt
```diff
@@ -45,31 +49,63 @@ external interface FallbackViewProps : Props {
...
+                    className = ClassName("col-4")
+                    p {
+                        className = ClassName("lead text-gray-800")
+                        +"Report a problem:"
+                    }
+
+                    a {
+                        className = ClassName("mt-3")
+                        href = "https://github.com/saveourtool/save-cloud"
+                        ReactHTML.img {
```
nulls, line 101:
```suggestion
                        img {
```

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/ProjectView.kt
```diff
@@ -219,7 +219,6 @@ class ProjectView : AbstractView<ProjectViewProps, ProjectViewState>(Style.SAVE_
             ProjectMenuBar.SETTINGS -> renderSettings()
             ProjectMenuBar.INFO -> renderInfo()
             ProjectMenuBar.DEMO -> renderDemo()
-            ProjectMenuBar.SECURITY -> renderSecurity()
```
nulls, line 222: we don't it anymore, do we?

akuleshov7: I still plan to use it one day

PR #2383, opened 2023-07-28
BrowserRouter

sanyavertolet:
It _might_ work

### What's done:
 * Enriched webpack configuration in order to support `BrowserRouter`
 * Replaced all the relative paths to resources to absolute ones
 * Removed all the logic connected with smart urls (each tab used to have its own url)
 * Replaced all the `Location.hash` occurrences with `Location.pathname`
 * Tested locally with and without `gateway` on


api-gateway/src/main/resources/error.html
```diff
@@ -10,7 +10,7 @@
 
     <title>SAVE Cloud Service</title>
 
-    <link rel="stylesheet" type="text/css" href="main.css">
+    <link rel="stylesheet" type="text/css" href="/main.css">
```
akuleshov7, line 13: Ah, that was the reason, lol

akuleshov7 (approved):
Let’s test on prod, but fix tests…

api-gateway/src/main/resources/error.html
sanyavertolet: The main reason for a problem was some plugin for HTML processing. It used to add a script tat to the head of html. The path was relative there.

sanyavertolet:
> but fix tests…

Seems that `AppTest` is not `App` test anymore. It is `IndexTest`. `BrowserRouter` seem to work wrong now, as within test it gives `Not Found` page while running the frontend locally works as expected: `indexView` is rendered.

akuleshov7:
> > but fix tests…
> 
> Seems that `AppTest` is not `App` test anymore. It is `IndexTest`. `BrowserRouter` seem to work wrong now, as within test it gives `Not Found` page while running the frontend locally works as expected: `indexView` is rendered.

Fuck tests then, delete them

Issue #2336, opened 2023-07-14
Separator for user and auth provider should be revised


kgevorkyan:
For now we are using `@` for separate the user name and auth provider **at least** in `WebSecurityConfig` and in `UserUtils.kt`

But this symbol **is not the variable** and just the hardcoded symbol, so its hard to maintain this functionality

First of all, it should be extracted into the variable in `save-cloud-common`

And after this, we probably will need to change `@` to something else, since `@` is too dangerous to be a spec symbol

This work should be done under separate issue and **very carefully,** since it influence on core functionality - `API` application, `auth`. `integration tests`, `db`

akuleshov7:
**CRITICAL thing we have just found**:
1) User VLAD comes from Github 
2) Initially spring register him as VLAD@GITHUB
3) Then he is redirected to a registration view, where he CAN change VLAD name to ANDREY
4) ANDREY comes from GITHUB 
5) Now ANDREY is not even able to press OAuth registration button

akuleshov7:
That's why we need to GENERATE a our own SAVE ID for user when he first time register!

kgevorkyan:
seems, that `@` should not influent on problems with `email` at all, since its only used in `basic auth`, but should be extracted anyway

Cheshiriks:
```
        .httpBasic { httpBasicSpec ->
            // Authenticate by comparing received basic credentials with existing one from DB
            httpBasicSpec.authenticationManager(
                UserDetailsRepositoryReactiveAuthenticationManager { username ->
                    // Looking for user in DB by received source and name
                    require(username.contains("@")) {
                        "Provided user information should keep the following form: oauth2Source@username"
                    }
```

Also be careful with Basic oauth

Cheshiriks:
> **That's why we need to GENERATE a our own SAVE ID for user when he first time register!**

```
    fun findByName(name: String): User? {
        val record = namedParameterJdbcTemplate.queryForList(
            "SELECT * FROM save_cloud.user WHERE name = :name",
            mapOf("name" to name)
        ).singleOrNull() ?:
        namedParameterJdbcTemplate.queryForList(
            "SELECT * FROM save_cloud.user WHERE id = (select user_id from save_cloud.original_login where name = :name)",
            mapOf("name" to name)
        ).singleOrNull()
            .orNotFound {
                "There is no user with name $name"
            }
        return record.toUserEntity()
    }
```

nulls:
_For history_
[...]

PR #2386, opened 2023-07-31
Pre authenticated user

nulls:
### What's done:
- supported pre-authenticated user

### How it does work:
Api Gateway fetches users from Backend to resolve save's user for OAuth2 and Basic.
Api Gateway saves id, name and role from resolved save's user in headers.
Backend (and Sandbox) using authentication-service module resolves `PreAuthenticatedAuthenticationToken` from values in headers without checking database.

It's part of #2336


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/configs/ApiSwaggerSupport.kt
```diff
@@ -27,12 +29,24 @@ import io.swagger.v3.oas.annotations.security.SecurityScheme
 annotation class ApiSwaggerSupport
 
 /**
- * Indicates that an operation requires `X-Authorization-Source` header to be set
+ * Indicates that an operation requires `X-Authorization-*` headers to be set
  */
 @Parameter(
```
nulls, line 34: these headers look are useless for me.
I think it should be removed since user cannot provide it (as `X-Authorization-Source` after migration to k8s), they are populated by api-gateway.

This change leads to big changes in source code and open api. Will do it in separate PR

save-backend/src/test/kotlin/com/saveourtool/save/backend/utils/SecurityTestingUtils.kt
```diff
@@ -4,14 +4,19 @@
...
 import org.springframework.security.core.context.SecurityContextHolder
 
 /**
  * @param action mutation for a mocked user in test security context
  */
-internal fun mutateMockedUser(action: UsernamePasswordAuthenticationToken.() -> Unit) {
+internal fun mutateMockedUser(id: Long, name: String = "N/A") {
     SecurityContextHolder.getContext().apply {
-        authentication = (authentication as UsernamePasswordAuthenticationToken).apply(action)
+        authentication = AuthenticationUserDetails(
```
nulls, line 16: currently we do:
```kotlin
@WithMockUser(name = "something", roles = ["SINGLE_ROLE"])
fun test() {
    mutateMockedUser(id = 99)
    TODO()
}
```

It can be refactored to:
```kotlin
fun test() {
    mutateMockedUser(id = 99, name = "something", roles = "SINGLE_ROLE")
    TODO()
}
```

authentication-service/src/main/kotlin/com/saveourtool/save/authservice/utils/AuthenticationUserDetails.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    companion object {
+        @Suppress("GENERIC_VARIABLE_WRONG_DECLARATION")
+        private val log = getLogger<AuthenticationUserDetails>()
+        private const val NO_CREDENTIALS = "N/A"
+
+        /**
+         * @return [AuthenticationUserDetails] created from values in headers
+         */
+        fun ServerWebExchange.toAuthenticationUserDetails(): AuthenticationUserDetails? {
```
nulls, line 59: can be covered by test

nulls: done

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/AuthorizationHeadersGatewayFilterFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+import reactor.kotlin.core.publisher.toMono
+import java.security.Principal
+
+/**
+ * Filter, that mutate existing exchange,
+ * inserts user's info into Authorization headers instead of existing value, not paying attention to the credentials,
+ * since at this moment they are already checked by gateway.
+ */
+@Component
+class AuthorizationHeadersGatewayFilterFactory(
```
nulls, line 26: it's to hard to cover by test, will be done in separate task in scope of #2336

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
```diff
@@ -22,42 +21,35 @@ import reactor.core.publisher.Mono
...
         .retrieve()
         .onStatus({ it.is4xxClientError }) {
             Mono.error(ResponseStatusException(it.statusCode()))
         }
-        .toEntity<String>()
-        .map {
-            springUserDetailsReader.readValue(it.body)
+        .toEntity<AuthenticationUserDetails>()
+        .flatMap { responseEntity ->
+            responseEntity.body.toMono().orNotFound { "Body is empty" }
```
kgevorkyan, line 52: maybe 
```suggestion
            responseEntity.body.toMono().orNotFound { "Authentication body is empty" }
```

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/AuthorizationHeadersGatewayFilterFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+                    }
+                    .build()
+            }
+            .defaultIfEmpty(exchange)
+            .flatMap { chain.filter(it) }
+    }
+
+    private fun resolveSaveUser(principal: Principal): Mono<AuthenticationUserDetails> = when (principal) {
+        is OAuth2AuthenticationToken -> backendService.findByOriginalLogin(principal.authorizedClientRegistrationId, principal.name)
+        is UsernamePasswordAuthenticationToken -> (principal.principal as? AuthenticationUserDetails)
```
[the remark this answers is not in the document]
nulls: we expect here only `uthenticationUserDetails` and even we throw exception here. I think it can be safely changed to `as`.

about what we discussed, it was the statement like `when (value) { is A -> value as A }`. when type is matched in section `is` and then in body

authentication-service/src/main/kotlin/com/saveourtool/save/authservice/config/WebSecurityConfig.kt
```diff
@@ -53,9 +67,9 @@ class WebSecurityConfig(
             // FixMe: Properly support CSRF protection https://github.com/saveourtool/save-cloud/issues/34
             csrf().disable()
         }
-        .httpBasic {
-            it.authenticationManager(authenticationManager)
-        }
+        .httpBasic()
+        .disable()
```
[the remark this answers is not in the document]
nulls: we don't need basic auth (or even any spring's auth) in backend and sandbox -- we use pre-authenticated user in internal communication.
if request goes through api-gateway, it's authenticated.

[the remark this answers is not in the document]
nulls: actually, httpBasic is not enabled by default. Wil remove `.httpBasic().disable()`

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/AuthorizationHeadersGatewayFilterFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+    private val backendService: BackendService,
+) : AbstractGatewayFilterFactory<Any>() {
+    override fun apply(config: Any?): GatewayFilter = GatewayFilter { exchange: ServerWebExchange, chain: GatewayFilterChain ->
+        exchange.getPrincipal<Principal>()
+            .flatMap { resolveSaveUser(it) }
+            .map { user ->
+                exchange.mutate()
+                    .request { builder ->
+                        builder.headers { headers: HttpHeaders ->
+                            headers.remove(HttpHeaders.AUTHORIZATION)
```
akuleshov7, line 36: btw - why do you remove authorization headers? 
May be not create it initially for particular endpoints?

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
```diff
@@ -68,8 +61,8 @@ class UsersController(
     @GetMapping("/find-by-name/{userName}")
     fun findByName(
         @PathVariable userName: String,
-    ): Mono<StringResponse> = userService.findByName(userName).map {
-        ResponseEntity.ok().body(springUserDetailsWriter.writeValueAsString(it))
+    ): Mono<AuthenticationUserDetailsResponse> = userService.findByName(userName).map {
+        ResponseEntity.ok().body(AuthenticationUserDetails(it))
```
akuleshov7, line 65: Ah, you moved it from a simple string to some data class, that's why you need new UserDetails for authorization? (this is how I get it)

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/configs/ApiSwaggerSupport.kt
akuleshov7: yes, definitely, we are isolated with k8s

akuleshov7 (approved):

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/utils/AuthorizationHeadersGatewayFilterFactory.kt
nulls: we don't use authorization header in backend. I just cleanup it to be sure that we don't pass through this value from external request

authentication-service/src/main/kotlin/com/saveourtool/save/authservice/utils/AuthenticationUserDetails.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.security.core.authority.AuthorityUtils
+import org.springframework.security.core.userdetails.UserDetails
+import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken
+
+/**
+ * @property id [com.saveourtool.save.entities.User.id]
+ * @property name [com.saveourtool.save.entities.User.name]
+ * @property role [com.saveourtool.save.entities.User.role]
+ */
+data class AuthenticationUserDetails(
```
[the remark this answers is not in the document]
nulls: we have `AuthenticationDetails` which contains `id` and it's set to `Authentication.details`.
But `Authentication.principal` can be any object and Spring supports `UserDetails` to resolve `Authentication.name`.
I decided to store all required info pre-authenticated user in principal and `UserDetails` interface allows to pass username to Spring engine

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
nulls: yes, but here is another change:
I don't why but we didn't use the Spring engine to parse Object provided by backend to api-gateway. We created a custom `objectMapper` to decode it from String.
I guess the reason is in `getters`. Even in current implementation I have to add `@JsonIgnore` in `AuthenticationUserDetails`. Previously we used `org.springframework.security.core.userdetails.User` and additionally registered `org.springframework.security.jackson2.SecurityJackson2Modules` (before it we have a custom extension for `SpringUser` and a custom serializer for it)

PR #2397, opened 2023-08-02
Show only active organizations in Profile view

Cheshiriks:
Show only active organizations in Profile view


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -153,10 +153,10 @@ internal class OrganizationController(
         description = "Get list of all organizations where current user is a participant.",
     )
     @ApiResponse(responseCode = "200", description = "Successfully fetched list of organizations.")
-    fun getOrganizationsByUserName(
```
kgevorkyan, line 156: may be add new endpoint for showing only created? or add argument `CREATED` to this common method?

Cheshiriks: This method is not used anywhere else and was created only to display organizations and the user profile. That's why I didn't add a new method

kgevorkyan (approved):

nulls (approved):

sanyavertolet (approved):

akuleshov7 (approved):

PR #2398, opened 2023-08-02
Show avatars for non-login users

Cheshiriks:
Show avatars for non-login users


authentication-service/src/main/kotlin/com/saveourtool/save/authservice/config/WebSecurityConfig.kt
```diff
@@ -108,6 +108,7 @@ class WebSecurityConfig(
             "/api/$v1/vulnerabilities/by-name-with-description",
             "/api/$v1/comments/get-all",
             "/api/$v1/users/all",
+            "/api/$v1/avatar/*/*",
```
nulls, line 111: maybe `"/api/$v1/avatar/**"`?

sanyavertolet (approved):

authentication-service/src/main/kotlin/com/saveourtool/save/authservice/config/WebSecurityConfig.kt
Cheshiriks: Fixed

akuleshov7 (approved):

PR #2399, opened 2023-08-02
Hotfix broken redirects

sanyavertolet:
### What's done:
 * Reused `react-router` navigation instead of `window.location` manipulations

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -153,10 +153,10 @@ internal class OrganizationController(
         description = "Get list of all organizations where current user is a participant.",
     )
     @ApiResponse(responseCode = "200", description = "Successfully fetched list of organizations.")
-    fun getOrganizationsByUserName(
+    fun getOrganizationsByUserNameAndCreatedStatus(
         @RequestParam userName: String,
     ): Flux<OrganizationDto> = blockingToFlux {
-        lnkUserOrganizationService.findAllByUserName(userName).map { it.organization.toDto() }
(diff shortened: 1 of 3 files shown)


kgevorkyan (approved):

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/FallbackView.kt
```diff
@@ -74,7 +75,7 @@ class FallbackView : AbstractView<FallbackViewProps, State>(Style.SAVE_LIGHT) {
             }
 
             if (props.withRouterLink == true) {
-                Link {
+                Navigate {
                     to = "/"
```
kgevorkyan, line 78: may be create global `homePage = /` ?

kgevorkyan (approved):

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/FallbackView.kt
sanyavertolet: Not sure what you mean. New `FC`?

nulls: I guess he means
```kotlin
# Constants.kt
const val homePage = "/"
```

sanyavertolet: oh
not sure if it is needed

sanyavertolet: will merge now, later will fix if needed

Issue #2313, opened 2023-07-10
Prepare an API for massive (up to 50k of objects) post upload of OSV (extended). (@Cheshiriks under @nulls review).


nulls:

nulls:
Requirements:
- Support `Multi-Part` upload
- Support upload several files within a single request
- A file can contains a single vulnerability or array of vulnerabilities
- save-backend already has database with own vulnerabilities, we need to create entries there for uploaded vulnerabilities

PR #2401, opened 2023-08-02
Support upload vulnerability in COSV -- backend only

nulls:
### What's done:
- added repository for COSV (only interface)
- added S3 storage for COSV as implementation for repository
- added controllers to upload COSV as a String or a bunch of files
- added service to create Vulnerability entity in Save database when OSV is uploaded

It's part of #2313


save-backend/src/main/kotlin/com/saveourtool/save/osv/controllers/OsvController.kt
sanyavertolet, line 1: Isn`t it better to move all the processing logic to a new spring service? Like `save-preprocessor` but for Vulnerabilities

nulls: I'm agree, but it should be migrated with current vulnerability logic.

Parsing of OSV is just `reading a json file`. Json schema (OSV Schema) is in the dedicated project (osv4k).
OSV's endpoints just read a string content (or files), parse them and put as is to S3.

sanyavertolet: Are you planning to use database somewhere here? Cause if not, it might be a good idea to create `save-vuln-uploader` or smth.

I guess vulnerabilities will never be moved from gateway (unless some intern would like to show up)Are you planning to use database somewhere here? Cause if not, it might be a good idea to create `save-vuln-uploader` or smth.

I guess vulnerabilities will never be moved from gateway (unless some intern would like to show up)

nulls: Database is already used in backend and OSV uses it too: user uploads OSV, we creates vulnerability in databse and saves OSV in S3

sanyavertolet: Got it, then let's leave as it 🥹

nulls: `save-vuln-uploader` does what? Reads JSON using model and saves to S3 only? I don't think we need a separate application for it (or module even)

But I'm agree that whole logic with vulnerabilities should be in separate application as sandbox and demo. I can start with it in scope of this PR.

The plan:
1. Create a new module `save-vulnerabilities'. In phase one, it's just a module, not a separate application. More new logic to this module (for OSV).
2. Make a separate application from `save-vulnerabilities` and hide all logic from required for vulnerabilities from backend by rest requests to backend.
3. Move all logic required to vulnerabilities to the new application.
4. Profit.

I ready to make #1 and return to left points later.

sanyavertolet: Looks nice!

nulls:
@akuleshov7 please review the idea of OSV in Save:
- we store it in S3 as is (in future can be replaced by NoSql database)
- we create an entry in save's database with required info

What's we can do additionally:
- migrate our vulnerabilities to OSV too: create OSV object\entity instead of Save's one

TODO:
- Frontend part: loading by selecting file or by content (currently tested on swagger ui)

akuleshov7:
> @akuleshov7 please review the idea of OSV in Save:
> 
> * we store it in S3 as is (in future can be replaced by NoSql database)
> * we create an entry in save's database with required info
> 
> What's we can do additionally:
> 
> * migrate our vulnerabilities to OSV too: create OSV object\entity instead of Save's one
> 
> TODO:
> 
> * Frontend part: loading by selecting file or by content (currently tested on swagger ui)

Yeah, this is how we discussed it: 
- users anyway will provide us files 
- We store them in S3
- Make processing and store in DB entities

nulls:
> > @akuleshov7 please review the idea of OSV in Save:
> > 
> > * we store it in S3 as is (in future can be replaced by NoSql database)
> > * we create an entry in save's database with required info
> > 
> > What's we can do additionally:
> > 
> > * migrate our vulnerabilities to OSV too: create OSV object\entity instead of Save's one
> > 
> > TODO:
> > 
> > * Frontend part: loading by selecting file or by content (currently tested on swagger ui)
> 
> Yeah, this is how we discussed it:
> 
> * users anyway will provide us files
> * We store them in S3
> * Make processing and store in DB entities

Ok, will do in phase 2:
- refactor creating Vulnerabilities in Save by using OSV schema instead of save's one

save-osv/src/main/kotlin/com/saveourtool/save/osv/controllers/OsvController.kt
```diff
@@ -0,0 +1,87 @@
...
+@RestController
+@RequestMapping("/api/$v1/osv")
+class OsvController(
+    private val osvService: OsvService,
+) {
+    /**
+     * @param name vulnerability name in save db
+     * @return content of OSV
+     */
+    @GetMapping(path = ["/get-by-save-name/{name}"], produces = [MediaType.APPLICATION_JSON_VALUE])
```
akuleshov7, line 34: may be change an API for the future:
/get-by-id/save/{name}

It will be extended with:
/get-by-id/cve/{name}

save-osv/src/main/kotlin/com/saveourtool/save/osv/service/OsvService.kt
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+        .flatMap { id ->
+            osvStorage.downloadLatest(id)
+                .collectToInputStream()
+                .map { json.decodeFromStream<RawOsvSchema>(it) }
+        }
+
+    companion object {
+        private fun RawOsvSchema.toSaveVulnerabilityDto(): VulnerabilityDto = VulnerabilityDto(
```
akuleshov7, line 110: I quickly looked at this PR and cannot see adapter where you need to convert RAW -> DTO, where is it?

nulls: Currently there is no a dedicated service for it or any engine. Now it's just a private function which creates `VulnerabilityDto` from `RawOsvSchema`. It's this method

save-osv/src/main/kotlin/com/saveourtool/save/osv/controllers/OsvController.kt
nulls: actually, I added this method is for testing purpose -- to show that the link does exist.
I think we should return by a single ID (primary id).

Right now we create an entity `com.saveourtool.save.entities.vulnerabilities.Vulnerability` with generated name (id) and `vulnerabilityIdentifier` from original OSV file. We should not generate save's vulnerability id for such uploads (use id from OSV file).
We should generate ID only when we is original for vulnerability.

I can support using the original ID already in this PR

nulls: supported and refactored these methods

save-osv/src/main/kotlin/com/saveourtool/save/osv/service/OsvService.kt
nulls: introduced OsvProcessor with default implementation now.
In future, we can detect OsvProcessor by ID and finish parsing of custom fields and then convert to VulnerabilityDto.

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/ApiGroupsConfiguration.kt
```diff
@@ -21,7 +21,10 @@ class ApiGroupsConfiguration {
     fun openApiAll(): GroupedOpenApi? = GroupedOpenApi.builder()
         .group("all")
         .pathsToMatch("/api/**")
-        .packagesToScan("com.saveourtool.save.backend.controllers")
+        .packagesToScan(
+            "com.saveourtool.save.backend.controllers",
```
kgevorkyan, line 25: if you will have the time, could you please add to the readme the link to this file, I guess anyone will forget about this config, but now we start to actualize our readme's, probably it will be usefull too

kgevorkyan: to any readme

kgevorkyan: also, I didnt remember why I put it to the `utils` of backend , probably its better to put it to `config`

kgevorkyan (approved):

save-osv/src/main/kotlin/com/saveourtool/save/osv/service/OsvService.kt
```diff
@@ -0,0 +1,137 @@
...
+            name = "TO_GENERATE",  // will be generated on saving to database
+            vulnerabilityIdentifier = id,
+            progress = 0,
+            projects = emptyList(),
+            description = details,
+            shortDescription = summary.orEmpty(),
+            relatedLink = null,
+            language = VulnerabilityLanguage.OTHER,
```
akuleshov7: language can be mapped to existing languages (if found in our set)

nulls:
Need to support all mentioned prefix in schema

akuleshov7:
> Need to support all mentioned prefix in schema

Suggestion to think where to put it: to tags may be?
I am pretty sure, that WE will have different prefixes, as in OSV there is no java and cpp prefixes, for example. However is a bunch of Java CVEs

save-osv/src/main/kotlin/com/saveourtool/save/osv/processor/AbstractOsvProcessor.kt
nulls, line 1: discussed offline: need to migrate to COSV and use only fields from core schema without database\ecosystem specific fields

save-backend/src/main/kotlin/com/saveourtool/save/backend/utils/ApiGroupsConfiguration.kt
nulls: updated

save-osv/src/main/kotlin/com/saveourtool/save/osv/processor/AbstractOsvProcessor.kt
nulls: discussed additionally: we can ask the user to provide osv-processor id.
Will leave the OsvProcessor functionality, but will remove detecting id by osv.id

nulls:
need to migrate to `cosv`

nulls:
@akuleshov7 please review

PR #2410, opened 2023-08-03
User length name

Cheshiriks:
User length name

Снимок экрана 2023-08-03 в 23 56 32


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -100,8 +100,13 @@ class UsersDetailsController(
      */
     @PostMapping("/save")
     @PreAuthorize("isAuthenticated()")
+    @Suppress("MagicNumber")
     fun saveUser(@RequestBody newUserInfo: UserInfo, authentication: Authentication): Mono<StringResponse> =
             Mono.just(newUserInfo)
+                .filter { newUserInfo.name.length <= 22 }
```
sanyavertolet, line 106: **Must** be moved to `const val`

Or even reuse `isValidName`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputForms.kt
```diff
@@ -14,6 +14,10 @@ private const val EMAIL_PLACEHOLDER = "test@example.com"
 private const val NAME_TOOLTIP = "Allowed symbols: letters, digits, dots, hyphens and underscores." +
         "No dot, hyphen or underscore at the beginning and at the end of the line."
 
+private const val NAME_ORG_PROJECT_TOOLTIP = "Name must not be longer than 22 characters." +
```
sanyavertolet, line 17: 22 should be extracted

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/CreateOrganizationView.kt
```diff
@@ -148,7 +148,8 @@ class CreateOrganizationView : AbstractView<Props, OrganizationSaveViewState>()
                                             form = InputTypes.ORGANIZATION_NAME
                                             conflictMessage = state.conflictErrorMessage
                                             textValue = state.organizationDto.name
-                                            validInput = (state.organizationDto.name.isEmpty() || state.organizationDto.validateName()) && state.conflictErrorMessage == null
+                                            validInput = (state.organizationDto.name.isEmpty() || state.organizationDto.validateName() ||
+                                                    state.organizationDto.name.length <= 22) && state.conflictErrorMessage == null
```
sanyavertolet, line 152: Must be moved to const val

sanyavertolet (changes requested):

akuleshov7 (approved):

nulls:
do we need a separate PR for `val 22`?

sanyavertolet:
> do we need a separate PR for `val 22`?

not sure about separate PR, but it definitely should be fixed

PR #2414, opened 2023-08-04
Changed logic for max length name

Cheshiriks:
Changed logic for max length name


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/ValidationUtils.kt
```diff
@@ -8,6 +8,7 @@ package com.saveourtool.save.validation
  * Default amount of characters allowed for names
  */
 const val NAMING_ALLOWED_LENGTH = 64
+const val NAMING_MAX_LENGTH = 22
```
sanyavertolet, line 11: Why `allowed` is bigger than `max`?

sanyavertolet (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/ValidationUtils.kt
Cheshiriks: Because if something is impossible, but you really want it, then sometimes you can :)

PR #2415, opened 2023-08-04
COSV is source for vulnerabilities

nulls:
### What's done:
- Now COSV is main source for vulnerabilities 
- Entity `Vulnerability` is just a meta data with additional info about vulnerability


nulls:
covered by #2524

PR #2419, opened 2023-08-06
Propagating userInfo changes and updates to the App.kt

akuleshov7:
### What's done:
- Propagating all updates of userInfo to App.kt level except avatars (we need to have extra support for them)
- Adding INITIAL support for avatars selection

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
@@ -22,7 +22,6 @@ import react.dom.html.ReactHTML.main
 import web.cssom.*
 import web.html.InputType
 
-import kotlinx.browser.window
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 
@@ -31,6 +30,7 @@ val cardHeight: CSSProperties = jso {
 }
 
 val userSettingsView: FC<SettingsProps> = FC { props ->
+
     useBackground(Style.SAVE_LIGHT)
     main {
         className = ClassName("main-content")
@@ -48,6 +48,7 @@ val userSettingsView: FC<SettingsProps> = FC { props ->
                         rightSettingsColumn {
                             this.userInfo = props.userInfo
                             this.type = props.type
+                            this.userInfoSetter = props.userInfoSetter
                         }
                     } ?: main {
                         // FixMe: some light 404
@@ -66,7 +67,7 @@ typealias FieldsStateSetter = StateSetter<SettingsInputFields>
 @Suppress("MISSING_KDOC_CLASS_ELEMENTS")
 external interface SettingsProps : PropsWithChildren {
     /**
-     * Currently logged in user or null
+     * Currently logged-in user or null
      */
     var userInfo: UserInfo?
 
@@ -74,6 +75,14 @@ external interface SettingsProps : PropsWithChildren {
      * just a flag for a factory
      */
     var type: FrontendRoutes
+
+    /**
+     * After updating user information we will update userSettings without re-rendering the page
+     * PLEASE NOTE: THIS PROPERTY AFFECTS RENDERING OF WHOLE APP.KT
+     * IF YOU HAVE SOME PROBLEMS WITH IT, CHECK THAT YOU HAVE PROPAGATED IT PROPERLY:
+     * { this.userInfoSetter = (!) PROPS (!) .userInfoSetter }
+     */
+    var userInfoSetter: StateSetter<UserInfo?>
 }
 
 /**
@@ -133,11 +142,13 @@ fun saveUser(
     settingsInputFields: SettingsInputFields,
     setFieldsValidation: FieldsStateSetter,
 ) = useDeferredRequest {
+    // this new user info will be sent to backend and also will be set in setter,
+    // so frontend will recalculate it on the fly at least for SettingsView (need to extend it later)
+    val newUserInfo = settingsInputFields.toUserInfo(props.userInfo!!)
     val response = post(
         "$apiUrl/users/save",
         jsonHeaders,
-
-        Json.encodeToString(settingsInputFields.toUserInfo(props.userInfo!!)),
+        Json.encodeToString(newUserInfo),
         loadingHandler = ::loadingHandler,
         responseHandler = ::noopResponseHandler
     )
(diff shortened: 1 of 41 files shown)


akuleshov7:
@sanyavertolet please review: may be I have forgotten something while I was propagating UserInfo

sanyavertolet:
> @sanyavertolet please review: may be I have forgotten something while I was propagating UserInfo.
> But I tested it - it works well even with TopBar

Will take a look in 8 hours

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
```diff
@@ -66,14 +67,22 @@ typealias FieldsStateSetter = StateSetter<SettingsInputFields>
...
     var userInfo: UserInfo?
 
     /**
      * just a flag for a factory
      */
     var type: FrontendRoutes
+
+    /**
+     * After updating user information we will update userSettings without re-rendering the page
+     * PLEASE NOTE: THIS PROPERTY AFFECTS RENDERING OF WHOLE APP.KT
```
sanyavertolet, line 81: Does it? I have heart that react re-renders only components that have changes (their props or/and states).

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
```diff
@@ -66,14 +67,22 @@ typealias FieldsStateSetter = StateSetter<SettingsInputFields>
...
+     * IF YOU HAVE SOME PROBLEMS WITH IT, CHECK THAT YOU HAVE PROPAGATED IT PROPERLY:
+     * { this.userInfoSetter = (!) PROPS (!) .userInfoSetter }
+     */
+    var userInfoSetter: StateSetter<UserInfo?>
```
sanyavertolet, line 85: It also might be a good idea to pass a callback that triggers user info fetch instead of passing a `StateSetter`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
```diff
@@ -133,11 +142,13 @@ fun saveUser(
...
 ) = useDeferredRequest {
+    // this new user info will be sent to backend and also will be set in setter,
+    // so frontend will recalculate it on the fly at least for SettingsView (need to extend it later)
+    val newUserInfo = settingsInputFields.toUserInfo(props.userInfo!!)
     val response = post(
         "$apiUrl/users/save",
         jsonHeaders,
-
-        Json.encodeToString(settingsInputFields.toUserInfo(props.userInfo!!)),
+        Json.encodeToString(newUserInfo),
```
sanyavertolet, line 151: I dare you I double dare you to rename this method to `useSaveUser`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/right/Profile.kt
```diff
@@ -47,6 +45,7 @@ val profileSettingsCard: FC<SettingsProps> = FC { props ->
     // === states ===
     val avatarWindowOpen = useWindowOpenness()
     val (settingsInputFields, setSettingsInputFields) = useState(SettingsInputFields())
+    val (selectedAvatar, setSelectedAvatar) = useState<String>()
```
sanyavertolet, line 48: in this case, `selectedAvatar` is either `String` or `undefined`. It might be more clear and Kotlin-style to use `useState<String?>(null)` (if I am not mistaken)

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
akuleshov7: yes, I mean that it will affect all components, because we put props.userInfo mostly everywhere

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/right/Profile.kt
akuleshov7: yes, this one is good - will set `null` by for now, anyway it is not used yet

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
akuleshov7: ok, you won

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
akuleshov7: I actually don't get this suggestion 😄  - to my mind it is quite clear to propagate setter as a callback and update userInfo only on the high level of App.kt

Can you pls clarify?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
sanyavertolet: Got it

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
sanyavertolet: My idea is to fetch `UserInfo` from backend instead of updating `UserInfo` that was once fetched. We could create some `fetchUserInfo` callback that would fetch the data and update the `userInfo` that is passed everywhere. That might be a better idea sometimes

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsViewLeftColumn.kt
```diff
@@ -67,6 +67,13 @@ val leftSettingsColumn: FC<SettingsProps> = FC { props ->
                                         +(props.userInfo?.name ?: "")
                                     }
                                 }
+                                div {
+                                    className = ClassName("row justify-content-center")
+                                    h6 {
+                                        className = ClassName("mb-0 text-gray-800")
+                                        +(props.userInfo?.realName ?: "")
```
0x6675636b796f75676974687562, line 74:
```suggestion
                                        +props.userInfo?.realName.orEmpty()
```

akuleshov7: need a rule in diktat :)

PR #2421, opened 2023-08-07
Vulnerability filter search

sanyavertolet:
### What's done:
 * Implemented tags as a part of filter search on `backend`
 * Supported tags search on `frontend`
 * Added placeholder for `inputWithDebounce` empty option list

### How it looks:

https://github.com/saveourtool/save-cloud/assets/35039155/856986db-64f1-485c-99c3-b26749b40ebe

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -164,8 +164,7 @@ class VulnerabilityService(
                     } ?: cb.and()
 
                     val tagJoin: Join<LnkVulnerabilityTag, Tag> = root.join("tag")
-                    val inExpression = cb.`in`(tagJoin.get<String>("name"))
-                    val tagPredicate = inExpression.`in`(tags)
+                    val tagPredicate = cb.`in`(tagJoin.get<String>("name")).apply { tags.forEach(::value) }
 
                     cb.and(
                         namePredicate,
@@ -174,7 +173,9 @@ class VulnerabilityService(
                         cb.equal(vulnerabilityJoin.get<Vulnerability>("status"), status),
                     )
                 }
-            }.map { it.vulnerability }
+            }
+                .map { it.vulnerability }
+                .distinctBy { it.requiredId() }
         }
 
         val userIds = vulnerabilities.map { it.userId }
@@ -219,8 +220,7 @@ class VulnerabilityService(
                     }
 
                     val tagJoin: Join<LnkVulnerabilityTag, Tag> = root.join("tag")
-                    val inExpression = cb.`in`(tagJoin.get<String>("name"))
-                    val tagPredicate = inExpression.`in`(tags)
+                    val tagPredicate = cb.`in`(tagJoin.get<String>("name")).apply { tags.forEach(::value) }
 
                     cb.and(
                         namePredicate,
@@ -228,7 +228,9 @@ class VulnerabilityService(
                         cb.equal(vulnerabilityJoin.get<Vulnerability>("status"), status),
                     )
                 }
-            }.map { it.vulnerability }
+            }
+                .map { it.vulnerability }
+                .distinctBy { it.requiredId() }
         }
 
         val userIds = vulnerabilities.map { it.userId }
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -22,7 +22,7 @@ import org.springframework.transaction.annotation.Transactional
 import org.springframework.web.server.ResponseStatusException
 
 import java.time.LocalDateTime
-import javax.persistence.criteria.Join
+import javax.persistence.criteria.*
 
 import kotlin.random.Random
 import kotlinx.datetime.toJavaLocalDateTime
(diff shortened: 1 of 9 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -164,8 +164,7 @@ class VulnerabilityService(
                     } ?: cb.and()
 
                     val tagJoin: Join<LnkVulnerabilityTag, Tag> = root.join("tag")
-                    val inExpression = cb.`in`(tagJoin.get<String>("name"))
-                    val tagPredicate = inExpression.`in`(tags)
+                    val tagPredicate = cb.`in`(tagJoin.get<String>("name")).apply { tags.forEach(::value) }
```
Cheshiriks, line 167: why do you need this duplicate code?
Maybe you can do it simply:
```val tagPredicate = if (filters.tags.isEmpty()) else cb.`in`(tagJoin.get<String>("name")).apply { tags.forEach(::value) }```

File: save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt [hunk 2]
```diff
@@ -219,16 +220,17 @@ class VulnerabilityService(
                     }
 
                     val tagJoin: Join<LnkVulnerabilityTag, Tag> = root.join("tag")
-                    val inExpression = cb.`in`(tagJoin.get<String>("name"))
-                    val tagPredicate = inExpression.`in`(tags)
+                    val tagPredicate = cb.`in`(tagJoin.get<String>("name")).apply { tags.forEach(::value) }
```
Cheshiriks, line 223: Here too

Cheshiriks (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: The point is that if there are tags in filters, we need to interact with `lnk_vulnerability_tag` table, while `vulnerability` should be used if no tags are set

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Same story here

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/VulnerabilitiesFiltersRow.kt
```diff
@@ -35,46 +29,65 @@ val vulnerabilitiesFiltersRow: FC<NameFilterRowProps> = FC { props ->
...
+                        }
+                    }
+                }
+                div {
+                    className = ClassName("col-auto")
+                    inputWithDebounceForString {
+                        selectedOption = tagPrefix
+                        setSelectedOption = { setTagPrefix(it) }
+                        getUrlForOptionsFetch = { prefix -> "$apiUrl/tags/vulnerabilities?prefix=$prefix" }
+                        placeholder = "Tag..."
```
0x6675636b796f75676974687562, line 53: Up to you, but I'd rather use a single-character ellipsis, there's even a designated HTML entity `&hellip;`

0x6675636b796f75676974687562 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/table/filters/VulnerabilitiesFiltersRow.kt
sanyavertolet: The point is that this symbol is very hard to read in IDEA, so I would leave as is

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: And, these are different repositories, I understand.
I think then you can still do something like this:
val newRoot = if (filters.tags.isEmpty()) root else root.join("tags")

akuleshov7 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Fixed - take a look at my gpt4 solution and comment pls

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: Also done

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
Cheshiriks: Thank you! Looks great :)

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityCollectionView.kt
```diff
@@ -128,21 +128,38 @@ val vulnerabilityCollectionView: FC<FossGraphCollectionViewProps> = FC { props -
...
+                            filterValue?.let {
+                                setVulnerabilityFilters(filterValue)
+                                val fullUrl = buildString {
                                     "/${FrontendRoutes.VULN}"
-                                    filterValue?.let { append("?vulnerabilityName=$filterValue") }
+                                    var isFirst = true
+
+                                    if (filterValue.prefixName.isNotBlank()) {
+                                        if (isFirst) {
+                                            append("?")
```
nulls, line 141: @0x6675636b796f75676974687562 introduced http methods with typed parameters (`RequestUtils#254`) to avoid such manipulation. 
You can reuse it instead of checking `isFirst`

PR #2425, opened 2023-08-07
Added your organizations on index view

Cheshiriks:
- Added your organizations on index view
- Added controller for count if vulnerabilities

Снимок экрана 2023-08-08 в 00 00 45


sanyavertolet:
To my mind, `indexView` is now a bit narrow. Text columns are so small while there are quite many words. Moreover, your organisation card looks too small and not that expressive. Finally, paddings should be fixed as, for example, padding between `ProfileSettings` button and your title is less then between the title and organisation list

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardUser.kt
```diff
@@ -29,12 +31,10 @@ import kotlinx.datetime.toLocalDateTime
 
 // FixMe: List of organizations where user included, if not - link to creation of organization
```
sanyavertolet, line 32: Is this one also done?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardUser.kt
```diff
@@ -158,5 +168,38 @@ val cardUser: FC<IndexViewProps> = FC { props ->
                     }
             }
         }
+
+        div {
```
sanyavertolet, line 172: I'd also wrap the title and organisations with a `div` and set paddings in high-level `div`

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardUser.kt
Cheshiriks: Removed

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardUser.kt
Cheshiriks: Fixed

akuleshov7 (approved):

PR #2435, opened 2023-08-08
Added links to gitHub for About us block

Cheshiriks:
Added links to gitHub for About us block

Снимок333


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardAboutUs.kt
```diff
@@ -57,17 +59,30 @@ val cardAboutUs: FC<IndexViewProps> = FC { props ->
...
                 p {
-                    +"We kindly ask you not to break this service and report any problems that you will find to our Github. Please also read our"
+                    +"We kindly ask you not to break this service and report any problems that you will find to our Github."
+                }
+                Link {
+                    className = ClassName("btn btn-secondary rounded-pill")
+                    to = "https://github.com/saveourtool/save-cloud"
+                    fontAwesomeIcon(icon = faGithub)
+                    +"  Save-cloud"
+                }
```
sanyavertolet, line 77: This is not how you should use `Link`. Use `a` for this purpose. If I am not mistaken, `Link` is used for navigation inside your `Router`

sanyavertolet (approved):
We need to do something with card width. Three words on the same like is not so good.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/index/CardAboutUs.kt
Cheshiriks: fixed

PR #2436, opened 2023-08-08
Fixes for names (orgs, users, projects) validation and polishing of Setting Views

akuleshov7:
### What's done:
- Validation of new User, Organization and Project is now working correctly
- Settings view updated: organizations and email

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
@@ -77,25 +77,44 @@ class UserDetailsService(
         ?: Role.VIEWER
 
     /**
-     * @param user
+     * @param newUser
      * @param oldName
+     * @param oldUserStatus
      * @return UserSaveStatus
      */
     @Transactional
-    fun saveUser(user: User, oldName: String?): UserSaveStatus =
-            if (userRepository.validateName(user.name) != 0L) {
-                oldName?.let {
-                    userRepository.deleteHighLevelName(oldName)
-                    userRepository.saveHighLevelName(user.name)
-                    userRepository.save(user)
-                    UserSaveStatus.UPDATE
-                } ?: run {
-                    userRepository.save(user)
-                    UserSaveStatus.UPDATE
-                }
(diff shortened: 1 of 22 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/UserDetailsService.kt
```diff
@@ -82,17 +82,20 @@ class UserDetailsService(
      * @return UserSaveStatus
      */
     @Transactional
-    fun saveUser(user: User, oldName: String?): UserSaveStatus = if (oldName == null) {
-        userRepository.save(user)
-        UserSaveStatus.UPDATE
-    } else if (userRepository.validateName(user.name) != 0L) {
```
akuleshov7, line 88: Stupid bug: if you save new users, you don't validate. 
Usually it can be easily tested by simply creating a new user locally and checking it's functionality, but no one fucking cared 🤣

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/right/Email.kt
```diff
@@ -9,23 +9,65 @@ import com.saveourtool.save.frontend.components.views.usersettings.SettingsProps
...
+        h3 {
+            className = ClassName("mt-2 mr-2 text-gray-800")
+            +"Login and Email"
+        }
+
+        Link {
+            to = "/${FrontendRoutes.TERMS_OF_USE}"
+            buttonBuilder(
+                "terms of usage",
+                style = "outline-secondary rounded-pill btn-sm",
```
sanyavertolet, line 50: `style` should have only `secondary`, while all the rest should be set either with flags or with `classes`

sanyavertolet (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/right/Email.kt
akuleshov7: good catch, fixed

PR #2438, opened 2023-08-09
Vulnerability filters part III

sanyavertolet:
Part of #2430 

### What's done:
* Added author name filter
* Added organization name filter
* Added `isDisabled` flag to `InputWithDebounceProps`

### TODO:
 * Display vulnerability filters in table

### How it looks:

https://github.com/saveourtool/save-cloud/assets/35039155/9a5c74de-8a7b-419a-adea-b8fdeeedc62f

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputWithDebounce.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputWithDebounce.kt
@@ -2,20 +2,20 @@
 
 package com.saveourtool.save.frontend.components.inputform
 
+import com.saveourtool.save.entities.OrganizationDto
+import com.saveourtool.save.frontend.components.basic.renderAvatar
 import com.saveourtool.save.frontend.utils.*
 import com.saveourtool.save.frontend.utils.noopLoadingHandler
 import com.saveourtool.save.frontend.utils.noopResponseHandler
 import com.saveourtool.save.info.UserInfo
 import com.saveourtool.save.utils.DEFAULT_DEBOUNCE_PERIOD
-import com.saveourtool.save.v1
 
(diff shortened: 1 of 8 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputWithDebounce.kt
```diff
@@ -89,35 +99,75 @@ external interface InputWithDebounceProps<T> : PropsWithChildren {
...
     }
 }
 
+/**
+ * Default renderer for [inputWithDebounceForOrganizationDto]
+ *
+ * @param childrenBuilder [ChildrenBuilder] instance
+ * @param organizationDto [OrganizationDto] to display
+ */
+internal fun renderOrganizationWithAvatar(childrenBuilder: ChildrenBuilder, organizationDto: OrganizationDto) {
```
Cheshiriks, line 137: I think this function is a duplicate. We already have a ```renderOrganizationWithName``` method

Cheshiriks (changes requested):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/inputform/InputWithDebounce.kt
sanyavertolet: This method has its signature and it is used for `inputWithDebounce`. it is nice to use it as `::renderOrganizationWithAvatar`.

sanyavertolet: As you can see, this file contains three `render...` functions - all of them are used together with `inputWithDebounce`

Cheshiriks (approved):

PR #2457, opened 2023-08-14
Hotfix for vulnerability tags for unauthorized

sanyavertolet:
### What's done:
* Added tags fetch on vuln fetch for unauthorized

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
@@ -42,6 +42,14 @@ class VulnerabilityService(
     private val lnkVulnerabilityTagRepository: LnkVulnerabilityTagRepository,
     private val tagRepository: TagRepository,
 ) {
+    @Suppress("TYPE_ALIAS")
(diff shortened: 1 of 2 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
```diff
@@ -42,6 +42,14 @@ class VulnerabilityService(
     private val lnkVulnerabilityTagRepository: LnkVulnerabilityTagRepository,
     private val tagRepository: TagRepository,
 ) {
+    @Suppress("TYPE_ALIAS")
+    private val getTagMap: List<Vulnerability>.() -> Map<Vulnerability, Set<Tag>> = {
```
Cheshiriks, line 46: I think you can try changing this to a private function

Cheshiriks (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/vulnerability/VulnerabilityService.kt
sanyavertolet: I guess it is fine

PR #2470, opened 2023-08-16
Settings fields editing

akuleshov7:
### What's done:
- Fixes for editing of field on settings view
- Fix for profile view rendering
- Stubs for validation of input values

image


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
```diff
@@ -34,8 +34,6 @@ val cardHeight: CSSProperties = jso {
 }
 
 val userSettingsView: FC<SettingsProps> = FC { props ->
-    // This is needed for us to wait for userInfo uploaded
-    val (loadedUserInfo, setLoadedUserInfo) = useStateFromProps(props.userInfo ?: UserInfo(""))
```
akuleshov7, line 38: here @sanyavertolet . Previously it was:

App.kt -> BasicRouting { userInfo } -> userSettingsView { userInfo } -> userSettingsView { useStateFromProps(userInfo) } -> RightCol { userInfo }

Cheshiriks (approved):

PR #2482, opened 2023-08-21
Changed upload avatar method for users

Cheshiriks:


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
```diff
@@ -126,7 +127,7 @@ internal class AvatarController(
             blockingToMono {
                 when (type) {
                     AvatarType.ORGANIZATION -> organizationService.updateAvatarVersion(owner)
-                    AvatarType.USER -> userDetailsService.updateAvatarVersion(owner)
+                    AvatarType.USER -> userDetailsService.updateAvatarVersion(authentication.username())
```
akuleshov7, line 130: may be add unauthorised error?

akuleshov7 (approved):

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/AvatarController.kt
Cheshiriks: Yes, in the service we are looking for a user by name in the database.
Or do you mean something like this: _owner == authentication.username()_ ?

Cheshiriks: I just don't see the point of passing the name from the front at all. After all, we can get a nickname from _authentication_. So no one can change someone else's avatar.

PR #2490, opened 2023-08-23
Polishing rendering of vulnerability view part #1

akuleshov7:
And + two animations: 
- on timeline
- on addition of libs

image


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
```diff
@@ -173,53 +174,53 @@ val vulnerabilityView: FC<VulnerabilitiesViewProps> = FC { props ->
 
     useOnce(fetchVulnerability)
 
-    div {
-        className = ClassName("")
+    val isSuperAdmin = props.currentUserInfo?.globalRole?.isHigherOrEqualThan(Role.SUPER_ADMIN) == true
```
Cheshiriks, line 177: ```val isSuperAdmin = props.currentUserInfo?.isSuperAdmin() == true```

Cheshiriks (approved):

PR #2491, opened 2023-08-23
First steps into localization

sanyavertolet:
### What's done:
 * Added `i18n`
 * Added `react-i18n`
 * Added kotlin wrapper for `useTranslation` hook

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
@@ -1,0 +1,41 @@
+/**
+ * File containing initialization of i18next
+ */
+
+package com.saveourtool.save.frontend.externals.i18next
+
+/**
+ * Function that encapsulates i18n initialization.
+ *
+ * @param isDebug flag to set debug mode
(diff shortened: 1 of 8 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+@Suppress("UNUSED_PARAMETER")
+fun initI18n(isDebug: Boolean = false, interpolationEscapeValue: Boolean = false) {
+    js("""
+        var i18n = require("i18next");
+        var reactI18n = require("react-i18next");
+        var resources = {
+            en: {
+                translation: {
+                    "Notifications": "Notifications"
```
Cheshiriks, line 21: Then there will be a json reading?

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
sanyavertolet: Yep, there's an ability to pass a path to file

PR #2492, opened 2023-08-24
Language selector

sanyavertolet:
### What's done:
 * Added language selector
 * Added several languages to be supported
 * Fixed invoke operator in `Translation` class
 * Added language selector to `TopBar`

### How it looks:
https://github.com/saveourtool/save-cloud/assets/35039155/664f8353-e307-4539-9454-e0844d4d065e


akuleshov7 (approved):
Looks good, but may be you will add translation for different files?

PR #2497, opened 2023-08-26
Tags should not be clickable without permissions

akuleshov7:
### What's done:
- if user has no permissions to edit tags, he will be redirected to FILTERED page with vulnerabilities by proper URL: `vuln/list?tags=hu`

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/FrontendRoutes.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/FrontendRoutes.kt
@@ -39,6 +39,7 @@ enum class FrontendRoutes(val path: String) {
     SETTINGS_ORGANIZATIONS("$SETTINGS/organizations"),
     SETTINGS_PROFILE("$SETTINGS/profile"),
     SETTINGS_TOKEN("$SETTINGS/token"),
+    SIGNLE_VULNERABILITY("$VULN/collection"),
     TERMS_OF_USE("terms-of-use"),
     VULN("vuln"),
     VULNERABILITIES("$VULN/list"),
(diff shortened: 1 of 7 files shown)


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/FrontendRoutes.kt
```diff
@@ -39,6 +39,7 @@ enum class FrontendRoutes(val path: String) {
     SETTINGS_ORGANIZATIONS("$SETTINGS/organizations"),
     SETTINGS_PROFILE("$SETTINGS/profile"),
     SETTINGS_TOKEN("$SETTINGS/token"),
+    SIGNLE_VULNERABILITY("$VULN/collection"),
```
sanyavertolet, line 42: Should `SINGLE_VULNERABILITY` be `/collection`?

sanyavertolet (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/validation/FrontendRoutes.kt
akuleshov7: no, why? 
/vuln/collection, because it's in /vuln domain

sanyavertolet: I mean, _collection_ is not about _single_ vulnerability

akuleshov7: > I mean, _collection_ is not about _single_ vulnerability

ah, got it. Anyway, users will not see it on UI (only search panel): they will see vuln/list/NAME

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityTagsComponent.kt
```diff
@@ -106,13 +111,18 @@ val vulnerabilityTagsComponent: FC<VulnerabilityTagsComponentProps> = FC { props
     div {
         className = ClassName("col pl-0")
         props.tags.forEach { tag ->
+            val tagEditable = props.userInfo?.name == props.vulnerability.name
```
nulls, line 114: is a tag editable when userName is equal to vulnerability name?

akuleshov7: fixed

PR #2499, opened 2023-08-28
Topbar localization

sanyavertolet:
### What's done:
 * Removed `width` from `TopBarLink` - now it is useless
 * Fixed width of top bar links
 * Added russian translation

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
@@ -49,3 +49,38 @@ const val TABLE_HEADERS_EN = """
     "Tag": "Tag"
 }
 """
+
+const val TOPBAR_RU = """
+{
+    "Demo": "Демо",
+    "CPG": "CPG",
+    "Awesome Benchmarks": "Awesome Benchmarks",
+    "Try SAVE format": "Попробуйте SAVE-формат",
+    "Projects board": "Проекты",
+    "Contests": "Соревнования",
+    "About us": "О нас",
+    "Propose vulnerability": "Предложите уязвимость",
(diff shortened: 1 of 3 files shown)


Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
```diff
@@ -49,3 +49,38 @@ const val TABLE_HEADERS_EN = """
...
+    "About us": "О нас",
+    "Propose vulnerability": "Предложите уязвимость",
+    "Vulnerabilities": "Уязвимости",
+    "Top Rating": "Рейтинг",
+    "SAVE Projects list": "SAVE-проекты",
+    "Vulnerabilities list": "Список уязвимостей"
+}
+"""
+
+const val TOPBAR_EN = """
```
nulls, line 70: do we need it? why fallback is not working here?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
sanyavertolet: It works, but if we want our Chinese colleagues help with translation, we should provide them a good-looking example of language they can understand

nulls: `English words` is a key already. I don't think that additional mapping `A` to `A` will help them, but it adds work for you

sanyavertolet: Sometimes keys do not match with values

I'd like to leave as is for now

nulls (approved):

PR #2507, opened 2023-08-29
Better cookies

sanyavertolet:
### What's done:
 * Added better wrappers for `js-cookie`
 * Added cookie policy view
 * Added localization for cookie policy view
 * Added cookie banner
 * Added localization for cookie banner

### How it looks:
screenshot
screenshot


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
```diff
@@ -84,3 +84,27 @@ const val TOPBAR_EN = """
     "Vulnerabilities list": "Vulnerabilities list"
 }
 """
+
+const val COOKIES_EN = """
```
kgevorkyan, line 88: may be some proofread by chartgpt?

kgevorkyan (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/CookieBanner.kt
```diff
@@ -0,0 +1,52 @@
...
+            className = ClassName("fixed-bottom bg-light px-4 d-flex justify-content-between align-items-center")
+            style = jso {
+                zIndex = (MAX_Z_INDEX - 1).unsafeCast<ZIndex>()
+            }
+            div {
+                className = ClassName("pt-2")
+                markdown("We value your privacy".t().trimIndent())
+            }
+            div {
+                buttonBuilder("Decline".t(), "secondary", classes = "mx-2") {
```
nulls, line 38: do we need allow to decline? maybe use phrase like `You agree with using cookies by continue to use our site`?

nulls: because we will have to check `cookie.isAccepted()` in all places when we will use cookies

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/cookie/Cookie.kt
```diff
@@ -2,9 +2,161 @@
...
+ * Class that encapsulates the cookie information
+ *
+ * @property key [String] cookie name
+ * @property expires amount of days before a cookie is considered to be expired, [DEFAULT_EXPIRES] by default
+ */
+sealed class CookieKeys(val key: String, val expires: Int = DEFAULT_EXPIRES) {
+    /**
+     * Cookie that indicates that user as accepted cookie policy
+     */
+    object IsCookieOk : CookieKeys("isCookieOk")
```
nulls, line 92: why sealed classes? Suggest to use Enum and collect all cookies name with description to print these info on cookies page. It allows to avoid changing cookies page when a new cookie is added

akuleshov7 (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/CookieBanner.kt
sanyavertolet: > do we need allow to decline? maybe use phrase like `You agree with using cookies by continue to use our site`?

It might break some countries laws :)

sanyavertolet: > because we will have to check `cookie.isAccepted()` in all places when we will use cookies

Yep, we will

But on the other hand, this check is quite light, so it should not cause a lack of performance

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/cookie/Cookie.kt
sanyavertolet: Our translation is not ready yet. Will do it in next PR

And there probably was a point of why I decided to use `sealed class`, but I do not remember the reason :)

Perhaps it gives us more flexible interface (e.g. create your own cookies during runtime)

sanyavertolet: but for now it seems that there is no actual profit lol

sanyavertolet: Oh sure, I wanted it to have configurable attributes (e.g. `expires` and so on)

So I would leave as is

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/locales/Translations.kt
sanyavertolet: Good point, will do so

sanyavertolet: Done

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/CookieBanner.kt
nulls: > > do we need allow to decline? maybe use phrase like `You agree with using cookies by continue to use our site`?
> 
> It might break some countries laws :)

I'm not sure, we notice user and if user agrees, user continues to use the site.

nulls: for example image

PR #2512, opened 2023-08-30
Fixed basic auth in gateway

nulls:
### What's done:
- backend sends token to gateway
- renamed `AuthenticationUserDetails` to `SaveUserDetails`

### DIFF
--- a/api-gateway/src/main/resources/logback.xml
+++ b/api-gateway/src/main/resources/logback.xml
@@ -5,7 +5,7 @@
         </encoder>
     </appender>
 
-    <root level="DEBUG">
+    <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
 </configuration>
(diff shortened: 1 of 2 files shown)


api-gateway/src/main/resources/logback.xml
```diff
@@ -5,7 +5,7 @@
         </encoder>
     </appender>
 
-    <root level="info">
+    <root level="DEBUG">
```
kgevorkyan, line 8: better leave debug?

kgevorkyan (approved):

api-gateway/src/main/resources/logback.xml
sanyavertolet: I guess debug logging is fine as the product is still in development mode. In this case, it might be easier to trace errors on prod

sanyavertolet (approved):

api-gateway/src/main/resources/logback.xml
nulls: actually, I forgot to rollback this change.
Debug is useful, we have enable debug logging for `logging.level.org.springframework` in backend, demo and orchestrator.
Will do in the same way

PR #2523, opened 2023-09-01
Added bulk upload files

Cheshiriks:
- Added bulk upload files

Снимок

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
@@ -226,6 +226,48 @@ internal class OrganizationController(
             ResponseEntity.ok("Organization updated")
         }
 
+    @PostMapping("/{organizationName}/manage-bulk-upload-permission")
+    @RequiresAuthorizationSourceHeader
+    @PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")
+    @Operation(
+        method = "POST",
+        summary = "Make an organization to be able to bulk upload files.",
+        description = "Make an organization to be able to bulk upload files.",
+    )
+    @Parameters(
+        Parameter(name = "organizationName", `in` = ParameterIn.PATH, description = "name of an organization", required = true),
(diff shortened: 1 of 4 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -226,6 +226,48 @@ internal class OrganizationController(
...
+        organizationName
+    )
+        .flatMap {
+            organizationService.findByNameAndCreatedStatus(organizationName).toMono()
+        }
+        .switchIfEmptyToNotFound {
+            "No organization with name $organizationName was found."
+        }
+        .filter {
+            organizationPermissionEvaluator.hasGlobalRoleOrOrganizationRole(authentication, it.name, Role.SUPER_ADMIN)
```
sanyavertolet, line 259: Can be omitted here, as there is `@PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")`

sanyavertolet (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
Cheshiriks: Agree. Removed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/OrganizationController.kt
```diff
@@ -239,7 +239,6 @@ internal class OrganizationController(
         Parameter(name = "isAbleToToBulkUpload", `in` = ParameterIn.QUERY, description = "new flag for bulk upload files ability", required = true),
     )
     @ApiResponse(responseCode = "200", description = "Successfully changed ability to bulk upload files.")
-    @ApiResponse(responseCode = "403", description = "Could not change ability to bulk upload files due to lack of permission.")
```
sanyavertolet: I guess `403` is returned if user is not `SUPER_ADMIN`

Cheshiriks: Returned it

PR #2527, opened 2023-09-01
Refactoring vulnerability view to split header in the separate file

akuleshov7:
- Just simple refactoring: moved a block of of code to a separate method
- Also fixed links to /vuln/collection on some pages


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
```diff
@@ -334,6 +152,17 @@ external interface VulnerabilitiesViewProps : Props {
     var currentUserInfo: UserInfo?
 }
 
+/**
+ * @property isSuperAdmin
+ * @property isOwner
+ * @property isParticipant
+ */
+internal data class Permissions(
```
nulls, line 160: maybe move it to save-common and return it together with vulnerability?

PR #2536, opened 2023-09-04
Bugfix for editing of vulnerability based on permissions

akuleshov7:
### What's done:
- permission bugfix
- added a functionality to show editing projects components only when edit button is pressed


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
```diff
@@ -81,6 +81,6 @@ class UsersController(
     }
 
     companion object {
-        private val roleForNewUser = listOf(Role.VIEWER).joinToString(",")
```
akuleshov7, line 84: this is weird. Also I still do not know how just "VIEWER" without ROLE prefix worked

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
```diff
@@ -172,3 +177,8 @@ private fun ChildrenBuilder.languageSpan(language: VulnerabilityLanguage) {
         +language.value
     }
 }
+
+internal fun hasRightsToEdit(currentUserInfo: UserInfo?, vulnerability: VulnerabilityDto) =
+    // globalRole is enriched on the App.kt level with "users/global-role" (based on ROLE from spring security authority)
+    // and is not stored in DB, so we cannot check userInfo directly, because in the getAllParticipants() users will have this field null
+    currentUserInfo.isSuperAdmin() || currentUserInfo?.originalLogins in vulnerability.getAllParticipants().map { it.originalLogins }
```
akuleshov7, line 184: I do not want to touch globalRole and breaking something, so this is my choice :)

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
nulls: it's value for database entity, we will store it as `ROLE_VIEWER` in database.
I guess it can be fixed only on Spring level by changing `com.saveourtool.save.authservice.utils.SaveUserDetails#getAuthorities` -- here we create `SimpleGrantedAuthority` and need to add prefix there

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
```diff
@@ -172,3 +177,8 @@ private fun ChildrenBuilder.languageSpan(language: VulnerabilityLanguage) {
         +language.value
     }
 }
+
+internal fun hasRightsToEdit(currentUserInfo: UserInfo?, vulnerability: VulnerabilityDto) =
+    // globalRole is enriched on the App.kt level with "users/global-role" (based on ROLE from spring security authority)
+    // and is not stored in DB, so we cannot check userInfo directly, because in the getAllParticipants() users will have this field null
```
nulls, line 183: globalRole is `User.role`

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
nulls, line 184: why we check originalLogins? we can check only `currentUserInfo.name`

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
nulls: btw, checked -- old users have prefix `ROLE_` in database, the new ones -- don't

akuleshov7: Actually no, I see in permission validation endpoint that we check the value stored in Spring security with “as spring value” with ROLE_ prefix

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
akuleshov7: Yes and no. GlobalRole is not simple set from role - basically enriched from authority. But initial value stored in User.role, yes. 

But in this comment I am pointing that globalRole is not set in UserDetails directly and details are only enriched later with this value on the frontend with a separate request

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/internal/UsersController.kt
nulls: > So basically in role field it is much logical to write it with ROLE_ prefix

I'm not sure that it's more logical (we can read as enum from database and make logic on it), but historically we do it. Will fix existed values in prod database

nulls: > Actually no, I see in permission validation endpoint that we check the value stored in Spring security with “as spring value” with ROLE_ prefix.

I see that we search a Role (our enum) using Spring security values (taking into account prefix) and then we validate permission using our Role only.

> also tested - mapping in the frontend does not work without a ROLE_ in a role field in a database

Yes, it's a bug. Will fix entries in database

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
nulls: raised to look at it later: https://github.com/saveourtool/save-cloud/issues/2540

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
akuleshov7: yes, true, I misguided myself by thinking that it can be changed and will be reflected on the frontend but not on the backend. But it is not possible, you are right

akuleshov7: fixed

nulls (approved):

Issue #2451, opened 2023-08-10
Limit user's registration in the Vulnerabilities Archive


akuleshov7:
1. New Users and Organizations should be approved after registration in the system. 
2. We should be able to ban users
3. All checks should be done inside a common logic in gateway

#### upd:
Vlad has partially done the task. Here is what remains:
- [x] https://github.com/saveourtool/save-cloud/pull/2546
- [x] https://github.com/saveourtool/save-cloud/pull/2552
- [x] https://github.com/saveourtool/save-cloud/pull/2553
- [x] https://github.com/saveourtool/save-cloud/pull/2560
- [x] https://github.com/saveourtool/save-cloud/pull/2562
- [x] https://github.com/saveourtool/save-cloud/pull/2565

### What's done:
- https://github.com/saveourtool/save-cloud/pull/2503
- https://github.com/saveourtool/save-cloud/pull/2489

nulls:
btw, we have `/api/` which is hidden under authorization, but also we have a set of endpoints which we open by listing them in the special list in WebConfig.
Maybe introduce a new sub-path: `/public-api/` and open it instead of a list of open endpoints

akuleshov7:
@Cheshiriks is it done?

sanyavertolet:
Done

PR #2552, opened 2023-09-06
Not approved should work

sanyavertolet:
This PR is a part of #2451 

### What's done:
 * Started to set user status on backend depending on current status
 * Added `useUserStatusRedirects` custom hook to encapsulate redirects on App FC
 * Added `useRedirectToIndexIf` custom hook to encapsulate redirects from "special" pages to index view


Cheshiriks (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/UsersDetailsController.kt
```diff
@@ -103,46 +103,50 @@ class UsersDetailsController(
...
+        .filter { newUserInfo.name.isValidLengthName() }
+        .switchIfEmptyToResponseException(HttpStatus.CONFLICT) {
+            UserSaveStatus.INVALID_NAME.message
+        }
+        .blockingMap {
+            val user: User = userRepository.findByName(newUserInfo.oldName ?: newUserInfo.name).orNotFound()
+            val oldStatus = user.status
+            if (user.id == authentication.userId()) {
+                val newStatus = when (oldStatus) {
+                    UserStatus.CREATED -> UserStatus.NOT_APPROVED
```
nulls, line 116:
```suggestion
                    UserStatus.CREATED, UserStatus.NOT_APPROVED -> UserStatus.NOT_APPROVED
```

nulls (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/UsersDetailsController.kt
sanyavertolet: There is actually no way to change user data of `NOT_APPROVED` on frontend so I decided to forbid it for api as well

PR #2556, opened 2023-09-06
Added user permissions in cosv service

Cheshiriks:
Added user permissions in cosv service

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/security/UserPermissionEvaluator.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/security/UserPermissionEvaluator.kt
@@ -1,7 +1,8 @@
 package com.saveourtool.save.backend.security
 
 import com.saveourtool.save.authservice.utils.username
-import com.saveourtool.save.backend.service.LnkUserOrganizationService
+import com.saveourtool.save.backend.repository.LnkUserOrganizationRepository
+import com.saveourtool.save.entities.OrganizationStatus
 import com.saveourtool.save.info.UserPermissions
 import com.saveourtool.save.info.UserPermissionsInOrganization
 import org.springframework.security.core.Authentication
@@ -12,7 +13,8 @@ import org.springframework.stereotype.Component
  */
 @Component
 class UserPermissionEvaluator(
-    private var lnkUserOrganizationService: LnkUserOrganizationService,
+    private var lnkUserOrganizationRepository: LnkUserOrganizationRepository,
+    // private var lnkUserOrganizationService: LnkUserOrganizationService,
 ) {
     /**
      * @param authentication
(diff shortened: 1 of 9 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/security/UserPermissionEvaluator.kt
```diff
@@ -12,7 +13,8 @@ import org.springframework.stereotype.Component
  */
 @Component
 class UserPermissionEvaluator(
-    private var lnkUserOrganizationService: LnkUserOrganizationService,
+    private var lnkUserOrganizationRepository: LnkUserOrganizationRepository,
+    // private var lnkUserOrganizationService: LnkUserOrganizationService,
```
nulls, line 17: service can be removed?

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
```diff
@@ -55,6 +56,7 @@ data class UserInfo(
     val freeText: String? = null,
     val realName: String? = null,
     val createDate: LocalDateTime? = null,
+    val userPermissions: UserPermissions = UserPermissions.empty,
```
nulls, line 59: having such optional fields complicates the logic -- `UserInfo` doesn't guarantee that it has `userPermissions`, because it's required using some another endpoint.

Can be keep `UserInfo` is consistent? If we need `userPermissions`, we can make an another request to backend or create a new data class with extended info about user

nulls (changes requested):

save-backend/src/main/kotlin/com/saveourtool/save/backend/security/UserPermissionEvaluator.kt
Cheshiriks: Removed

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/info/UserInfo.kt
Cheshiriks: Ok, agree. Removed it

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/CosvService.kt
```diff
@@ -56,6 +59,12 @@ class CosvService(
     ): Flux<String> {
         val user = backendService.getUserByName(authentication.name)
         val organization = backendService.getOrganizationByName(organizationName)
+        val userPermissions = backendService.getUserPermissionsByOrganizationName(authentication, organizationName)
+
+        if (userPermissions.inOrganizations[organizationName]?.canDoBulkUpload != true) {
+            throw ResponseStatusException(HttpStatus.FORBIDDEN, "You do not have permission to upload COSV files on behalf of this organization: $organizationName")
```
nulls, line 65: can we return here just `Flux.error()`?

Cheshiriks: Changed it

PR #2557, opened 2023-09-06
Added copying files to minio for local run

### LINKED ISSUE #2554: Start Minio using files from resources
Need to support start minio with files from test resources for local run

nulls:
### What's done:
- added a new service to copy files to minio on start

It closes #2554

### DIFF
--- a/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
+++ b/buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
@@ -159,6 +159,23 @@ fun Project.createStackDeployTask(profile: String) {
                            |      MINIO_ROOT_USER: admin
                            |      MINIO_ROOT_PASSWORD: adminadmin
                            |
+                           |  minio-startup:
+                           |    image: minio/mc:latest
+                           |    depends_on:
+                           |      - minio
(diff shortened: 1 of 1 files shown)


sanyavertolet (approved):

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
```diff
@@ -159,6 +159,23 @@ fun Project.createStackDeployTask(profile: String) {
                            |      MINIO_ROOT_USER: admin
                            |      MINIO_ROOT_PASSWORD: adminadmin
                            |
+                           |  minio-startup:
```
akuleshov7, line 162: I thought you will copy it inside the task `startMinio`

but this is even better

akuleshov7 (approved):

buildSrc/src/main/kotlin/com/saveourtool/save/buildutils/DockerStackConfiguration.kt
nulls: startMinio -- it;'s gradle task. we still use this name.
but we can just copy a file to S3, need some application which does it (mc).
but also we shouldn't use folder from source code as volume for S3 -- new files are added there

PR #2559, opened 2023-09-06
Editing view preparations: propagating of edit flag to children components

akuleshov7:
### What's done:
- Minor Spring Security Timeout update (not tested)
- Propagating edit button actions to have a common editable page
- Adding creation date to timeline (so it's always shown)
- Adding ribbon for approved/not approved
- Fix for editing short and long description
- Minor fixes

https://github.com/saveourtool/save-cloud/assets/58667063/c2b7c74c-1c54-4490-a621-5113ddfca1fb

### DIFF
--- a/api-gateway/src/main/resources/application.yml
+++ b/api-gateway/src/main/resources/application.yml
@@ -117,3 +117,8 @@ spring:
             scope:
               - user_info
 ---
+# logout timeout
+server:
+  servlet:
+    session:
+      timeout: 15m
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TimelineComponent.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TimelineComponent.kt
@@ -2,6 +2,8 @@
 
 package com.saveourtool.save.frontend.components.basic
 
+import com.saveourtool.save.entities.vulnerability.VulnerabilityDateType
+import com.saveourtool.save.entities.vulnerability.VulnerabilityDto
 import com.saveourtool.save.frontend.utils.buttonBuilder
 import react.*
 import react.dom.html.ReactHTML.div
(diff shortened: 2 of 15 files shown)


api-gateway/src/main/resources/application.yml
```diff
@@ -117,3 +117,8 @@ spring:
             scope:
               - user_info
 ---
+# logout timeout
+server:
+  servlet:
+    session:
+      timeout: 15m
```
sanyavertolet, line 124: finally

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TimelineComponent.kt
```diff
@@ -28,45 +30,50 @@ val timelineComponent: FC<TimelineComponentProps> = FC { props ->
...
-                        .sortedBy { it.second }
-                        .forEach { (label, dateTime) ->
-                            div {
-                                className = ClassName("step $hoverable")
+                    className = ClassName("line")
+                }
+                props.dates
+                    .plus(
+                        VulnerabilityDateType.SUBMITTED.value to
+                                (props.vulnerability.creationDateTime ?: LocalDateTime(0, 1, 1, 0, 0, 0, 0))
```
sanyavertolet, line 49: Do we want such a placeholder or it is better to hide `SUBMITTED` date?

sanyavertolet (approved):

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/TimelineComponent.kt
akuleshov7: No, not submitted, we would like to have a timeline always. 
A different point is that we need to move it to constant, I guess

PR #2560, opened 2023-09-07
Add GitHub on registration

sanyavertolet:
This PR is a part of #2451 

### What's done:
 * Added `GitHub` input field to `registrationView`
 * Added website input field to `registrationView`
 * Made `GitHub` required for registration

### How it looks:
screenshot


nulls (approved):

akuleshov7:
Let's make all fields NOT required (except login), but we will explicitly write it explanation WHY WE ASK FOR IT - add text that now the access to our platform is limited that's why we carefully review all registrations and approve them, so we need as much info about you as possible

PR #2573, opened 2023-09-11
Hotfix for gateway user status check

sanyavertolet:
### What's done:
 * Added `/api/$v1/avatar/**` to `allowedForInactiveEndpoints`
 * Supported unauthorized users in gateway `WebSecurityConfig`

### DIFF
--- a/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
+++ b/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
@@ -81,8 +81,8 @@ class WebSecurityConfig(
                 // any authentication data at all.
                 // backend returns 401 for those endpoints that require authentication
                 .pathMatchers(*allowedForInactiveEndpoints.toTypedArray()).access(::defaultAuthorizationDecision)
-                .pathMatchers("/api/**").access { authorization, authorizationContext ->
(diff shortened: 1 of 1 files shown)


Cheshiriks (commented):
What about the 403 error for authorized users?

sanyavertolet:
> What about the 403 error for authorized users?

I was unable to reproduce

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
```diff
@@ -177,18 +185,25 @@ fun passwordEncoder(): PasswordEncoder = PasswordEncoderFactories.createDelegati
  * @param authorizationContext
  * @return [Mono] of [AuthorizationDecision]
  */
-private fun defaultAuthorizationDecision(
+private fun authorizationManagerAuthorizationDecision(
```
akuleshov7, line 188: can you please move it below method `defaultAuthorizationDecision`, where it is called ?

akuleshov7 (approved):

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
sanyavertolet: Moved all dependant methods below `defaultAuthorizationDecision` and `userStatusBasedAuthorizationDecision`

PR #2575, opened 2023-09-11
Hotfix for 403 when user changed name

sanyavertolet:
### What's done:
 * Moved logic connected with authorization cast to `BackendService`
 * Replaced `BackendService#findByName` with `BackendService#findByAuthentication`

### DIFF
--- a/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
+++ b/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
@@ -4,6 +4,7 @@
 
 package com.saveourtool.save.gateway.security
 
+import com.saveourtool.save.authservice.utils.username
 import com.saveourtool.save.gateway.config.ConfigurationProperties
 import com.saveourtool.save.gateway.service.BackendService
 import com.saveourtool.save.gateway.utils.StoringServerAuthenticationSuccessHandler
@@ -195,7 +196,7 @@ private fun userStatusBasedAuthorizationDecision(
     backendService: BackendService,
     authentication: Mono<Authentication>,
     authorizationContext: AuthorizationContext,
-) = authentication.flatMap { backendService.findByName(it.name) }
+) = authentication.flatMap { backendService.findByAuthentication(it) }
     .filter { it.isEnabled }
     .flatMap { authorizationManagerAuthorizationDecision(authentication, authorizationContext) }
     .defaultIfEmpty(AuthorizationDecision(false))
--- a/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
+++ b/api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
@@ -6,7 +6,10 @@ import com.saveourtool.save.gateway.config.ConfigurationProperties
 import com.saveourtool.save.utils.orNotFound
 
 import org.springframework.http.MediaType
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken
+import org.springframework.security.core.Authentication
 import org.springframework.security.core.userdetails.UserDetails
+import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.toEntity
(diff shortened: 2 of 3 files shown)


api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
```diff
@@ -52,6 +55,38 @@ class BackendService(
...
+    /**
+     * Find current user [SaveUserDetails] by [authentication].
+     *
+     * @param authentication current user [Authentication]
+     * @return current user [SaveUserDetails]
+     */
+    fun findByAuthentication(authentication: Authentication): Mono<SaveUserDetails> = when (authentication) {
+        is UsernamePasswordAuthenticationToken -> findByName(authentication.name)
+        is OAuth2AuthenticationToken -> {
+            val source = authentication.authorizedClientRegistrationId
```
akuleshov7, line 67: I would actually move this code to some common  method...

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
```diff
@@ -52,6 +55,38 @@ class BackendService(
...
+     *
+     * @param authentication current user [Authentication]
+     * @return current username
+     */
+    fun findNameByAuthentication(authentication: Authentication?): Mono<String> = when (authentication) {
+        is UsernamePasswordAuthenticationToken -> authentication.name.toMono()
+        is OAuth2AuthenticationToken -> {
+            val source = authentication.authorizedClientRegistrationId
+            val nameInSource = authentication.name
+            findByOriginalLogin(source, nameInSource).map { it.name }
```
akuleshov7, line 85: and in this method just do commonMethod.name() ...

akuleshov7 (approved):

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
sanyavertolet: the point is that `UsernamePasswordAuthenticationToken` path varies - sometimes we want to make a request while sometimes we don't want

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
sanyavertolet: When `authentication` is `OAuth2AuthenticationToken`, we must send a request, while when `authentication` is `UsernamePasswordAuthenticationToken`, there are to different variants.

nulls: ok, then
```kotlin
fun findNameByAuthentication(authentication: Authentication?): Mono<String> = when (authentication) {
        is UsernamePasswordAuthenticationToken -> authentication.name.toMono()
        else -> findByAuthentication(authentication).map { it.name }
    }
```

just to minimize duplication

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
nulls: and btw, we have similar logic here `AuthorizationHeadersGatewayFilterFactory`

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/security/WebSecurityConfig.kt
```diff
@@ -195,7 +196,7 @@ private fun userStatusBasedAuthorizationDecision(
     backendService: BackendService,
     authentication: Mono<Authentication>,
     authorizationContext: AuthorizationContext,
-) = authentication.flatMap { backendService.findByName(it.name) }
```
nulls, line 198: actually, you don't need to get a request to backend here -- authentication.name for basic auth is the save user name here

sanyavertolet: I do need to make a request in order to get `isActive` flag

api-gateway/src/main/kotlin/com/saveourtool/save/gateway/service/BackendService.kt
sanyavertolet: Well, this one is ok, but there are two type casts...

PR #2587, opened 2023-09-13
Refactored storage for COSV files

### LINKED ISSUE #2558: Need to store Cosv files using ID from database
`modified` can omit seconds and mills

nulls:
### What's done:
- added table cosv_file
- added a new storage CosvFileStorage which uses database
- added a migration storage to migrate COSV file

It closes #2558


sanyavertolet (approved):
lgtm

PR #2603, opened 2023-09-19
Added a button to download vulnerability in COSV

### LINKED ISSUE #2602: Allow to download COSV file

nulls:
### What's done:
- added a new method in CosvRepository to download COSV as a stream
- added an endpoint to download vulnerability as COSV
- added a button on FE

It closes #2602


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityHeader.kt
```diff
@@ -247,6 +249,18 @@ internal val headerMenu: FC<HeaderMenuProps> = FC { props ->
                                 props.setIsTableView { !it }
                             }
                         }
+                        if (props.selectedMenu == VulnerabilityTab.INFO) {
+                            a {
```
Cheshiriks, line 253: I think ```Link { }``` would be preferable here

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityHeader.kt
nulls: looks like `Link` is a link to some page on site.
I tried it and it redirect to main page and that all

Cheshiriks: Oh, I didn’t notice that you have a file download here. OK

nulls: found example with full url, will try to again

nulls: tried with `Link {}` and relative path -- doesn't work too

PR #2608, opened 2023-09-22
Translate Base Score Calculator

Cheshiriks:


nulls (approved):

save-frontend/src/main/resources/locales/cn/proposing.json
```diff
@@ -7,7 +7,6 @@
   "Original vulnerability identifier": "原始漏洞ID",
   "Related link": "相关链接",
   "Language": "语言",
-  "Criticality": "危险程度",
```
nulls, line 10: why removed?

Cheshiriks: This field has been removed from the frontend

nulls: In this PR or in another one? Because I don't see these changes here

PR #2621, opened 2023-09-27
Another try to fix Dockerfile

nulls:
### What's done:
- added arg for BASE_IMAGE


akuleshov7 (approved):

save-agent/src/commonMain/kotlin/com/saveourtool/save/agent/SaveAgent.kt
```diff
@@ -210,7 +208,7 @@ class SaveAgent(
             }
 
         // a temporary workaround for python integration
-        executeAdditionallySetup(targetDirectory, agentInitConfig.additionalFileNameToUrl.keys, agentInitConfig.setupShTimeoutMillis)
```
akuleshov7, line 213: unused or what?

PR #2622, opened 2023-09-27
Changed vulnerabilityCollectionView with vulnerabilityDto to vulnerabilityMetadataDto

Cheshiriks:
Changed vulnerabilityCollectionView with vulnerabilityDto to vulnerabilityMetadataDto


akuleshov7 (approved):
very small and easy solution
But where do you update tags?

nulls:
> very small and easy solution But where do you update tags?

we do it using a dedicated endpoint for it: `com.saveourtool.save.backend.controllers.vulnerability.VulnerabilityController#saveTag`

PR #2639, opened 2023-09-29
One contest per execution

### LINKED ISSUE #2636: Delete all executions is broken
500 error on the history view

```
    private fun deleteExecutionsBuilder() {
        scope.launch {
            val responseFromDeleteExecutions =
                    post(
                        "$apiUrl/execution/delete-all-except-contest?name=${props.name}&organizationName=${props.organizationName}",
                        jsonHeaders,
                        undefined,
                    )

            if (responseFromDeleteExecutions.ok) {
                window.location.href = "${window.location.origin}#/${props.organizationName}/${props.name}"
            }
        }
    }
```

nulls:
### What's done:
- added contain in lnk_contest_execution
- updated test data

It closes #2636


save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
```diff
@@ -190,7 +191,7 @@ class ExecutionController(private val executionService: ExecutionService,
         Permission.DELETE,
         messageIfNotFound = "Could not find the project with name: $name and owner: $organizationName or related objects",
     )
-        .map { project ->
```
kgevorkyan, line 193: blockingMap prohibits different context, because until first one context is not over, the second one wont be started for one execution?

kgevorkyan (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/ExecutionController.kt
nulls: nope, blockingMap is useful here because `executionService.deleteExecutionExceptParticipatingInContestsByProjectNameAndProjectOrganization(project.name, project.organization)` is blocking operation (database update inside transaction). It does use a special scheduler which cannot be suspended till the operation is ended.
The main fix is in updating test data and adding constrain on database level

kgevorkyan: > The main fix is in updating test data and adding constrain on database level

but this does not exclude the case, when we will add several contests for one execution, isnt it?
now we will just fail on such case, but the main algorithm allow such behavior

PR #2659, opened 2023-10-03
Fixed Enum value for RawCosvFileStatus

nulls:
### What's done:
- fixed enum type for RawCosvFileStatus
- added liquibase script to fix the data

It's part of 2658

### DIFF
--- a/save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
+++ b/save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
@@ -6,10 +6,7 @@ import com.saveourtool.save.spring.entity.BaseEntityWithDtoWithId
 import com.saveourtool.save.spring.entity.IBaseEntityWithDate
 
 import java.time.LocalDateTime
-import javax.persistence.Entity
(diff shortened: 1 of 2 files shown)


Cheshiriks (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
```diff
@@ -34,7 +31,7 @@ class RawCosvFile(
     @ManyToOne
     @JoinColumn(name = "organization_id")
     var organization: Organization,
-    @Enumerated
+    @Enumerated(EnumType.STRING)
```
kgevorkyan, line 34: old files were not removed, because of this
but how they were **correctly** uploaded with incorrect type?

kgevorkyan (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
nulls: we have event on database level which removes stale entities in `raw_cosv_file` with status = `PROCESSED`

PR #2671, opened 2023-10-06
Migration to VulnerabilityExt: part 2 (tags)

akuleshov7:
### What's done:
- migrating /update endpoint to VulnerabilityExt (including tags)

### What's left:
- migration of /save endpoint for new vulnerabilities


akuleshov7:
Screen Recording 2023-10-06 at 22 06 47

akuleshov7:
Screen Recording 2023-10-06 at 22 06 47 (1)

save-backend/src/main/kotlin/com/saveourtool/save/backend/service/TagService.kt
```diff
@@ -33,6 +36,8 @@ class TagService(
      */
     @Transactional
     fun addVulnerabilityTag(identifier: String, tagName: String): LnkVulnerabilityMetadataTag {
+        log.info("Trying to add $tagName to $identifier vulnerability")
```
nulls, line 39:
```suggestion
        log.info { "Trying to add $tagName to $identifier vulnerability" }
```

akuleshov7: whyyyyyyyy?

akuleshov7: it's not idiomatic Kotlin, only Unit-like parameters should have it curly brackets

nulls: because now string is built each time, with usign lambda -- it will be calculated only when info is enabled

akuleshov7: ah, we have utils for it

nulls (approved):

PR #2689, opened 2023-10-11
Fix the bug with multipart upload

### LINKED ISSUE #2688: A bug in multipart upload with min size of part

nulls:
### What's done:
- added a method for projectreactor: `bufferAccumulatedUntil`
- applied the new method to multi part upload with min size 5 Mb
- supported abort for multipart upload in case of any error

It closes #2688


save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/DefaultStorageProjectReactor.kt
```diff
@@ -188,7 +191,18 @@ class DefaultStorageProjectReactor<K : Any>(
                 { function(this) }.toMono()
             }
 
-    private fun Pair<CompletedPart, Long>.completedPart() = first
+    companion object {
+        private const val MULTI_PART_UPLOAD_MIN_PART_SIZE_IN_BYTES = 5 * 1024 * 1024
```
kgevorkyan, line 195: please add comments, or link to docs/stackoverflow about this
I beleive we will forget about it very soon without explanation

kgevorkyan (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/storage/DefaultStorageProjectReactor.kt
nulls: added link to S3 limits + supported abort operation for case when we fail to upload some part

akuleshov7 (approved):

0x6675636b796f75676974687562 (approved):

PR #2697, opened 2023-10-12
Fixed criticality score

Cheshiriks:
Fix criticality score

### DIFF
--- a/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
+++ b/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
@@ -145,9 +145,7 @@ class VulnerabilityMetadataService(
         ): VulnerabilityMetadata = apply {
             summary = entry.summary ?: entry.details?.take(SUMMARY_LENGTH_FROM_DETAILS)?.let { "$it ..." } ?: "Summary not provided"
             details = entry.details ?: "Details not provided"
-            severityNum = entry.severity?.firstOrNull()
-                ?.scoreNum
-                ?.toFloat() ?: 0f
+            severityNum = entry.severity?.firstOrNull()?.let { getScore(it) } ?: 0f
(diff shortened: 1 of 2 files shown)


akuleshov7 (approved):
Why we don’t have criticality in OSV data please check!

Cheshiriks:
> Why we don’t have criticality in OSV data please check!

Yes, in all files the severityNum is null
Снимок экрана 2023-10-12 в 12 50 27

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
```diff
@@ -145,9 +145,7 @@ class VulnerabilityMetadataService(
         ): VulnerabilityMetadata = apply {
             summary = entry.summary ?: entry.details?.take(SUMMARY_LENGTH_FROM_DETAILS)?.let { "$it ..." } ?: "Summary not provided"
             details = entry.details ?: "Details not provided"
-            severityNum = entry.severity?.firstOrNull()
-                ?.scoreNum
-                ?.toFloat() ?: 0f
+            severityNum = entry.severity?.firstOrNull()?.let { getScore(it) } ?: 0f
```
nulls, line 148:
```suggestion
            severityNum = entry.severity?.firstOrNull()?.let { it.scoreNum?.toFloat() ?: getScore(it) } ?: 0f
```

nulls:
> Why we don’t have criticality in OSV data please check!

it's a new field in COSV, not in OSV

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
Cheshiriks: It is not necessary. We decided to always calculate the value from the vector and not take it from the string. But we compare them and log them if they do not match.

PR #2725, opened 2023-10-16
Replaced old navigation

### LINKED ISSUE #2637: Migrate all window.location.href to use navigate
window.location.href does not work anymore in the project

DrAlexD:
### What's done:
- Replaced all `window.location.href` to `react.router.useNavigate()`

Closes #2637

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/HistoryView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/HistoryView.kt
@@ -43,6 +43,7 @@ import kotlinx.datetime.Instant
 import kotlinx.datetime.LocalDateTime
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
+import react.router.useNavigate
 
 /**
  * [Props] for tests execution history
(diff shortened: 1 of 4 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/HistoryView.kt
```diff
@@ -121,6 +122,8 @@ external interface FiltersProps : TableProps<ExecutionDto> {
     "TOO_MANY_LINES_IN_LAMBDA",
 )
 class HistoryView : AbstractView<HistoryProps, HistoryViewState>(Style.SAVE_LIGHT) {
+    val navigate = useNavigate()
```
nulls, line 125: it can be private, but I'm surprise that it does work.
I thought `useSomething` can be called only inside FC (@akuleshov7 @sanyavertolet )

sanyavertolet: You definitely should not use hooks inside class components

akuleshov7 (approved):
Finishing it?

akuleshov7:
@DrAlexD what's about this one? I think it will affect @Cheshiriks 's work on the split of SAVE.
What is the blocker here, why not finished?

DrAlexD:
As far as I remember, there was some kind of navigation error that I couldn't figure out. I'll watch it again soon.

DrAlexD:
Decided to close it due to the split of SAVE. It needs to be investigated anew.

PR #2742, opened 2023-10-18
Fixed issue with contentLength

nulls:
### What's done:
- moved checking of error to controller
- fixed setting contentLength

raw-cosv-files-content-length

### DIFF
--- a/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/storage/RawCosvFileS3KeyManager.kt
+++ b/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/storage/RawCosvFileS3KeyManager.kt
@@ -107,11 +107,15 @@ class RawCosvFileS3KeyManager(
     }
 
     @Transactional
-    override fun updateKeyByContentLength(key: RawCosvFileDto, contentLength: Long): RawCosvFileDto =
-            key.contentLength?.let {
-                repository.getByIdOrNotFound(key.requiredId())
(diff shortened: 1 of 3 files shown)


save-cosv/src/main/kotlin/com/saveourtool/save/cosv/storage/RawCosvFileS3KeyManager.kt
```diff
@@ -107,11 +107,15 @@ class RawCosvFileS3KeyManager(
...
-                repository.getByIdOrNotFound(key.requiredId())
-                    .let { entity ->
-                        repository.save(entity.apply { this.contentLength = contentLength }).toDto()
-                    }
-            } ?: key
+    override fun updateKeyByContentLength(
+        key: RawCosvFileDto,
+        contentLength: Long,
+    ): RawCosvFileDto = key.contentLength
+        ?.let { key }
```
nulls, line 114: the main error

akuleshov7 (approved):
While CI is running, please attach a video with upload and I archiving

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/storage/RawCosvFileS3KeyManager.kt
akuleshov7: Got it, actually I would have made the same mistake

PR #2749, opened 2023-10-18
Quick fix for redirect to COSV on authorization

akuleshov7:


api-gateway/src/main/resources/application.yml
```diff
@@ -117,7 +117,7 @@ spring:
           huawei:
             provider: huawei
             # hardcoded as for gitee
-            redirect-uri: 'https://saveourtool.com/login/oauth2/code/huawei'
+            redirect-uri: 'https://cosv.gitlink.org.cn/login/oauth2/code/huawei'
```
nulls, line 120: what's about using `{baseUrl}`?

akuleshov7: Actually yes, very good point

akuleshov7: But stackoverflow actually says different:  https://stackoverflow.com/questions/66985696/how-does-default-redirect-url-work-in-spring-security-5

akuleshov7: If you use https://someserver:port/*, then it will match all endpoints on that server. When relying on the default Spring Security endpoint {baseUrl}/login/oauth2/code/{registrationId}, it's better to be specific in the redirect URL configuration and use https://someserver:port/login/oauth2/code/{registrationId} as the value. Both options work correctly, but the latter is more secure.

akuleshov7: But I will change to {baseUrl}

Cheshiriks (approved):

kgevorkyan (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/Utils.kt
```diff
@@ -135,6 +136,14 @@ fun String.dateStringToLocalDateTime(time: LocalTime = LocalTime(0, 0, 0)) = Loc
     time,
 )
 
+/**
+ * Dirty hack for the COSV location
+ * Should be removed in future
+ *
+ * @return true if we are in COSV domains range
+ */
+fun Location.isCosvDomain() = this.hostname in setOf("cosv.dev", "cosv.gitlink.org.cn")
```
nulls, line 145: maybe?
```suggestion
fun isCosvDomain() = window.location.hostname in setOf("cosv.dev", "cosv.gitlink.org.cn")
```

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/Utils.kt
akuleshov7: In one of your calls you do window.location.run { <it>.isCosv && <it>.blabla }
So I decided to have such granularity for more simplification

PR #2785, opened 2023-10-23
Added count comments on tab

Cheshiriks:
Снимок экрана 2023-10-23 в 12 36 22


nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
```diff
@@ -39,8 +40,22 @@ val vulnerabilityView: FC<VulnerabilitiesViewProps> = FC { props ->
...
 
+    useRequest {
+        val newComments = post(
+            url = "$apiUrl/comments/get-all-count",
+            headers = jsonHeaders,
+            body = window.location.pathname,
+            loadingHandler = ::noopLoadingHandler,
+        ).unsafeMap {
+            it.decodeFromJsonString<Int>()
+        }
```
DrAlexD, line 54: Why do we need POST request here if we just get count?

nulls: because GET doesn't support `body`, but @Cheshiriks maybe refactor location to request param?

Cheshiriks: Comments are tied to the section in which we are located. Therefore, in the body of the request we need to pass a string indicating this section. And `GET` has no `body`.

Cheshiriks: Ok, changed it

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/CommentController.kt
```diff
@@ -60,10 +60,10 @@ class CommentController(
     @PostMapping("/get-all")
```
DrAlexD, line 60: Сan we change to GET request here too?

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityView.kt
DrAlexD: Can we change it to smth like this? 
```
val newComments = get(
            url = "$apiUrl/comments/get-all-count?section=$window.location.pathname",
            headers = jsonHeaders,
        ).unsafeMap {
            it.decodeFromJsonString<Int>()
        }
```

Cheshiriks: Yes, agree. Changed it

nulls: please use `params`, it's more clear way and also it use URLEncoder for params

DrAlexD (approved):

PR #2793, opened 2023-10-23
Zip is not selectable

### LINKED ISSUE #2773: First-time users may get confused about the need to unpack vulnerabilities from a ZIP archives
Once a ZIP archive is uploaded, its name appears in the UI, along with 3 buttons (_download_, _unpack_, and _delete_). Now, a first-tie user may not understand he/she needs to _unpack_ the archive first, and proceed immediately to _Submit_ or _Submit all uploaded_, resulting in JSON parser choking on a ZIP header:

image

There're three possible solutions for this problem

- Automatically unpack any ZIP content before attempting to parse JSON (probably unsuitable, since the user needs to see what exactly he/she is uploading)
- Hide or disable the _Submit..._ buttons until at least a single JSON is _unpacked_ or _selected_.
- Leave the _Submit..._ buttons enabled, but prompt the user via "You may have forgotten to unpack the content of `%s`.zip and `%s`.zip" or smth like this.

nulls:
### What's done:
- disabled checkbox for ZIP archives
- added a tooltip with reason why file is disabled

It closes #2773

### DIFF
--- a/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/controllers/RawCosvFileController.kt
+++ b/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/controllers/RawCosvFileController.kt
@@ -269,7 +269,9 @@ class RawCosvFileController(
         authentication: Authentication,
     ): Mono<StringResponse> = rawCosvFileStorage.listByOrganizationAndUser(organizationName, authentication.name)
         .map { files ->
-            files.map { it.requiredId() }
+            files
+                .filter { !it.isZipArchive() && it.status == RawCosvFileStatus.UPLOADED }
+                .map { it.requiredId() }
         }
         .flatMap { ids ->
             doSubmitToProcess(organizationName, ids, authentication)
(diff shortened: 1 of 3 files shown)


save-cosv/src/main/kotlin/com/saveourtool/save/cosv/controllers/RawCosvFileController.kt
```diff
@@ -269,7 +270,9 @@ class RawCosvFileController(
         authentication: Authentication,
     ): Mono<StringResponse> = rawCosvFileStorage.listByOrganizationAndUser(organizationName, authentication.name)
         .map { files ->
-            files.map { it.requiredId() }
+            files
+                .filter { !it.isZipArchive() && it.status == RawCosvFileStatus.UPLOADED }
```
kgevorkyan, line 274: May be even create `it.isJsonFile()`?

kgevorkyan (approved):

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/controllers/RawCosvFileController.kt
nulls: will create a new issue for it. I think we should also skip non json and non zip files on backend side

nulls: created #2802

Cheshiriks (approved):

0x6675636b796f75676974687562 (approved):

Issue #2790, opened 2023-10-23
Polishing left column in vulnerability view (part 2)


nulls:
Follow up after: #2519

- [ ] Editing of info should work properly: now it does not work for all fields.
- [ ] We need extra information: propagate `aliases` to the frontend. 
- [ ] The second step is to **approve aliases** by vuln owners, so no one will add some invalid links.
- [ ] Show Related Vulnerabilities: related[]: List of String

PR #2799, opened 2023-10-24
Add aliases field to the left column

kgevorkyan:
### What's done:
* Add aliases field to the left column

Part of https://github.com/saveourtool/save-cloud/issues/2790

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
@@ -220,6 +220,7 @@ val vulnerabilityGeneralInfoProps: FC<VulnerabilityGeneralInfoProps> = FC { prop
                         }
                         onChange = { event ->
                             props.setVulnerability { vulnerability ->
+                                @Suppress("COMPLEX_EXPRESSION")
                                 vulnerability?.copy(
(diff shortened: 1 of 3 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
```diff
@@ -221,14 +221,13 @@ val vulnerabilityGeneralInfoProps: FC<VulnerabilityGeneralInfoProps> = FC { prop
                         onChange = { event ->
                             props.setVulnerability { vulnerability ->
                                 vulnerability?.copy(
-                                        cosv = cosv.copy(aliases = event.target.value.split(PRETTY_DELIMITER))
+                                    cosv = cosv.copy(aliases = event.target.value.split(PRETTY_DELIMITER))
```
nulls, line 224: maybe add checking on duplicates in aliases?

nulls (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
kgevorkyan: add `distinct`

nulls (approved):

akuleshov7 (approved):
About Aliases: we wanted to have something like a relation between COSV vulnerabilities. 
The idea is the following: 
- we have one COSV that can be mapped to another (or to duplicates)

Cheshiriks (approved):

PR #2803, opened 2023-10-24
Added notifications for approving new users

Cheshiriks:
- added UserListener class
 - added notification and lnkUserNotification tables


kgevorkyan (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/LnkUserNotification.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @ManyToOne
+    @JoinColumn(name = "notification_id")
+    var notification: Notification,
+
+    @ManyToOne
+    @JoinColumn(name = "user_id")
+    var user: User,
+
+    var isShow: Boolean,
```
nulls, line 24: do we need it? maybe just remove notification if it's shown?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/LnkUserNotification.kt
```diff
@@ -0,0 +1,26 @@
...
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+
+/**
+ * @property notification
+ * @property user
+ * @property isShow
+ */
+@Entity
+class LnkUserNotification(
```
nulls, line 14: do we need it? maybe add `user` and `isShow` to `Notification`?
I understand that you try to avoid duplicates in notification, but I think it can complicate the logic

db/v-3/core/init/_all_changelogs.xml
```diff
@@ -15,5 +15,7 @@
     <include file="lnk-project-github.xml" relativeToChangelogFile="true"/>
     <include file="comments.xml" relativeToChangelogFile="true"/>
     <include file="tag.xml" relativeToChangelogFile="true"/>
+    <include file="notification.xml" relativeToChangelogFile="true"/>
```
nulls, line 18: why you add it here?

`init` was added for existed tables, for new one we can create a separate changelog file.
also, we can follow your suggestion and create a single changelog file for several tables (which are changed in scope of a single task)

Cheshiriks: Removed the table from _init. As I understand it, we want to create new tables in separate files, but make changes in one common one?

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/LnkUserNotification.kt
Cheshiriks: Yes, I wanted to get rid of duplicates, but let it be as you suggest.

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/LnkUserNotification.kt
Cheshiriks: Yes, I agree, I deleted it

nulls (approved):

db/v-3/core/init/_all_changelogs.xml
nulls: >  we want to create new tables in separate files
I didn't want (originally it was added for existed tables), but it sounds good.

So,
1. we create a single file for first creation table in `init`
2. all changes are in separate files in `db/v-3/<PART>/`

DrAlexD (approved):

PR #2806, opened 2023-10-25
Common props for user info

nulls:
### What's done:
- added a common props for user info
- reused it in all props

It's part of #2221

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
@@ -11,7 +11,6 @@ import com.saveourtool.save.frontend.components.modal.modal
 import com.saveourtool.save.frontend.components.views.usersettings.right.SettingsInputFields
 import com.saveourtool.save.frontend.externals.i18next.TranslationFunction
 import com.saveourtool.save.frontend.utils.*
-import com.saveourtool.save.info.UserInfo
 import com.saveourtool.save.validation.FrontendRoutes
 
 import js.core.jso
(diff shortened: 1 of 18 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/usersettings/SettingsView.kt
```diff
@@ -93,24 +92,11 @@ typealias FieldsStateSetter = StateSetter<SettingsInputFields>
  * `Props` retrieved from router
  */
 @Suppress("MISSING_KDOC_CLASS_ELEMENTS")
-external interface SettingsProps : PropsWithChildren {
-    /**
-     * Currently logged-in user or null
-     */
-    var userInfo: UserInfo?
-
+external interface SettingsProps : UserInfoAwareMutablePropsWithChildren {
```
akuleshov7, line 95: Actually, I do not like the idea of extending plain classes and making inheritance on properties. 
Properties are simple structures aligned to a specific view, usually they should not have a common part. In my opinion - if needed, we can have common properties for views with multiple FC in it, but it also should be plain.

This particular inheritance can cause copy-paste issues for new people who will be simply copying properties and inheritance on classes and bring userInfo setter all over the code. 

I do not have strong objection about this change, just a note.

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/utils/UserInfoAwareProps.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Property to propagate user info from App with ability to update it
+ */
+external interface UserInfoAwareMutableProps : UserInfoAwareProps {
+    /**
+     * Setter of user info (it can be updated in settings on several views)
+     *
+     * After updating user information we will update userSettings without re-rendering the page
+     * PLEASE NOTE: THIS PROPERTY AFFECTS RENDERING OF WHOLE APP.KT
+     * IF YOU HAVE SOME PROBLEMS WITH IT, CHECK THAT YOU HAVE PROPAGATED IT PROPERLY:
```
akuleshov7, line 37: Users of this code will definitely forget about it and will have hours of investigation where did they forgot to pass `this.setter = setter` in a call chain. No doubts, but anyway, lgtm

akuleshov7 (approved):
lgtm

Cheshiriks (approved):

0x6675636b796f75676974687562 (approved):

PR #2817, opened 2023-10-27
Archives with raw COSV files should be at the top of the list

### LINKED ISSUE #2815: Archives with raw cosv files should be at the top of the list

DrAlexD:
### What's done:
- added sorting of raw COSV files depending on whether the file is an archive.

Closes #2815

gif8


kgevorkyan (approved):

Cheshiriks (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
```diff
@@ -34,6 +37,8 @@ class RawCosvFile(
     var organization: Organization,
     @Enumerated(EnumType.STRING)
     var status: RawCosvFileStatus,
+    @Formula("status = 'UPLOADED' AND LOWER(file_name) LIKE '%_.zip'")
```
nulls, line 40: I don't think that need to check additonally status -- file name will be enough

nulls (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/entities/cosv/RawCosvFile.kt
DrAlexD: We work with archives when they have only `UPLOADED` status. They can't have any other status. So to show this I decided that it would be better to also add status check. Should I remove this check? Or I can rename this field to `isUploadedZip`?

nulls: I think better to remove: it also remove hardcoding on status name  (`UPLOADED`)

DrAlexD: Done

nulls (approved):

PR #2833, opened 2023-11-01
Added notification for propose new vulnerability

Cheshiriks:
Added notification for propose new vulnerability

### DIFF
--- a/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/event/VulnerabilityMetadataListener.kt
+++ b/save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/event/VulnerabilityMetadataListener.kt
@@ -5,6 +5,7 @@ import com.saveourtool.save.backend.service.UserDetailsService
 import com.saveourtool.save.domain.Role
 import com.saveourtool.save.entities.Notification
 import com.saveourtool.save.entities.cosv.VulnerabilityMetadata
+import com.saveourtool.save.entities.cosv.evententities.VulnerabilityMetadataEvent
 import com.saveourtool.save.entities.vulnerability.VulnerabilityStatus
 import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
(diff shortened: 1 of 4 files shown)


save-backend/src/main/kotlin/com/saveourtool/save/backend/event/VulnerabilityMetadataListener.kt
```diff
@@ -0,0 +1,46 @@
+package com.saveourtool.save.backend.event
```
akuleshov7, line 1: please create events in vuln/event/ directory, so it will be easier to split it in the future

save-backend/src/main/kotlin/com/saveourtool/save/backend/event/VulnerabilityMetadataListener.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Component
+class VulnerabilityMetadataListener(
+    private val userDetailsService: UserDetailsService,
+    private val notificationService: NotificationService,
+) {
+    /**
+     * @param vulnerability new VulnerabilityMetadata
+     */
+    @EventListener
```
akuleshov7, line 23: So this event is triggered when vulnerabilityMetadata is changed?

Cheshiriks: Only in the place where we subscribed up for it. Right now I have only added the subscription to the create method. Then I want to add to change the vulnerability. But now we have the same status for both new and changed vulnerabilities, so we cannot distinguish between them.

save-backend/src/main/kotlin/com/saveourtool/save/backend/event/VulnerabilityMetadataListener.kt
Cheshiriks: Moved to a new directory

save-backend/src/main/kotlin/com/saveourtool/save/backend/event/VulnerabilityMetadataListener.kt
akuleshov7: Where is the subscription? I do not see it in this PR

akuleshov7 (approved):

save-backend/src/main/kotlin/com/saveourtool/save/backend/event/VulnerabilityMetadataListener.kt
Cheshiriks: it: `applicationEventPublisher.publishEvent(newMetadata)`

akuleshov7: missed that, yes, sorry

save-backend/src/main/kotlin/com/saveourtool/save/backend/controllers/vulnerability/event/VulnerabilityMetadataListener.kt
```diff
@@ -0,0 +1,46 @@
...
+@Component
+class VulnerabilityMetadataListener(
+    private val userDetailsService: UserDetailsService,
+    private val notificationService: NotificationService,
+) {
+    /**
+     * @param vulnerability new VulnerabilityMetadata
+     */
+    @EventListener
+    fun createVulnerabilityMetadata(vulnerability: VulnerabilityMetadata) {
```
akuleshov7, line 24: the only thing I am worried here is about the argument type. I would have made some Event class for it, it can be a value/inline class, for example

nulls: the same for User, after it we checks the vulnerability status. I guess we create create a required event in place when `applicationEventPublisher` is called

Cheshiriks: Added Event class

DrAlexD (approved):

PR #2840, opened 2023-11-02
Added CREATED status for new vulnerability

Cheshiriks:
- Added CREATED status for new vulnerability

### DIFF
--- a/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/vulnerability/VulnerabilityStatus.kt
+++ b/save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/vulnerability/VulnerabilityStatus.kt
@@ -21,6 +21,11 @@ enum class VulnerabilityStatus {
     AUTO_APPROVED,
 
     /**
+     * Created status
+     */
+    CREATED,
+
+    /**
      * Review status
      */
     PENDING_REVIEW,
--- a/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
+++ b/save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
@@ -102,7 +102,7 @@ class VulnerabilityMetadataService(
             user = user,
             organization = organization,
             language = getLanguage() ?: VulnerabilityLanguage.OTHER,
-            status = isAutoApprove.toVulnerabilityStatus(true),
+            status = toVulnerabilityStatus(isAutoApprove, true),
             latestCosvFile = cosvFile,
         )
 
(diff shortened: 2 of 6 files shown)


save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/vulnerability/VulnerabilityStatus.kt
```diff
@@ -20,6 +20,11 @@ enum class VulnerabilityStatus {
      */
     AUTO_APPROVED,
 
+    /**
+     * Created status
+     */
+    CREATED,
+
```
DrAlexD, line 27: Is it clear enough for users and admins that `CREATED` status also requires approve from admins, just like `PENDING_REVIEW`? Is status `CREATED` necessary? Maybe we should rename it somehow and then maybe also `PENDING_REVIEW` status?

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
```diff
@@ -172,6 +171,12 @@ class VulnerabilityMetadataService(
...
                 } ?: 0f
 
-        private fun Boolean.toVulnerabilityStatus() = if (this) VulnerabilityStatus.AUTO_APPROVED else VulnerabilityStatus.PENDING_REVIEW
+        private fun Boolean.toVulnerabilityStatus(isNew: Boolean) = if (this) {
+            VulnerabilityStatus.AUTO_APPROVED
+        } else if (isNew) {
+            VulnerabilityStatus.CREATED
+        } else {
+            VulnerabilityStatus.PENDING_REVIEW
+        }
```
DrAlexD, line 180: Maybe change this function to something like `private fun getVulnerabilityStatus(isAutoApprove: Boolean, isNew: Boolean)`?

kgevorkyan (approved):

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/vulnerability/VulnerabilityStatus.kt
Cheshiriks: Statuses are hidden in the system. There are only two states for users: _Approved_ and _Not Approved_.
But it is important for us to distinguish new vulnerabilities from those that have simply been changed.
Changing new ones will not have to create new versions. While old vulnerabilities must have a copy of the version before and after the change.

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/service/VulnerabilityMetadataService.kt
Cheshiriks: Ok, changed it

save-cloud-common/src/commonMain/kotlin/com/saveourtool/save/entities/vulnerability/VulnerabilityStatus.kt
DrAlexD: Ok, agreed

nulls (approved):

PR #2851, opened 2023-11-08
Fix error after clicking on the buttons to go to Customize profile, Settings, Manage organizations

kgevorkyan:
### What's done:
* Fix ERROR after clicking on the buttons to go to Customize profile, Settings, Manage organizations

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
@@ -37,7 +37,8 @@ fun initI18n(isDebug: Boolean = false, interpolationEscapeValue: Boolean = false
                     'vulnerability',
                     'comments',
                     'dates',
-                    'index'
+                    'index',
+                    'profile'
                 ],
                 backend: {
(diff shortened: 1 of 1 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
```diff
@@ -37,7 +37,8 @@ fun initI18n(isDebug: Boolean = false, interpolationEscapeValue: Boolean = false
                     'vulnerability',
                     'comments',
                     'dates',
-                    'index'
+                    'index',
+                    'profile'
```
nulls, line 41: can be added a `,` at the end to avoid change in last line when a new value is added

nulls (approved):

akuleshov7 (approved):

Cheshiriks (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/externals/i18next/InitI18n.kt
kgevorkyan: Didnt know supported it by `js` native code or not, so add without `,`

PR #2857, opened 2023-11-10
Distinct user in Contributors and Credits on vulnerability page

kgevorkyan:
### What's done:
* Distinct user in Contributors and Credits on vulnerability page

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
@@ -335,7 +335,7 @@ val vulnerabilityGeneralInfoProps: FC<VulnerabilityGeneralInfoProps> = FC { prop
                         +"Contributors".t()
                     }
                     userBoard {
-                        users = saveContributors
+                        users = saveContributors.distinctBy { it.name }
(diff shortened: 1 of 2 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/views/vuln/VulnerabilityGeneralInfoProps.kt
```diff
@@ -348,8 +348,8 @@ val vulnerabilityGeneralInfoProps: FC<VulnerabilityGeneralInfoProps> = FC { prop
                         className = ClassName("font-weight-bold text-primary-blue mb-4")
                         +"Credits"
                     }
-                    cosv.credits!!.forEachIndexed { creditInd, credit ->
-                        val userPrefix = if (credit.name in saveContributors.map { it.name }) "[SAVE] user" else "External user"
+                    cosv.credits!!.distinctBy { it.name }.forEachIndexed { creditInd, credit ->
```
nulls, line 351: I'm not sure that we should hide some content from origin file.
can we merge them at least? if we have a duplicate by `it.name` then merge content of contacts

kgevorkyan: Good point, changed

nulls (approved):

PR #2859, opened 2023-11-13
migration to multiplatform

nulls:


save-frontend/src/jsMain/resources/locales/en/profile.json
```diff
@@ -1,17 +1,17 @@
-{
```
akuleshov7, line 1: strange that GitHub shows it as a change

akuleshov7 (approved):
can you please add a description with links for our future younger generations with the reason of why this change is needed after K2. And link to the Kotlin recommendation about migrating to Multiplatform

save-frontend/src/jsMain/resources/locales/en/profile.json
nulls: +1

nulls:
> can you please add a description with links for our future younger generations with the reason of why this change is needed after K2. And link to the Kotlin recommendation about migrating to Multiplatform

actually, this migration is not required for us. I tried to resolve issue with Serialization on Kotlin JS.

But this the link to suggestion to migration to kotlin multiplatform: https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#kotlin-js-plugin-deprecation

PR #2860, opened 2023-11-13
Fixed text displaying in tables on user and organization pages

### LINKED ISSUE #2643: Tables: check project and organization and remove ugly tables (replace them with our good-looking tables). Start from vulnerabilities - unify it's logic

DrAlexD:
### What's done:
- vulnerabilities and users tabs are now only displayed if there are new users to approve.
- added text centering in tables on user, organization and project pages.
- added colors for `Status` field depending on its value in the vulnerability table.
- added colors for `Criticality` field depending on its value in the vulnerability table.

image

Closes #2643

### DIFF
--- a/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
+++ b/save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
@@ -40,6 +40,7 @@ private val contestsTable: FC<OrganizationContestsTableProps<ContestDto>> = tabl
             column(id = "name", header = "Contest Name", { name }) { cellContext ->
                 Fragment.create {
                     td {
+                        className = ClassName("align-middle text-center")
                         Link {
                             to = "/contests/${cellContext.row.original.name}"
                             +cellContext.value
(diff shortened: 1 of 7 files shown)


save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
```diff
@@ -89,8 +94,6 @@ private val contestsTable: FC<OrganizationContestsTableProps<ContestDto>> = tabl
             }
         }
     },
-    initialPageSize = 10,
```
kgevorkyan, line 92: ?

kgevorkyan (approved):

save-frontend/src/main/kotlin/com/saveourtool/save/frontend/components/basic/organizations/OrganizationContestsMenu.kt
DrAlexD: There is no need to pass `initialPageSize = 10` because the default value for this property is 10 and is contained in the `INITIAL_TABLE_PAGE_SIZE` constant.

Cheshiriks (approved):

nulls (approved):

PR #2864, opened 2023-11-28
Removed duplication for MAGIC_NUMBER in diktat-analysis.yml

nulls:
### What's done:
- removed duplication in diktat-analysis.yml
- fixed reference to param


diktat-analysis.yml
```diff
@@ -107,8 +107,10 @@
   configuration:
     valueArgument: true
     valueParameter: true
+# reduces speed of development on the FE
+# will remove it by for now
 - name: MAGIC_NUMBER
-  enabled: true
+  enabled: false
```
kgevorkyan, line 113: there is no any configuration, such we would disable rule only on frontend paths? `save-frontend/*`?

kgevorkyan (approved):

Cheshiriks (approved):

diktat-analysis.yml
nulls: we can put `diktat-analysis.yml` inside to save-frontend. Will test it after splitting

nulls: actully, it doesn't work: we configure diktat in extension by the following:
```kotlin
diktat {
    diktatConfigFile = rootProject.file("diktat-analysis.yml")
}
```

Looks like we need to modify this configuration to something like:
```kotlin
diktatConfigFile = project.file("diktat-analysis.yml").takeIf { it.exists() } ?: rootProject.file("diktat-analysis.yml")
```

PR #2887, opened 2023-12-20
Feature/his internal profile

acies312:


api-gateway/src/main/resources/application-internal.yml
```diff
@@ -0,0 +1,50 @@
+gateway:
+  backend:
+    url: http://localhost:5800
```
nulls, line 3: no need to duplicate these properties from `dev`. we can enable two profiles at the same time: `dev,internal`.
I think need to leave only: `spring.security.oauth2.client.provider.codehub.*` and `spring.security.oauth2.client.registration.codehub.*`

acies312: deleted all unnecessary properties so that we can use only `internal` profile in HIS cloud

nulls (approved):

api-gateway/src/main/resources/application-internal.yml
```diff
@@ -0,0 +1,22 @@
...
+            user-name-attribute: username
+            user-info-authentication-method: post
+        registration:
+          codehub:
+            provider: codehub
+            redirect-uri: '${gateway.frontend.url}/{action}/oauth2/code/{registrationId}'
+            authorization-grant-type: authorization_code
+            client-authentication-method: client_secret_post
+            # can be
+            #  - openid (doesn't work due to invalid_nonce)
```
nulls, line 19: can you please remove this comment, since it relates to Huawei provider\registration

acies312: done

save-frontend-common/src/main/kotlin/com/saveourtool/save/frontend/common/utils/LoginUtils.kt
```diff
@@ -110,5 +110,6 @@ fun mapKnownUploadedIcons(registrationId: String) =
             "gitee" -> "/img/gitee.svg"
             "github" -> "/img/github.svg"
             "google" -> "/img/google.svg"
+            "codehub" -> "/img/codehub.svg"
```
nulls, line 113: I guess then we can delete `"codehub" -> faCopyright` at 96 line. Also as `faCopyright` if it's not used anymore

acies312: That was my attempt to become a part time frontend developer
to remove it i think i need help @Cheshiriks. Just deleting them cause many problems

Cheshiriks: The icon itself is still used in the Organization view. Therefore, here it is enough to delete only line 96

acies312: done

nulls (approved):

api-gateway/src/main/resources/application-internal.yml
```diff
@@ -0,0 +1,50 @@
+gateway:
```
nulls, line 1: looks like the old version of application-internal.yml was committed

acies312: yes, fixed it

PR #2894, opened 2023-12-21
Transferring service and controller vulnerabilities to cosv

Cheshiriks:


nulls (commented):
should we delete these classes from backend?

akuleshov7 (approved):
As soon as you will delete extra classes and finish service, can merge.

Cheshiriks:
> should we delete these classes from backend?

Yes. Deleted it

PR #2899, opened 2023-12-29
Deleted backend service from cosv

Cheshiriks:


kgevorkyan:
why its so many duplicate ci jobs?

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/repository/OrganizationRepository.kt
```diff
@@ -33,4 +35,58 @@ interface OrganizationRepository {
...
     )
     fun getOrganizationByName(@Param("name") name: String): Organization
+
+    /**
+     * @param userId
+     * @param organizationId
+     * @return found [Role]
+     */
+    @Query(
+        value = "select role from save_cloud.lnk_user_organization where user_id = :user_id and organization_id = :organization_id",
```
kgevorkyan, line 45: why we need native quires here?

akuleshov7: two different data sources in one app and Spring cannot work with it well

akuleshov7 (approved):

save-cosv/src/main/kotlin/com/saveourtool/save/cosv/repository/OrganizationRepository.kt
akuleshov7: but may be Vlad can use this: https://www.baeldung.com/spring-boot-configure-multiple-datasources

Cheshiriks: It's like that. In the next few PR I'll try to do it a differently

PR #2903, opened 2024-01-12
Added spring boot application for cosv

Cheshiriks:


akuleshov7:
we will need a lot of renaming...

PR #2923, opened 2024-02-14
Move WebConfig.kt to common

Cheshiriks:


kgevorkyan (approved):

save-cloud-common/src/jvmMain/kotlin/com/saveourtool/save/configs/WebConfig.kt
```diff
@@ -24,9 +24,9 @@ import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder
 import org.springframework.web.reactive.config.WebFluxConfigurer
 
 @Configuration
-class WebConfig {
+open class WebConfig {
```
nulls, line 27: allopen plugin doesn't work here?

Cheshiriks: Yes

PR #2924, opened 2024-02-20
Rename save-cosv module to cosv-backend

Cheshiriks:
- Rename `save-cosv` module to `cosv-backend`
- Rename `com.saveourtool.save.cosv` to `com.saveourtool.cosv.backend`


cosv-backend/src/main/kotlin/com/saveourtool/cosv/backend/CosvApplication.kt
```diff
@@ -9,9 +9,18 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Import
 
 /**
- * An entrypoint for spring for save-cosv
+ * An entrypoint for spring for cosv-backend
  */
-@SpringBootApplication(scanBasePackages = ["com.saveourtool.save.utils", "com.saveourtool.save"])
+@SpringBootApplication(scanBasePackages = [
+    "com.saveourtool.save.configs",
```
nulls, line 15: now I think we can have only two packages: `com.saveourtool.cosv` and `com.saveourtool.save`

save-cloud-charts/save-cloud/values.yaml
```diff
@@ -23,7 +23,7 @@ backend:
 backend_cosv:
```
nulls, line 23: should it be named to cosv_backend?

nulls (approved):

save-cloud-charts/save-cloud/values.yaml
Cheshiriks: I think this can not be changed. Our frontend is named according to the same logic.
Снимок экрана 2024-02-20 в 14 53 01

cosv-backend/src/main/kotlin/com/saveourtool/cosv/backend/CosvApplication.kt
Cheshiriks: Yes. Changed it

save-cloud-charts/save-cloud/values.yaml
kgevorkyan: so may be rename frontend too?

PR #2969, opened 2024-10-31
fix spotless linter errors

icemachined:


0x6675636b796f75676974687562 (approved):

0x6675636b796f75676974687562:
Apparently, either _Diktat_ configuration or versions are different when running the build locally vs via _GitHub Actions_.

The workaround when building locally is to skip the `spotlessKotlinCheck` task:

```bash
./gradlew build -x spotlessKotlinCheck
```

